PR #5, opened 2018-09-10
Stop on first type error

hugomg:
Simplify the type checker by stopping after the first type error.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -407,13 +407,11 @@ describe("Titan type checker", function()
         local prog, errors = run_checker([[
             record Point
                 x: float
-                y: float
             end
             local p: Point = { }
         ]])
         assert.falsy(prog)
         assert.matches("required field x is missing", errors)
-        assert.matches("required field y is missing", errors)
     end)
 
     it("forbids type hints that are not array or records", function()
@@ -579,7 +577,7 @@ describe("Titan type checker", function()
         local code = [[
             function fn(x: integer): integer
                 local i: integer = 15
-                while i do
+                while i > 0 do
                     local s: string = i
                 end
                 return x
@@ -1263,14 +1261,17 @@ describe("Titan type checker", function()
 
     it("catches assignment to function", function ()
         local code = [[
-            function foo(): integer
-                foo = 2
+            function f()
+            end
+
+            function g()
+                f = g
             end
         ]]
         local prog, errs = run_checker(code)
         assert.falsy(prog)
         assert.match(
-            "attempting to assign to toplevel constant function foo",
+            "attempting to assign to toplevel constant function f",
             errs, nil, true)
     end)
 
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -1194,6 +1194,38 @@ describe("Titan type checker", function()
         end
     end
 
+    for _, op in ipairs({"+", "-", "*", "//", "/", "^"}) do
+        for _, t in ipairs({"{integer}", "boolean", "string"}) do
+            it("cannot use arithmetic operator " .. op .. " when left hand side is not a  number", function()
+                local code = [[
+                    function fn(a: ]] .. t .. [[, b: float): float
+                        return a ]] .. op .. [[ b
+                    end
+                ]]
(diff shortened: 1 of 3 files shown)


titan-compiler/checker.lua
```diff
@@ -93,11 +100,67 @@ local function coerce_numeric_exp_to_float(exp)
...
-check_program = function(prog, errors)
+typecheck = function(prog)
+    local ok, err = xpcall(check_program, debug.traceback, prog)
+    if ok then
+        return prog, {} -- TODO:  no {}
+    else
+        if getmetatable(err) == TypeError then
+            return false, { err.msg } -- TODO  no {}
+        else
+            error(err)
```
gligneul: If we call error here it will print two stack traces. We probably should do:
```
io.stderr:write(err, '\n')
os.exit(1)
```

titan-compiler/checker.lua
```diff
@@ -373,81 +422,80 @@ check_var = function(var, errors)
...
         exp._type = types.T.Void()
     else
         error("impossible")
     end
 end
 
 -- @param typehint Expected type; Used to infer polymorphic/record constructors.
-check_exp = function(exp, errors, typehint)
+check_exp = function(exp, typehint)
+    assert(typehint  ~= nil)
```
gligneul: Remove the extra space before `~=`.

titan-compiler/checker.lua
```diff
@@ -137,16 +200,15 @@ check_type = function(typ, errors)
     elseif tag == ast.Type.Name then
         local decl = typ._decl
         if decl._tag == ast.Toplevel.Record then
-            return decl._type
+            return assert(decl._type)
         else
-            type_error(errors, typ.loc, "'%s' isn't a type", typ.name)
-            return types.T.Invalid()
+            type_error(typ.loc, "'%s' isn't a type", typ.name)
```
gligneul: Couldn't we have a type_assert function? Most of the cases now are something like:
```
if something wrong then
  type_error(...)
end
```

spec/checker_spec.lua
```diff
@@ -407,13 +407,11 @@ describe("Titan type checker", function()
         local prog, errors = run_checker([[
             record Point
                 x: float
-                y: float
```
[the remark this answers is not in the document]
hugomg: Como agora o typechecker só retorna o primeiro erro, achei melhor mudar o programa para ter só um erro (x faltando) em vez de dois (x faltando e y faltando).

Se o que te incomoda for o Point ter só um "x" acho que dá para resolver de outra forma.

spec/checker_spec.lua
```diff
@@ -1196,6 +1194,38 @@ describe("Titan type checker", function()
...
+                    end
+                ]]
+                local prog, errs = run_checker(code)
+                assert.falsy(prog)
+                assert.match("left hand side of arithmetic expression is a", errs)
+            end)
+        end
+    end
+
+    for _, op in ipairs({"+", "-", "*", "//", "/", "^"}) do
```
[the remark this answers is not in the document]
hugomg: I did it this way to be more similar to the surrounding code.

It is true that there is some code duplication going on here but maybe its best if we fix all of it at once in a different commit?

titan-compiler/checker.lua
hugomg: I'm not fond of using os.exit in an internal module like this one. And if we don't use os.exit I think the duplicate stack trace is unavoidable.

titan-compiler/checker.lua
hugomg: Some cases don't work with type_assert. For example, sometimes the type error appears as part of an ifelseif chain:

```
if bla._tag == "AAA" then
   type error
elseif bla._tag == "BBB" then
   do something
else
   error("impossible")
end
```

gligneul: We can have both, not only one.

titan-compiler/checker.lua
gligneul: Indeed.

spec/checker_spec.lua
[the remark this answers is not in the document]
gligneul: Ok, no problem.

PR #6, opened 2018-09-24
Reorder arguments in checkmatch

hugomg:
To be more consistent with the type_error function

### DIFF
--- a/titan-compiler/checker.lua
+++ b/titan-compiler/checker.lua
@@ -56,30 +56,38 @@ local function type_error(loc, fmt, ...)
 end
 
 -- Checks if two types are the same, and logs an error message otherwise
---   term: string describing what is being compared
+--   loc: location of the term that is being compared
 --   expected: type that is expected
 --   found: type that was actually present
---   loc: location of the term that is being compared
-local function checkmatch(term, expected, found, loc)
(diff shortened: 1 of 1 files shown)


titan-compiler/checker.lua
```diff
@@ -356,19 +368,26 @@ check_stat = function(stat, rettypes)
...
 
-        checkmatch("'for' start expression",
-            stat.decl._type, stat.start._type, stat.start.loc)
-        checkmatch("'for' finish expression",
-            stat.decl._type, stat.finish._type, stat.finish.loc)
-        checkmatch("'for' step expression",
-            stat.decl._type, stat.inc._type, stat.inc.loc)
+        checkmatch(stat.start.loc,
+            stat.decl._type, stat.start._type,
+            "'for' start expression")
```
gligneul: In repeat and while we have `<stat> statement ...` and here we have `'<stat>' ...`. I think we could chose one and have a standard.

hugomg: Sure... But I couldn't think of a good way to rewrite these for the "for". Any suggestions?

gligneul: We could use `''` in all of them.

PR #9, opened 2018-09-27
Desugar if-elseif into if-then-else

hugomg:
This commit gets rid of the awkward representation we have of if statments right
now, with the "Thens" array and optional else statement.

I believe the current representation started out as a side effect of the parser
algorithm and then we kept it made the generated code prettier:

```
/* old generated code */
if (c1) {
   ...
} else if (c2) {
   ...
}
```

instead of

```
/* obvious generated code*/
if (c1) {
} else {
    if (c2) {
    } else {
    }
}
```

However, this motivation doesn't apply any more now that we are switching our
generated code to a "flat" list of basic blocks.

### DIFF
--- a/titan-compiler/parser.lua
+++ b/titan-compiler/parser.lua
@@ -85,11 +85,7 @@ function defs.name_exp(pos, name)
 end
 
 function defs.ifstat(pos, exp, block, elseifs, elseopt)
-    local else_ = elseopt
-
-    if not else_ then
-        else_ = ast.Stat.Block(pos, {})
-    end
+    local else_ = elseopt or ast.Stat.Block(pos, {})
 
     for i = #elseifs, 1, -1 do
         local e = elseifs[i]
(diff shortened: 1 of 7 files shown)


titan-compiler/parser.lua
```diff
@@ -84,9 +84,23 @@ function defs.name_exp(pos, name)
     return ast.Exp.Var(pos, ast.Var.Name(pos, name))
 end
 
-function defs.ifstat(pos, exp, block, thens, elseopt)
-    table.insert(thens, 1, ast.Then.Then(pos, exp, block))
-    return ast.Stat.If(pos, thens, elseopt)
+function defs.ifstat(pos, exp, block, elseifs, elseopt)
+    local else_ = elseopt
```
gligneul: local else_ = elseopt or ast.Stat.Block(pos, {})

PR #27, opened 2018-10-03
Respect Lua's array-part size limits

### LINKED ISSUE #14: Fix limits in renormalize_array
When we implemented the array renormalize function we didn't put much thought into its limits. We should clean that up a bit.

* Is the initial size of 16 appropriate? Why not start at 1, as Lua does?
* What happens if `i` is a large value? Can the current function overflow or do something silly?
    * What are valid Lua indexes? My guess is 1 ≤ i ≤ 2<sup>63</sup> - 1 ?
    * What happens if `i` is large?
    * What happens if `i` is 1 or zero?

hugomg:
Closes #14

### DIFF
--- a/runtime/tcore.c
+++ b/runtime/tcore.c
@@ -134,20 +134,20 @@ TString *titan_string_concatN(lua_State *L, size_t n, TString **ss)
     }
 }
 
-// These definitions are from ltable.h:
+/* These definitions are from ltable.c */
 #define MAXABITS        cast_int(sizeof(int) * CHAR_BIT - 1)
 #define MAXASIZE        luaM_limitN(1u << MAXABITS, TValue)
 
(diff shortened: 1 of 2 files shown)


hugomg:
The bounds checking stuff and overflow-avoidance going on on this PR are tricky. Please pay close attention when reviewing.

runtime/tcore.c
```diff
@@ -133,15 +134,24 @@ TString *titan_string_concatN(lua_State *L, size_t n, TString **ss)
     }
 }
 
+// These definitions are from ltable.h:
```
gligneul: 1. Could we keep the C style comments like the generated code?
2. s/ltable.h/ltable.c/
3. I would remove the `:` at EOL.

PR #32, opened 2018-10-04
 Implement index and newindex metamethods for pallene records 

gligneul:


titan-compiler/coder.lua
```diff
@@ -645,127 +733,275 @@ local function rec_declare_struct(rec)
...
+        }
+    ]], {
+        ELSE = #fields > 0 and "else" or "",
+        KEY = key,
+    }))
+
+    return table.concat(cstats, "\n")
+end
+
+function RecordCoder:dyn_dispatch_bin(fields, key, action, ctx)
```
hugomg: What about receiving `fields`, `i` and `j` as a parameter instead of using a slice function to create lots of copies of the list?

titan-compiler/coder.lua
```diff
@@ -645,127 +733,275 @@ local function rec_declare_struct(rec)
...
+    local out
+    local n = #fields
+    if n <= 2 then
+        return self:dyn_dispatch_linear(fields, key, action, ctx)
```
hugomg: This is just a nested tree of if/elseif so I suspect there is no speed gain from switching to linear search in the end. Do we really need to fall back to linear search? 

1) If yes, I would like to know if you ran tests to determine the n ≤ 2 cuttof.
2) If not, I think we can make the `dispatch_bin` not call `dispatch_linear`, and put a usual base case at n = 0

hugomg: Another problem here is that there will be multiple calls to `titan_runtime_record_index_error`, which might clutter the code. Would it be possible to do something like `goto error` instead?

spec/coder_spec.lua
```diff
@@ -800,6 +800,21 @@ describe("Titan coder /", function()
             function make_gc(x: {integer}): Gc
```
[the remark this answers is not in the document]
gligneul: It's good to have test cases where either the number of primitives or GC fields are zero.

titan-compiler/coder.lua
gligneul: It shouldn't make a difference, it is there because I was testing with just linear search first. It is better to remove it.

PR #35, opened 2018-10-05
Use userdata instead of tables for upvalues

### LINKED ISSUE #34: Module-level values should be stored in an userdata array
We haven't updated our code to take advantage of this new Lua 5.4 feature yet.

hugomg:
Closes #34


hugomg:
This patch saves a memory load when accessing the upvalue array.

We still store primitive types in the array part of the userdata though. Changing it to work like records currently do would be a bit more work.

PR #41, opened 2018-10-08
Allow empty record types with no fields

### LINKED ISSUE #40: Allow empy records
We should allow record types with no fields

```
record Empty
end
```

For the parser I think it is just a matter of changing a `+` to a `*`, but I need to double check if the type checker and coder have any restrictions against them.

hugomg:
Closes #40


gligneul:
We should have some test cases for that. (I'm mostly worried about code generation).

hugomg:
I added a single coder test. I think that to add more tests we would need to merge the __index and __newindex functionality first.

PR #51, opened 2018-12-13
Util changes

hugomg:
I ~~messed around with~~ cleaned up the util.lua file.

The motivation is that in the very near future I will need that "redirect stderr to a string" functionality but I took the opportunity to change some other things along the way.


pallene/util.lua
```diff
@@ -53,32 +34,68 @@ function util.render(code, substs)
...
+    local out_file = os.tmpname()
+    local err_file = os.tmpname()
+
+    local redirected =
+        cmd ..
+        " > "  .. util.shell_quote(out_file) ..
+        " 2> " .. util.shell_quote(err_file)
+
+    local ok, err = util.execute(redirected)
+    local out_content = util.get_file_contents(out_file) or ""
```
gligneul: Is the `or ""` necessary?

spec/pallenec_spec.lua
```diff
@@ -22,21 +22,23 @@ describe("pallenec", function()
     end)
 
     it("Can compile pallene files", function()
-        util.shell("./pallenec test.pallene")
-        local out = util.shell("./lua/src/lua test_script.lua")
+        assert(util.execute("./pallenec test.pallene"))
+        local ok, err, out, _ = util.outputs_of_execute("./lua/src/lua test_script.lua")
```
gligneul:
```suggestion
        local ok, err, out = util.outputs_of_execute("./lua/src/lua test_script.lua")
```

spec/pallenec_spec.lua
```diff
@@ -22,21 +22,23 @@ describe("pallenec", function()
...
         assert.equals("17\n", out)
     end)
 
     it("Can compile C files", function()
-        util.shell("./pallenec --emit-c test.pallene")
-        util.shell("./pallenec --compile-c test.c")
-        local out = util.shell("./lua/src/lua test_script.lua")
+        assert(util.execute("./pallenec --emit-c test.pallene"))
+        assert(util.execute("./pallenec --compile-c test.c"))
+        local ok, err, out, _ = util.outputs_of_execute("./lua/src/lua test_script.lua")
```
gligneul:
```suggestion
        local ok, err, out = util.outputs_of_execute("./lua/src/lua test_script.lua")
```

spec/pallenec_spec.lua
hugomg: I think the `_` makes it clearer that outputs_of_execute is returning 4 values and that we are intentionally ignoring the last one.

Issue #24, opened 2018-10-02
Fix pending test in coder_spec


hugomg:
In line 853, there in the test for "check record tags" we still have a `pending("fix error message")`. Why is this still pending?

hugomg:
The offending error message is "expected userdata but found userdata". It should be updated to mention the appropriate record types instead.

I am leaving this issue open but I am removing the "pending" warning for now, because it is getting in the way when running tests.

PR #54, opened 2019-02-27
Remove "pending" warning from record tag errors

hugomg:
See discussion on issue #24

### DIFF
--- a/pallene/builtins.lua
+++ b/pallene/builtins.lua
@@ -2,23 +2,16 @@ local ast = require "pallene.ast"
 local location = require "pallene.location"
 local types = require "pallene.types"
 
-
 local builtins = {}
 
--- I don't expect this to ever be shown to the end user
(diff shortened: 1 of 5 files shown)


pallene/builtins.lua
```diff
@@ -1,7 +1,24 @@
...
+local builtins = {}
+
+-- I don't expect this to ever be shown to the end user
+local loc = location.new("(builtin)", 0,0)
+
+--
+-- Builtin functions
+--
+
+builtins.io_write = ast.Toplevel.Builtin(loc, "io.write")
```
gligneul: We could create a for loop here:
```
local functions = {
    "io.write",
}
for _, f in ipairs(functions) do
    local escaped = f:gsub('.', '_')
    builtins[escaped] = ast.Toplevel.Builtin(loc, f)
    builtins[escaped]._type = types.T.Builtin(builtins[escaped])
end
```

PR #57, opened 2019-03-11
Parser entry points

hugomg:
This set of patches adds additional entry points to the parser. It can now parse just a single expression, type, or statement into AST form, instead of the whole program.

It simplifies parser_spec at the cost of making parser.lua slightly more complex. I thing it could be a worthy tradeoff but I am not 100% sure. What do you think?

### DIFF
--- a/pallene/driver.lua
+++ b/pallene/driver.lua
@@ -32,7 +32,7 @@ local function compile_pallene_to_ast(pallene_filename, stop_after)
     if not input then return false, {err} end
 
     local prog_ast
-    prog_ast, errs = parser.parse(pallene_filename, input)
+    prog_ast, errs = parser.parse_program(pallene_filename, input)
     if stop_after == "parser" or not prog_ast then return prog_ast, errs end
 
     local stop_i = step_index(ast_passes, stop_after)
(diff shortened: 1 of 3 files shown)


hugomg:
My feeling is that whether this is a good idea or not is more about the implementation complexity and code clarity tradeoffs than about testing. For example, one pro of the new version is that the test cases for expressions do not depend anymore on the ast structure for the toplevel.

I felt that the test script wasn't very affected, actually. The biggest difference is in "syntax error" tests. If the inner entry point fails without raising a label then the result is a default error message ("SyntaxError"). If the same code fragment is inserted in a larger program it may produce a different error message in that case. For example, in two test cases involving return statements that are not the last in the block, the message had to be changed fro "missing 'end' to close function" to "missing 'end' to close block". Similarly, in the test case for " integer.new()" the bad symtax had to be included inside an assignment statement (as it used to be before the code changes) to get a real error message from it.

mascarenhas:
I am fully onboard with the improvements to the parser code and to code that needs to reuse a part of the parser! I just worry about having reduced the test coverage due to a lot of the tests not exercising the integration between the different parts of the grammar, but if there are still "end-to-end" tests for the grammar my worries might be unfounded.

hugomg:
I think Fabio has convinced that this is probably not such a good idea after all. Let's keep the parser code simpler :)

PR #58, opened 2019-03-12
Add extra documentation on how to get started

hugomg:
To welcome our newest contributors :)


hugomg:
We can wait to merge this until after @mambrozio finished his first pull-request. This way if we remember of more things that need to be here we can add them.

PR #69, opened 2019-06-15
Tag tostring

### LINKED ISSUE #29: Should we create a custom "inspect" function that knows about type tags?
We can use a special parameter to the inspect function to make it use typedecl.tostring to show things like `_tag = "ast.Stat.Block"` instead of `_tag = <function 1>`.

* pros: more readable
* cons:  introduces an additional dependency to the codebase

```lua
local inspect = require "inspect"

local function process function(item, path)
    if path[#path] == "_tag" then
        return td.tostring(item)
    else
        return item
    end
end

inspect.inspect(prog_ast, {process=process})
```

See https://github.com/kikito/inspect.lua

hugomg:
Closes #29. Now we won't need to call typedecl.tostring to be able to read what
the tag for a value is. The output of the "inspect" function is easier to read
now, which should help debugging. It also makes our code a tad simpler and less
magic.

The current rationale for function tags is twofold:

1. Avoid having to type "" in the if-then-else statements

2. Allow renaming the modules when requiring them. For example:

    ```lua
    local aaast = require 'pallene
    if blah._tag == aaast.Exp.Bool then
    ```

In hindsight, I think these are not very compelling reasons anymore. For (1), I
think the negative of needing to use typedecl.tostring outweighs the positive of
not needing to type quotes. For (2), the tostring function will still output
"ast.Exp.Bool" even if we reimport the ast module under a different name, so the
suggestion that we can rename it at will like that is actively misleading.

----

Another alternative solution that I explored was to replace function _tags by
special objects with a __call and __tostring metamethod. This would have been
less disruptive in terms of lines-of-code changed, but it would only add more
complexity to the code (instead of removing it) and it would also make the
"inspect" output clunkier, since it would now print those metatables. This
solution also would not fix the problem 2) that I mentioned above


hugomg:
Opening this PR so that we can have a discussion about whether this is a good or bad idea after all.

I expect these changes to create a gazillion merge conflicts for everything else so if we do decide that this is a good thing to do, we should probably merge the rest of the stuff in the pipeline first, and then rebase this branch on top of all of that.

PR #70, opened 2019-06-21
Use udatamemoffset instead of getudatamem

### LINKED ISSUE #66: Use udatamemoffset instead of getudatamem
Currently our generated code for accessing fields in a C record calls the `getdataumem` macro. I think we might be able to gain a bit of performance if we instead called the `udatamemoffset` macro directly.

```C
/* compute the offset of the memory area of a userdata */
#define udatamemoffset(nuv) \
	((nuv) == 0 ? offsetof(Udata0, bindata)  \
                    : offsetof(Udata, uv) + (sizeof(UValue) * (nuv)))

/* get the address of the memory block inside 'Udata' */
-#define getudatamem(u)	(cast_charp(u) + udatamemoffset((u)->nuvalue))
```

The `getudatamem` macro reads the `->nuvalue` field at runtime. We know at compilation time what value this should be but GCC does not, so it ends up spitting out a bunch of ASM instructions to do the field read and the conditional branch.

I suspect that if we called `udatamemoffset` directly, passing a constant integer literal for the `nuv` parameter, then GCC should be able to compile the field access down to a single mov instruction.

gligneul:
By using the macro udatamemoffset directly, the C compiler can compute during
compiler time the offset of C fields in a userdata.
Even thought the compiler generates better code, I did not experience
performance improvements in my computer (Intel Code i5-7360U/Macbook Pro 2017).

Closes #66

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -800,10 +800,12 @@ end
 
 function RecordCoder:primitive_slot(udata, field_name)
     local out = util.render(
-        [[((${STRUCT_NAME} *)getudatamem(${UDATA}))->${FIELD_NAME}]],
+        [[((${STRUCT_NAME} *)(cast_charp(${UDATA}) +
+            udatamemoffset(${NUVALUE})))->${FIELD_NAME}]],
     {
         STRUCT_NAME = self:struct(),
         UDATA = udata,
+        NUVALUE = self.layout.gc_size,
         FIELD_NAME = self:field(field_name),
     })
     return out
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -725,6 +725,10 @@ function RecordCoder:struct()
     return string.format("struct crecord_%s", self.tl_node.name)
 end
 
+function RecordCoder:getmem()
+    return string.format("crecord_%s_getmem", self.tl_node.name)
+end
+
 function RecordCoder:index(field)
     return string.format("crecord_index_%s_%s", self.tl_node.name, field)
 end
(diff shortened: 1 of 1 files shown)


hugomg:
Great! I checked here and sure enough the ASM code really is much simpler. I think that is worth it by itself, even if there isn't an immediately obvious performance benefit.

The only problem I see is that the generated C code became much harder to follow. For example, a line like `local dx = a.x - b.x` gets compiled down to the following:

```
lua_Number x6 = ((struct crecord_Point *)(cast_charp(x1) +
udatamemoffset(0)))->f1 - ((struct crecord_Point *)(cast_charp(x2) +
udatamemoffset(0)))->f1;
```

What about creating a small helper macro next to the struct definition?

```c
struct crecord_Point {
    lua_Number f1 /* x */;
    lua_Number f2 /* y */;
};

#define crecord_field_Point(u, f) ((struct crecord_Point *)(cast_charp(u) + udatamemoffset(0)))->f
```

Something along these lines would allow the previous code to be compiled into something more easy to read:

```c
lua_Number x6 = crecord_field_Point(x1, f1) - crecord_field_Point(x2, f1);
```

pallene/coder.lua
```diff
@@ -800,10 +800,12 @@ end
 
 function RecordCoder:primitive_slot(udata, field_name)
     local out = util.render(
-        [[((${STRUCT_NAME} *)getudatamem(${UDATA}))->${FIELD_NAME}]],
+        [[((${STRUCT_NAME} *)(cast_charp(${UDATA}) +
+            udatamemoffset(${NUVALUE})))->${FIELD_NAME}]],
```
hugomg: The pretty.lua formatter unfortunately removes this helpful indentation from the generated C file.

gligneul:
Sure, I will create a inline function.

pallene/coder.lua
```diff
@@ -790,6 +794,22 @@ function RecordCoder:declare_struct()
...
+            return (${STRUCT} *)(cast_charp(u) + udatamemoffset(${NUVALUE}));
+        }
+    ]], {
+        STRUCT = self:struct(),
+        GETMEM = self:getmem(),
+        NUVALUE = self.layout.gc_size,
+    })
+    return out
+end
+
```
hugomg: I think it would be helpful to add a small comment pointing out that this is a specialization of getudatamem that generates cleaner assembly code because NUVALUE is a compile-time constant.

PR #86, opened 2019-07-07
Reintroduce missing function in pallenec

hugomg:
At some point we accidentally removed the pallenec_abort function. This caused
pallenec to crash instead of printing out a nice error message.


gligneul:
We should have tests for pallenec as well.

PR #90, opened 2019-07-14
Rearchitect typechecker and code generator

hugomg:
This is a work-in progress PR that rearchitects the compiler by introducing a new intermediate representation between the typechecker and the code generator.

The objective is to reduce our reliance on the "_xxx" ast-annotations, and to simplify the code generation by having it receive lower-level input that is closer to the target language (C).

Some of the more noticeable differences:
* There is now an ir.Module data structure that keeps track of all the "toplevel" information. Record types, functions, etc. This is an alternative to using ast.Toplevel nodes for everything.
* scope_analysis and typechecking have been merged into a single pass.
* There is a new pass that converts the typed AST to a list of function bodies.
* The code generation is rewritten from scratch.
* The passes have been rewritten in "object oriented style". This makes is easier to have mutually-recursive functions, and removes the need to explicitly pass a "ctx" object to every function call.

There is still a lot to be done to have the code generator pass all the tests:

- [x] Array types
- [x] Userdata representation of records
- [x] Record types
- [x] Strings
- [x] Coercions to and from `value`
- [x] Calling Lua functions
- [x] First-class Pallene functions
- [x] Module-level variables
- [x] Garbage collector can be called (variables are saved to Lua stack)
- [x] Garbage collector can be called by Pallene (inside loops)
- [x] tofloat builtin
- [x] io.write builtin
- [x] array builtins
- [x] Check that there are no luacheck warnings
- [x] Add test for "tofloat" shadowing
- [x] Find out what is the deal with the -Wmaybe-uninitialized warnings.
- [x] Check if everything still works with Clang. Does it also complain about uninitialized TValues?
- [ ] Decide the final fate of all the _xxx annotations
- [x] Decide what to do with the `self.func` problem
- [x] Determine if GCC can avoid loading the "G" userdata in leaf functions that don't use it.


hugomg:
This change log documents the biggest changes I made, and the questions that I am specially looking to get feedback about.

**Q:** Let me know if you think some of the points here should be turned into comments inside the source code. I think the stuff that talks about the old coder only makes sense as part of this PR but maybe there are some things that only reference the new coder that deserve to be kept more permanently.

## All passes have been rewritten in OO-style

The old version of the compiler used sets of mutually-recursive local functions, while the new version uses methods called with `self:foo()` notation. The things I like about the new version are:

* Passing around a `self` is more readable than passing around a `ctx` or whatever we used to call it.
* We can write down the functions in any order, without forward declarations
* We don't have to nest big functions inside each other.

**Q:** Should we add a class-creation helper function to util.lua?

**Q:** Each pass instantiates and runs the root object in a different way. Can we change them to be consistent?

**Q:** What is the best way to deal with the "current function"? Currently, before our passes enter a function they set `self.func` and then they unset it after they are done. Should we create a second class for the methods that care about the current function? Should we have a "with_func" method like that one in the symbol table? Or perhaps just pass it around as a parameter?

## Scope analysis is now a part of the typechecking pass

Now the toplevel information like functions, records and so on are stored in an ir.Module object instead of being haphazardly encoded as annotations in the parse tree.

The main motivation for merging scope analysis and typechecking is that the entries in the ir.Module structure are typed names. So instead of adding the names in one pass and the types in a second, we can add both together in a single pass.

The new design allowed me to get rid of some of the `_decl` and `_type` annotations, but we still need to keep some annotations for the to_ir pass

* `_decl` annotations on ast.Var.Name nodes have been replaced by `_name` annotations. Instead of pointing to an AST node, they now are values of the checker.Name algebraic data type.  The meaning of the name is now explicit, instead of being implicit from the type of ast node it is pointing to. It also leads to cleaner code in the latter passes when they need to tell apart if something is a local variable, a global variable, or a toplevel funtion name.

* `_type` annotations remain in ast.Var and ast.Exp nodes. The `_type` annotations that went on function, type, and variable declarations now live in ir.Module and ir.Function.

## Lambda expressions

There is now an ast.Exp.Lambda node. Currently this is only an internal change, done because the type inference prefers it this way.

The language still does not support nested functions because neither the parser nor the subsequent passes allow it. But the Lambda nodes do help that...

## New to_ir pass between typechecking and code generation

The new IR is described in ir.Lua. The most important thing is that there are no nested subexpressions, so the coder doesn't need to recursively call itself everywhere, and it can also know ahead of time what are all the C variables that it is going to need to declare.

There is still nesting at the level of the control structures (loops and if statements). I think this is going to be easier to optimize than a fully generic graph of labels and jumps. I also have separate nodes for for-loops and while-loops to allow for some dumb optimizations on top of for-loops.

The left-hand side of the rules is always a variable name called "dst". The right hand sides are ir.Value, which can be either a variable name or a compile-time constant. This lets the generated C say things like `x2 = x1 + 1`; instead of `x3 = 1; x2 = x1 + x3`

The field names of the left-hand and right-hand sides are somewhat standardized, so the get_srcs function doesn't need to have a separate case for everything.

One thing I was very careful about with the to_ir algorithm was that I wanted assignments like `foo = bar` to compile down to a single `x2 = x1` instruction, without using intermediate variables like `x3 = x1; x2 = x3`. This was one of the reasons that I decided to keep the to_ir pass separate from the typechecker. When we are typechecking `local x = e` we need to typecheck `e` before we can infer the type of `x`, but when we are generating IR we want to know the type of `x` first so that when we generate the code for `e` we can tell where to assign the result to. Similarly, when we call a toplevel function `f(x)` we want to directly spit out a `CallStatic` node, without first assigning `f` to an intermediate function pointer variable.

**Q:** I haven't started implementing optimizations yet. I don't know if the IR will behave well in that aspect. Will adding, removing or moving Cmd nodes be pleasant? What about getting rid of unused variables?

## Changes to the C reformatting algorithm

* The old algorithm removed blank lines inside functions. The new algorithm preserves them with the same behavior as in the toplevel (multiple blanks in a row are merged in a single one)
* Special indentation rule for goto labels (`done:`)
* New C.comment function for generating single-line comments that don't confuse the indentation algorithm.

## Completely rewritten coder.lua

The coder.lua has been rewritten from scratch. The most important change is that now it has an analysis phase at the start and then when it calls `self:generate_module()` it stops mutating internal state. In particular it does not allocate new variables or keep track of a stack of stack slots as it goes along.

The Pallene functions now receive the upvalues table as an additional argument (Udata *G). This way we don't need to fetch it from the C-closure upvalue over and over if we actually do need it. (We could consider an optimization that removes this parameter from "leaf" Pallene functions that do not need it)

In the old coder, the "generation context" was used for not only the Pallene functions but also the luaopen function, the Lua entry points, and for the __index and __newindex record metatable. In the new coder the "GC-aware" infrastructure is only for the Pallene functions.

* The luaopen function now uses regular C-API functions. This way the code generation doesn't need to know how to implement many of the operations that happen here (such as allocating new string literals and metatables). The evaluation of the initializers for the toplevel variables now go in a separate $init function, which is a regular Pallene function.
* The Lua entry points don't need to care about garbage collection because everything they work with is firmly rooted in the Lua stack.
* The __index and __newindex functions have been removed. In the future they will be replaced by a simple __index table.

There is no more distinction between tvar and cvars. The Pallene variables are referred to by their numeric identifiers. And when a function needs to use a temporary C variable, it just gives it a fixed name instead of a generated one. For example, `TValue *slot` in ArrGet and ArrSet. (This works because we don't nest expressions anymore). The only catch was in the for loops -- I didn't want to turn off the shadowing warning so I have to generate fresh names for the loop variables.

To improve readability, some of the bigger chunks of code that the coder used to generate have been moved to macros or inline functions in pallene_runtime.h and pallene_runtime.c. Notably, the barrierback operations and the integer division, modulo, and shift operations.

**Q:** The stuff surrounding the toplevel $init function feels a bit hacky. For example, there is code that assumes that $init's ID is always 1, and in the "ast.Toplevel.Var" case of the checker it does a strange check_var call instead of a check_stat one.

## TODO

The main feature still left to implement is the integration with the garbage collector.

The other feature left is support for calling Lua functions (and first-class Pallene functions). This one depends on implementing the GC integration first because the Lua functions might call the GC or relocate the stack.

gligneul:
> Q: Should we add a class-creation helper function to util.lua?

Yes, sure.

> Q: Each pass instantiates and runs the root object in a different way. Can we change them to be consistent?

Not really sure what you meant, but I don't think that we need this kind of consistency.

> Q: What is the best way to deal with the "current function"? Currently, before our passes enter a function they set self.func and then they unset it after they are done. Should we create a second class for the methods that care about the current function? Should we have a "with_func" method like that one in the symbol table? Or perhaps just pass it around as a parameter?

Probably create another class given that we are going to the OO path. I would rather avoid relying on mutable global variables (to me, a field in such classes fits in this case). In particular, this will be an issue when introduce nested functions.

---

> The new design allowed me to get rid of some of the _decl and _type annotations, but we still need to keep some annotations for the to_ir pass

Given that we are refactoring everything, I would rather move these annotations to another table which is indexed by the AST node. So, instead of `node._decl` we would have `decls[node]`.

---

> Lambda expressions

Nice. I would call them prototypes to be consistent with the internal Lua nomenclature. But Lambda is also fine.

---

> There is still nesting at the level of the control structures (loops and if statements). I think this is going to be easier to optimize than a fully generic graph of labels and jumps. I also have separate nodes for for-loops and while-loops to allow for some dumb optimizations on top of for-loops.

Only with we perform poor optimizations. This is fine for now, but eventually we should change this IR to use SSA so we can implement more powerful optimizations that don't rely on specialised loop statements.

> Q: I haven't started implementing optimizations yet. I don't know if the IR will behave well in that aspect. Will adding, removing or moving Cmd nodes be pleasant? What about getting rid of unused variables?

I guess this question is not for me. Anyway, moving nodes around should be a requirement for the IR that will be optimized.

---

> The Pallene functions now receive the upvalues table as an additional argument (Udata *G). This way we don't need to fetch it from the C-closure upvalue over and over if we actually do need it. (We could consider an optimization that removes this parameter from "leaf" Pallene functions that do not need it)

Most likely, gcc inlines everything and removes the unused parameter; hence, this optimization should be low-priority.

---

> The Lua entry points don't need to care about garbage collection because everything they work with is firmly rooted in the Lua stack.

Ins't the overhead noticeable in this case?

> Q: The stuff surrounding the toplevel $init function feels a bit hacky. For example, there is code that assumes that $init's ID is always 1, and in the "ast.Toplevel.Var" case of the checker it does a strange check_var call instead of a check_stat one.

I guess we can't do much unless we manually create some IR nodes for the toplevel. It would also be a hack, but that is the nature of the current toplevel. Nonetheless, it would be better to perform this hack in an earlier step to keep the code dumber.

hugomg:
>  Not really sure what you meant, but I don't think that we need this kind of consistency.

I am talking specifically about `checker.check()`, `to_ir.convert()` and `coder.generate()`. Take a look at how they create and invoke the Checker, ToIR and Coder objects.

I think this has to do with the `self.func` problem. Notice how to_ir passes func as a parameter but the checker and coder use that "global variable" pattern...

> Probably create another class given that we are going to the OO path

Would the child function-specific checker/coder carry a pointer to the toplevel object? Is it OK if all the things that we currently write `self.foo` and `self:bar()` become `self.p:foo` and `self.p:bar()`?

> Given that we are refactoring everything, I would rather move these annotations to another table which is indexed by the AST node.

We should discuss these on a case-by-case basis before we finish merging this PR.

>  but eventually we should change this IR to use SSA so we can implement more powerful optimizations that don't rely on specialized loop statements.

I think this has more to do with how we represent the control flow graph than with whether we use SSA or not.

The impression I have is that the main gain of SSA is for performance, but it comes at a cost of being a more complicated implementation. My current plan is to do all those optimizations in a non-SSA representation with dataflow analysis and use-def chains. (With Appel's compilers in ML book as the main reference)

> Most likely, gcc inlines everything and removes the unused parameter; hence, this optimization should be low-priority.

Good point, I should check if GCC can optimize the lua entry point for these leaf functions.

That said, the main reason I don't want to do this optimization is because what I really would love to be able to do is to statically allocate all the stuff that currently lives in G.

> Isn't the overhead noticeable in this case?

I only switched to using the  C-API in the luaopen function. Luaopen allocates some things (the closures list, the global userdata object, and the exports table) and I want to be sure that they don't get destroyed by the GC if the $init function causes it to run.

The Lua entry points still bypass the  C-API, just as before. They don't use the same GC-aware" "generation context" machinery as the Pallene functions (which is what we did in the old coder) but that is OK because there is nothing for the GC to destroy in this case.

The only difference between the old Lua entry point and the new one is that now we fetch the "G" upvalue in the lua entry point inside of doing it as the first statement in the Pallene function.

gligneul:
> I think this has to do with the self.func problem. Notice how to_ir passes func as a parameter but the checker and coder use that "global variable" pattern...

I see, I prefer passing it as a parameter.

gligneul:
> I think this has to do with the self.func problem. Notice how to_ir passes func as a parameter but the checker and coder use that "global variable" pattern...

I see, I prefer passing it as a parameter/creating another object.

> Would the child function-specific checker/coder carry a pointer to the toplevel object?

Yes, is there another option?

> Is it OK if all the things that we currently write self.foo and self:bar() become self.p:foo and self.p:bar()?

It is ok. Most likely some methods and fields will be under the new function object, not the global one.

> We should discuss these on a case-by-case basis before we finish merging this PR.

Ok, but I don't see a case where this is a problem, unless we are actually modifying the AST instead of just adding meta fields.

> The impression I have is that the main gain of SSA is for performance, but it comes at a cost of being a more complicated implementation. My current plan is to do all those optimizations in a non-SSA representation with dataflow analysis and use-def chains. (With Appel's compilers in ML book as the main reference)

Do you mean the compiler performance? That shouldn't be an issue for an AOT compiler. AFAIK, the only complicated part of SSA is converting from the register form to the SSA form. After that, it is much simpler to implement the optimizations in the SSA form.

> The Lua entry points still bypass the C-API, just as before. 

Oh, ok I miss understood.

hugomg:
> Yes, is there another option?

I think the two options are to have a separate function coder object (which would imply the self.p.foo idiom), or to pass the self.func as a parameter.

> unless we are actually modifying the AST instead of just adding meta fields.

The new checker modifies the AST by inserting explicit type coercion nodes and by adding the _name and _type annotations

In this aspect its similar to the old checker. I only changed what annotations it adds.

> Do you mean the compiler performance? That shouldn't be an issue for an AOT compiler. 

Yes, I was talking about the performance of the compiler itself.

> AFAIK, the only complicated part of SSA is converting from the register form to the SSA form. 

Perhaps. But I also am less comfortable with programming with SSA. :)

gligneul:
> I think the two options are to have a separate function coder object (which would imply the self.p.foo idiom), or to pass the self.func as a parameter.

Both are fine. I think this issue is a stylistic preference at this point.

> The new checker modifies the AST by inserting explicit type coercion nodes and by adding the _name and _type annotations

I don't mind adding new nodes to the AST. My issue is mutating the nodes with underscore fields.

> Perhaps. But I also am less comfortable with programming with SSA. :)

We could change to SSA later. The IR that you proposed already brings a lot of benefits.

hugomg:
After inspecting the generated assembly, it seems that for small functions that GCC is able top optimize away the "G" userdata for small functions. The pallene entry point is inlined inside the Lua entry point if G is not used then GCC optimizes away the operations for reading the first function upvalue.

I don't know how this holds up in more complicated examples. However, the "tiny leaf function" that GCC can optimize by itself is probably the most important one anyway. Larger functions are more likely to need G anyway and even if they don't the overhead of producing G will be a smaller percentage.

I therefore conclude that we can probably keep it like we are doing right now. The Lua entry points always compute G, the pallene functions always receive it as an argument, and GCC is free to optimize it if it can.

gligneul:
I would add this paragraph in the code for documentation. I think it is important to document these kind of conclusions inside the code so we don't forget it.

hugomg:
I have been thinking about the _name and _type annotations and I think I would prefer keeping them as they are right now. `exp._type` is very easy to read, and doesn't require us to pass around a bunch of `types` and `names` tables to the to_ir step. Additionally the "damage" of these annotations is contained only to "to_ir.lua". It doesn't go all the way to the coder (which is where things used to get really confusing before).

Right now, if you put a gun to my head and told me I had to get rid of those underscores at all costs, I think that instead of having the checker put the annotations in a separate table I would prefer creating a new datatype in `typed_ast.lua` which is exactly equal to `ast.lua` but where all expressions have  a `type` field, and all the names are `checker.Name` instead of strings. But that sounds like a lot of busywork compared to just adding some harmless _name and type annotations.

What do you think?

gligneul:
> What do you think?

We could leave it like that for now.

I liked the idea of a `typed_ast.lua`, but it introduces a little bit of boilerplate. We could generate this file with metaprogramming.

PR #98, opened 2019-09-21
Implement constant propagation of globals

hugomg:
This commit implements an optimization that removes module variables that are
constant or that are never read from. `GetGlobal` reads from constant variables
removed this way are replaced by a `Move` of the corresponding value.

For example, in the following code the N variable is optimized away, as if we
had written `for i = 1, 10`:

    local N = 10
    function f()
        for i = 1, N do
            blah()
        end
    end

Furthermore, in the following code the `x` variable is not created in the final
version of the code, since it is never read from. The `blah()` function still
gets called inside `f`, but its return value is thrown away.

    local x = 42
    function f()
        x = blah()
    end

The important limitation for now is that a module variable is only identified as
a constant if its initializer is a literal value (number, string, etc). This
means that in the following program `x` and `y` are NOT considered constant.

    local x = 1 + 1
    local x = y

### DIFF
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -1,0 +1,131 @@
+local ir = require "pallene.ir"
+
+local constant_propagation = {}
+
+local function is_constant_value(v)
+    local tag = v._tag
+    if     tag == "ir.Value.Nil"      then return true
+    elseif tag == "ir.Value.Bool"     then return true
+    elseif tag == "ir.Value.Integer"  then return true
+    elseif tag == "ir.Value.Float"    then return true
+    elseif tag == "ir.Value.String"   then return true
+    elseif tag == "ir.Value.LocalVar" then return false
+    elseif tag == "ir.Value.Function" then return true
+    else
+        error("impossible")
+    end
+end
+
+-- Replaces toplevel constant variables by their respetive values.
+--
+-- Currently assumes that the toplevel constant variable is initialized with a
+-- constant literal. Does not currently recognize non-trivial constant
+-- expressions as being constant.
+function constant_propagation.run(module)
+
+    local n_globals = #module.globals
+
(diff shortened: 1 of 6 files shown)


pallene/constant_propagation.lua
```diff
@@ -0,0 +1,131 @@
...
+
+    local n_reads  = {} -- { g_id => int }
+    local n_writes = {} -- { g_id => int }
+    for i = 1, n_globals do
+        n_reads[i]  = 0
+        n_writes[i] = 0
+    end
+
+    for _, func in ipairs(module.functions) do
+        for _, cmd in ipairs(ir.flatten_cmd(func.body)) do
```
gligneul: It would be nicer if we had an iterator for all commands instead of creating a new list and than using ipairs.
We could even keep this implementation, but hide it under a function.
```
for _, cmd in ir.flatten(func.body) do
```

pallene/constant_propagation.lua
```diff
@@ -0,0 +1,131 @@
...
+                    return ir.Cmd.Nop()
+                end
+            else
+                -- don't modify
+                return false
+            end
+        end)
+    end
+
+    -- 5) Done
```
gligneul: I would remove this one :)

pallene/constant_propagation.lua
hugomg: Good idea. There is now an ir.iter function replacing all uses of ir.flatten_cmd

I didn't delete the flatten_cmd because it is still used in one of my unpublished branches.

pallene/constant_propagation.lua
hugomg: Can I keep the `5)`? My intention was to clarify that the `ir.clean_all(module)` does not belong under section `4)`

PR #100, opened 2019-09-25
Introduce type alias in the top level

gligneul:


doc/manual.md
```diff
@@ -187,9 +187,23 @@ end
...
+### Type aliases
 
-Currently, type declarations consist solely of record declarations.
+Creates an alias for a previously-declared type with the following syntax:
+
+```
+type <name> = <type>
+```
+
+Type alias in Pallene cannot be used to declare recursive types, such as:
```
hugomg: Type aliases cannot ==> Type aliases currently cannot

PR #104, opened 2019-09-30
Implement uninitialized variables

### LINKED ISSUE #95: Allow variable declarations without initializers
For example,

```
local x: integer
if a < b then
    x = 17
else
    x = 42
end
```

The type annotation is mandatory if a variable is declared without an initializer expression.

We will have to implement a control-flow-based verification pass to ensure that variables are always initialized before they are used. (I think use-def chains would do the trick).

At the same time we do this we should remove the "always_returns" verification from the checker and transform it into a "return variables are initialized before use" check. This will also allow the return checks to be more robust in the presence of break, while true, and so on. We should also make sure that we reflect this in additional test cases.

hugomg:
This PR adds the option to declare uninitialized variables. Closes #95

    local x: integer
    if foo then
        x = 10
    else
        x = 20
    end
    return x

Instead of using a liveliness analysis pass, this patch builds upon the "always_returns" check that we used to have in checker.lua. It now works at the IR level and can also detect variables that are used before they are initialized.

Additionally, the new algorithm is smarter than the old one when analyzing loops. `while true` and `repeat until false` loops that don't contain a break statement no longer need to be followed by a return statement.

    function foo(): integer
        while true
            -- blablala
        end
        -- the compiler now knows this comment
        -- down here is unreachable
    end

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -195,7 +195,7 @@ Creates an alias for a previously-declared type with the following syntax:
 type <name> = <type>
 ```
 
-Type alias in Pallene curretly cannot be used to declare recursive types, such as:
+Type alias in Pallene currently cannot be used to declare recursive types, such as:
 
 ```
 type T = {T}
@@ -272,6 +272,15 @@ Unlike languages like C or Java, Pallene does not require type annotations on ev
 It uses a bidirectional type-checking system that is able to infer the types of almost all variables and expressions.
 Roughly speaking, you must include type annotations for the parameters and return types of toplevel functions, and almost everything else can be inferred from that.
 For example, notice how the `sum_floats` from the Brief Overview section does not include a type annotation for the `result` and `i` variables.
+
+If a local variable declaration doesn't have an initializer, it must have a type annotation
+```
+function contrived(): integer
+    local x:integer
+    x = 10
+    return x
+end
+```
 
 ### Automatic type coercions
 
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -273,7 +273,7 @@ It uses a bidirectional type-checking system that is able to infer the types of 
 Roughly speaking, you must include type annotations for the parameters and return types of toplevel functions, and almost everything else can be inferred from that.
 For example, notice how the `sum_floats` from the Brief Overview section does not include a type annotation for the `result` and `i` variables.
 
-If a local variable declaration doesn't have an initializer, it must have a type annotation
+If a local variable declaration doesn't have an initializer, it must have a type annotation:
 ```
 function contrived(): integer
     local x:integer
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -839,7 +839,6 @@ function FunChecker:check_initializer_exp(decl, exp, err_fmt, ...)
                 "uninitialized variable '%s' needs a type annotation",
                 decl.name))
         end
-
     end
 end
 
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -162,38 +162,6 @@ function Checker:from_ast_type(ast_typ)
     end
 end
 
--- Does this statement always call "return"?
-local function stat_always_returns(stat)
-    local tag = stat._tag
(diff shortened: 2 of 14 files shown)


pallene/checker.lua
```diff
@@ -866,12 +825,24 @@ function FunChecker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
...
-        return self:check_exp_synthesize(exp)
+        if exp then
+            local e = self:check_exp_synthesize(exp)
+            return e._type, e
+        else
+            type_error(decl.loc, string.format(
+                "uninitialized variable '%s' needs a type annotation",
+                decl.name))
+        end
+
```
gligneul:
```suggestion
```

doc/manual.md
```diff
@@ -273,6 +273,15 @@ It uses a bidirectional type-checking system that is able to infer the types of
 Roughly speaking, you must include type annotations for the parameters and return types of toplevel functions, and almost everything else can be inferred from that.
 For example, notice how the `sum_floats` from the Brief Overview section does not include a type annotation for the `result` and `i` variables.
 
+If a local variable declaration doesn't have an initializer, it must have a type annotation
```
gligneul:
```suggestion
If a local variable declaration doesn't have an initializer, it must have a type annotation:
```

spec/uninitialized_spec.lua
```diff
@@ -0,0 +1,57 @@
...
+    it("catches use of uninitialized variable", function()
+        assert_error([[
+            function foo(): integer
+                local x:integer
+                return x
+            end
+        ]], "variable x is used before being initialized")
+    end)
+end)
+
```
gligneul:
```suggestion
```

spec/checker_spec.lua
```diff
@@ -444,44 +454,6 @@ describe("Pallene type checker", function()
...
-                if a == 1 then
-                    return 10
-                elseif a == 2 then
-                else
-                    return 30
-                end
-            end
-        ]],
-        [[
-            function getval(a:integer): integer
```
gligneul: We could keep this test case.

pallene/uninitialized.lua
```diff
@@ -0,0 +1,184 @@
+local ir = require "pallene.ir"
+local location = require "pallene.location"
+
+local uninitialized = {}
+
+-- B <- A
+local function copy(A)
```
gligneul: These functions should go to util.lua.

pallene/uninitialized.lua
```diff
@@ -0,0 +1,184 @@
...
+--
+-- If the execution can possibly fall through to the next command, returns
+-- `true` and the updated set of uninitialized variables. Returns false if
+-- the execution unconditionally jumps or gets stuck in an infinite loop.
+--
+local function test(cmd, uninit, loop)
+
+    local function check_use(v)
+        if uninit[v] then
+            coroutine.yield({ v = v, cmd = cmd})
```
gligneul:
```suggestion
            coroutine.yield({v = v, cmd = cmd})
```

spec/uninitialized_spec.lua
```diff
@@ -0,0 +1,54 @@
...
+            function getval(a:integer): integer
+                if a == 1 then
+                    return 10
+                elseif a == 2 then
+                else
+                    return 30
+                end
+            end
+        ]], missing_return)
+    end)
```
gligneul: I think we should a test case similar to the following:
```
function foo(cond: boolean): integer
  local x: integer
  while cond do
    x = 0
    cond = x == 0
  end
  return x
end
```

pallene/uninitialized.lua
hugomg: I'd rather not change the merge function, as it is specialized to this algorithm. For example, a previous version of the algorithm use initialized sets instead of uninitialized sets and merge was an intersection instead of an union.

spec/uninitialized_spec.lua
hugomg: Good idea

Issue #89, opened 2019-07-14
Update our bundled version of Lua 5.4


hugomg:
There have been some new work releases since then.

gligneul:
This will fix the following bug:
```
function main(): nil
    local x: {value} = {}
    return x[10]
end

./pallenec test.pln && ./lua/src/lua -l test -e "print(test.main())"

./lua/src/lua: (command line):1: line 3: wrong type for downcasted value, expected nil but found nil
```

hugomg:
I started investigating this issue.

The first thing that broke is that the `Table` struct no longer has a "sizearray" field. Now there is something else called `alimit`. I need to check with Roberto how this affects us.

https://github.com/lua/lua/blob/9e6807c3c9d5036e999f636f936df07b72284442/lobject.h#L675  
https://github.com/lua/lua/blob/9e6807c3c9d5036e999f636f936df07b72284442/ltable.c#L195

PR #106, opened 2019-10-03
Lua 5.4 beta

hugomg:
Updates our version of Lua to the 5.4-beta version that was just released today.

This does not close issue #89 yet, because the "isempty" issue did not fix itself with the update, like we expected it would. I'll need to make some changes to the Pallene compiler but I think it will be clearer if I do that in a separate PR after this one.


gligneul:
Please add a test case for that particular bug when you fix it.

PR #107, opened 2019-10-03
Allow nils in arrays

### LINKED ISSUE #89: Update our bundled version of Lua 5.4

hugomg:
This PR has two changes

The first one is that now we allow arrays of nil, aka {nil}.

I believe that the original motivation for this restriction was because it was
impossible to tell apart in "in-bound" nil from an "out of bound" nil.
However,we have since changed our interpretation of Pallene to not have a
separate "array out of bounds" error and and treat "out of bounds" nils no
differently than an in-bound nil: both are just a run-time tag errors. Under
this interpretation there is no need to forbid {nil}.

Furthermore, even if we wanted to keep the concept of "out of bounds" there
would be no need to forbid {nil} anymore. The latest version of the Lua
interpreter uses a variant LUA_TEMPTY tag to distinguish the regular nils from
the nils that live out of bounds or in array holes.

This brings us to the second change in this patch. Under the previous version of
Pallene, we would return an "empty" value when reading from an out-of-bounds
index. This would result in strange behavior because this "empty" value is not
considered to be == nil !

    function get(xs: {value}, i:integer): value
        return xs[i]
    end

    -- Lua
    local x = test.get({}, 1)
    print(x)        --> nil
    print(nil == x) --> false (!)

This patch corrects this by ensuring that we always clear the LUA_TEMPTY tag
when we read a nil from an array. At first glance it might sound wrong to have
to do a tag check when reading from an `{value}` but rawget also has to do it so
I think it is unavoidable.

Closes #89

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1035,20 +1035,34 @@ gen_cmd["GetArr"] = function(self, cmd, _func)
     local i   = self:c_value(cmd.src_i)
     local dst_typ = cmd.dst_typ
     local line = C.integer(cmd.loc.line)
-    return (util.render([[
-        {
-            pallene_renormalize_array(L, $arr, $i, $line);
-            TValue *slot = &$arr->array[$i - 1];
-            ${check_tag}
-            $dst = $get_slot;
-        } ]], {
-            dst = dst,
-            arr = arr,
-            i = i,
-            line = line,
-            check_tag = self:check_tag(dst_typ, "slot", cmd.loc, "array element"),
-            get_slot = get_slot(dst_typ, "slot"),
-        }))
+
+    local parts = {}
+    table.insert(parts, "{")
+
+    table.insert(parts, util.render([[
+        pallene_renormalize_array(L, $arr, $i, $line);
+        TValue *slot = &$arr->array[$i - 1];]], {
+            arr = arr, i = i, line = line
+    }))
+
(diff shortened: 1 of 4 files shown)


pallene/coder.lua
```diff
@@ -1035,20 +1035,34 @@ gen_cmd["GetArr"] = function(self, cmd, _func)
...
+    table.insert(parts, "{")
+
+    table.insert(parts, util.render([[
+        pallene_renormalize_array(L, $arr, $i, $line);
+        TValue *slot = &$arr->array[$i - 1];]], {
+            arr = arr, i = i, line = line
+    }))
+
+    local check_tag = self:check_tag(dst_typ, "slot", cmd.loc, "array element")
+    if check_tag ~= "" then table.insert(parts, check_tag) end
```
gligneul: I wouldn't bother with this check. Also, I preferred the previous style with a single util.render.

hugomg: Doing it this way has one less `\n` in the end, which makes the generated code look better. (Maybe I should rethink the current strategy of signigicant '\n's)

gligneul: I thought we removed empty lines.

hugomg: I changed that a while ago, now it only merges the empty lines down to a single one.
I'll open an issue to continue the discussion.

PR #109, opened 2019-10-04
Check Lua major+minor version at runtime

### LINKED ISSUE #73: Check whether the Lua version is compatible with the Pallene module
A Pallene module should be loaded by a particular release of Lua.
We should add a check when the Pallene module is loaded to guarantee that the running version of Lua is the same one that the module was compiled for.
Notice that this should possible include the release (e.g. `5.4.0`), not just the version (e.g. `5.4`).

hugomg:
It is possible to check at runtime if we are running Lua 5.4, but it is not
possible to check for a minor release (like 5.4.0 or 5.4.1)

Closes #73

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1457,6 +1457,8 @@ function Coder:generate_luaopen_function()
     return (util.render([[
         int ${name}(lua_State *L)
         {
+            luaL_checkversion(L);
+
             lua_newuserdatauv(L, 0, $n_upvalues);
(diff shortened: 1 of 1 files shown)


gligneul:
Does it actually close the issue, given that we can't check the release number?

hugomg:
Verifying the release number would require patching the Lua lib. I'd rather not leave an issue open if it requires something that we are very unlikely to do.

If it turns out to be a problem we can always reopen it later.

PR #113, opened 2019-10-09
Change how we treat blank lines in the generated C

### LINKED ISSUE #110: Newlines in the generated code
In the first version of the C formatter for generated code, we would ignore blank lines inside C functions. Later, I changed the behavior to merge blank lines instead of removing them outright. This lets us use blank lines for formatting. For example, I like how the current luaopen function separates blocks of initializers with an empty line
  
    lua_pushstring(L, "check_nil");
    lua_geti(L, closures, 2);
    lua_settable(L, export_table);
    
    lua_pushstring(L, "check_boolean");
    lua_geti(L, closures, 3);
    lua_settable(L, export_table);
    
The downside is that the coder.lua has to be very careful about not generating a \n when it doesn't want to create a blank line. For example, there are places where we put the `]]` of a string literal on the same line instead of on a line of its own and there are times where we avoid adding empty strings to the `parts` array because the `table.concat(parts, "\n")` would add an extra `\n`.

Another option we could go with is to come up with a special sequence to denote the empty lines that we actually want to show up. For example, we could have lines with a `/*BLANK*/` comment be rendered as a blank line, and other blank lines be ignored. Or we could change the heuristic to collapse two blank lines into a single blank line but skip a blank line if it appears alone. In either of these cases the C formatting function would cease to be indempotent, but perhaps that is a better than having to deal with finnicky \n in the coder.lua?

hugomg:
Closes #110

The algorithm now preserves toplevel blank lines and blank lines containing tab characters. Clusters of consecutive blank lines are merged into a single blank line.

Empty `/**/` comments appearing by themselves are treated as if they were a tab character. It is hard to type a `\t` inside a [[ ]] string...

### DIFF
--- a/pallene/C.lua
+++ b/pallene/C.lua
@@ -64,24 +64,26 @@ end
 --   * /**/-style comments must not span multiple lines
 --   * Be careful about special characters inside strings and comments
 --   * goto labels must appear on a line by themselves
+--   * Use spaces for indentation instead of tabs
 --
 function C.reformat(input)
     local out = {}
     local depth = 0
     local previous_is_blank = true
     for line in input:gmatch("([^\n]*)") do
-        line = line:match("^[ \t]*(.-)[ \t]*$")
+        line = line:match("^ *(.-) *$")
 
-        -- Empty comments mark lines that we intentionally want to keep a blank
-        -- inside a function. (Note: this feature means that the reformat
-        -- function is not indempotent).
-        local intentional_blank = (line == "/**/")
-        if intentional_blank then
-            line = ""
+        -- We use tab characters to mark blank lines that should be preserved
+        -- in the output, for presentation purposes. (This trick means that the
+        -- reformat fucntion is still indempotent). However, typing a \t inside
+        -- [[ ]] strings is hard so we also use /**/ as a blank line marker.
+        if line == "/**/" then
+            line = "\t"
         end
 
-        -- Merge adjacent blank lines.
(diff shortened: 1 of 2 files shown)


pallene/C.lua
```diff
@@ -64,58 +64,75 @@ end
...
+            if     n_open > n_close then
+                depth = depth + 1
+            elseif n_open < n_close then
+                depth = depth - 1
             end
 
-            table.insert(out, string.rep(" ", nspaces))
-            table.insert(out, line)
-            table.insert(out, "\n")
+            if depth < 0 then
```
gligneul: Couldn't we add an assert here?

pallene/coder.lua
```diff
@@ -308,7 +310,7 @@ function Coder:pallene_entry_point_declaration(f_id)
     return (util.render([[
         static ${ret_type} ${name}(
             ${args}
-        )]], {
+        )]], { -- don't leave whitespace between ")" and upcoming ";"
```
gligneul: I didn't understand this comment.

hugomg: If we put the `]]` on the next line, our function declarations would end up looking like the following, with the `)` and `;` on separate lines

```
static void function_01(
    lua_State *L,
    Udata *G,
    StackValue *base
)
;
```

pallene/C.lua
hugomg: Good idea

PR #121, opened 2019-11-13
Serialize floats using decimal notation

### LINKED ISSUE #116: We should internally represent floating point literals as strings
This would let us avoid those hexadecimal literals in the output. Which is something not present in C90

hugomg:
Closes #116


gligneul:
What is the benefit of doing the code this way? It seems more convoluted.
Aren't we losing precision this way? What will happen if we support hexa floats in Pallene?

PR #125, opened 2019-11-25
Add sqrt builtin function

hugomg:
I'm going to be using it on some benchmarks.

The diff is easier to read if you only look at the first commit (7316848 ).

### DIFF
--- a/pallene/builtins.lua
+++ b/pallene/builtins.lua
@@ -4,6 +4,7 @@ local builtins = {}
 
 for lua_name, typ in pairs({
     ["io.write"]     = types.T.Function({types.T.String()}, {}),
+    ["sqrt"]         = types.T.Function({types.T.Float()}, {types.T.Float()}),
     ["tofloat"]      = types.T.Function({types.T.Integer()}, {types.T.Float()}),
 }) do
     local pallene_name = string.gsub(lua_name, "%.", "_")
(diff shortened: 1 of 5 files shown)


pallene/builtins.lua
```diff
@@ -4,6 +4,7 @@ local builtins = {}
 
 for lua_name, typ in pairs({
     ["io.write"]     = types.T.Function({types.T.String()}, {}),
+    ["sqrt"]         = types.T.Function({types.T.Float()}, {types.T.Float()}),
```
gligneul: Could we rename it to `math.sqrt`?

spec/coder_spec.lua
```diff
@@ -1082,6 +1082,30 @@ describe("Pallene coder /", function()
...
+    describe("sqrt builtin", function()
+        setup(compile([[
+            function square_root(x: float): float
+                return sqrt(x)
+            end
+        ]]))
+
+        it("works on positive numbers", function()
+            run_test([[
+                assert(1.0 == test.square_root(1.0))
```
gligneul: It might be nice to test some corner cases such as nan, inf and negative numbers.

pallene/builtins.lua
hugomg: Good idea. Is more consistent with io.write.

spec/coder_spec.lua
hugomg: Added NaN and inf. We already had tests for negative numbers.

PR #128, opened 2019-12-02
Fix floating point literals that are also integers

hugomg:
After I changed the C.float() function to print floating point numbers with
string.format("%g") instead of string.format("%a"), numbers like 0.0 and 1.0
started being printed as 0 and 1, without a decimal dot.

Most of the time, this did not make a difference because C automatically coerces
integer literals to double where needed. However, it produced the wrong result
when things like (1.0/2.0) in Pallene were converted to (1/2).

In addition to being more correct, adding a decimal dot to the floating point
literals in C also makes the generated C code a bit easier to read.

### DIFF
--- a/pallene/C.lua
+++ b/pallene/C.lua
@@ -37,7 +37,8 @@ function C.float(n)
     -- 17 decimal digits should be able to accurately represent any IEE-754
     -- double-precision floating point number except for NaN and infinities.
     -- The HUGE_VAL macro is a part of math.h. For more info, please see the
-    -- quotefloat function in lstrlib.c and https://stackoverflow.com/a/21162120
+    -- quotefloat function in lstrlib.c, tostringbuff in lobject.c, and
+    -- https://stackoverflow.com/a/21162120
     if n == math.huge then
         return "HUGE_VAL"
     elseif n == -math.huge then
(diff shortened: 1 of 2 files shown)


pallene/C.lua
```diff
@@ -37,15 +37,21 @@ function C.float(n)
...
     if n == math.huge then
         return "HUGE_VAL"
     elseif n == -math.huge then
         return "-HUGE_VAL"
     elseif n ~= n then
         error("NaN cannot be round-tripped")
     else
-        return string.format("%.17g", n)
+        local s = string.format("%.17g", n)
+        if s:match("^%-?[0-9]+$") then
```
gligneul: This seems overly-complicated. Couldn't we use `%e`?

hugomg: Adding the ".0" like this is the same idea that Lua uses in its `tostring` function.

The problem with %e is that, like %f, it always prints 17 digits after the decimal dot. For example, 0.0 becomes 0.00000000000000000e+00

gligneul: Well, if you are worried readability of the generated code this makes sense. Anyway, I would rewrite the code as the following:
```
if isinteger(n) then
   string.format("%d.0", n)
else
  string.format("%.17g", n)
end
```
`isinteger` could be something like `n == math.ceil(n)`

hugomg: I tested this version and unfortunately it does not always work. For large numbers such as 1e19, isinteger can be true but the number is too large to be converted to integer. In Lua, trying to print them with %d gives an error, while in C it leads to integer overflow.

PR #136, opened 2019-12-09
Fix bug with GC return values

hugomg:
Fix a bug where the value of `exp` in `return exp` would be garbage collected
even though it shouldn't. The problem was that we were failing to consider that
the internal "return" variables are live until the end of the function.

### DIFF
--- a/pallene/gc.lua
+++ b/pallene/gc.lua
@@ -59,6 +59,10 @@ function gc.compute_stack_slots(func)
             end
         end
     end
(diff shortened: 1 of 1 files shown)


hugomg:
With the tools we have right now, I can't think of a way to produce a reliable test case for this bug. The bug only happens when you write `return exp`, and there is no space to insert a call to `collectgarbage` between the `exp` and the `return`. I've opened issue #137 to discuss the problem.

PR #139, opened 2019-12-12
Implement tables in Pallene

gligneul:
Still missing:

- [x] Make sure that `==` is working
- [x] Coder tests
- [x] Documentation
- [x] Limit the field-name length to a short string because the coder assumes that.
- [x] Investigate if it is possible to merge arrays and tables in coder.


hugomg:
What do you think about using the `types.T.Record` type for both C and Lua records? I mean, add another field to the Record tagged-union that tells whether it is a C or Lua record, instead of having a separate Table and Record type.

One thing I found a bit strange is that inside the checker there is a piece of code that checks if `rec_type` is a Table or a Record, and then if it is either it proceeds assuming that the `field_types` field is present, and means the same thing in both cases. I think this doesn't look like how we normally do things.

runtime/pallene_core.h
```diff
@@ -226,4 +226,41 @@ int pallene_is_truthy(const TValue *v)
...
+        luaH_resize(L, t, 0, n);
+    }
+    return t;
+}
+
+static const TValue PALLENE_ABSENTKEY = {ABSTKEYCONSTANT};
+
+static inline
+TValue *pallene_getstr(Table *t, TString *key, int *pos)
+{
```
hugomg: Apparently there are two getstr's in Lua. One is a macro defined in lobject.h and one is a function from ltable.c. Maybe we should add a comment to clarify this is referring to luaH_getstr from ltable.c.

runtime/pallene_core.h
```diff
@@ -226,4 +226,41 @@ int pallene_is_truthy(const TValue *v)
...
+    if (PALLENE_LIKELY(*pos < sizenode(t))) {
```
hugomg: This test should also check that pos is non-negative. Use an unsigned comparison, like we do for arrays.

runtime/pallene_core.h
```diff
@@ -226,4 +226,41 @@ int pallene_is_truthy(const TValue *v)
...
+    Node *n = gnode(t, currpos);
+    for (;;) {
+        if (keyisshrstr(n) && eqshrstr(keystrval(n), key)) {
+            *pos = currpos;
+            return gval(n);
+        }
+        else {
+            int nx = gnext(n);
+            if (nx == 0)
+                return (TValue *)&PALLENE_ABSENTKEY;  /* not found */
```
hugomg: Would it make sense to update the inline cache in this not-found case? For example `*pos = -1`?

runtime/pallene_core.h
```diff
@@ -226,4 +226,41 @@ int pallene_is_truthy(const TValue *v)
...
+    if (PALLENE_LIKELY(*pos < sizenode(t))) {
+       Node *n = gnode(t, *pos);
+       if (PALLENE_LIKELY(keyisshrstr(n) && eqshrstr(keystrval(n), key)))
+           return gval(n);
+    }
+    int currpos = lmod(key->hash, sizenode(t));
+    Node *n = gnode(t, currpos);
+    for (;;) {
+        if (keyisshrstr(n) && eqshrstr(keystrval(n), key)) {
+            *pos = currpos;
```
hugomg: What do you think about `*pos = n - gnode(t, 0)`, to avoid needing to create a `currpos` variable?

pallene/coder.lua
```diff
@@ -1118,6 +1123,64 @@ gen_cmd["SetArr"] = function(self, cmd, _func)
...
+gen_cmd["GetTable"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dst)
+    local tab = self:c_value(cmd.src_tab)
+    local key = self:c_value(cmd.src_k)
+    local dst_typ = cmd.dst_typ
+    local line = C.integer(cmd.loc.line)
+
+    return (util.render([[
+        {
+            static int cache = 0;
```
hugomg: It probably doesn't matter much, since the initial value of the cache gets quickly overwritten, but should this perhaps be initialized to -1 instead of 0?

pallene/coder.lua
```diff
@@ -1118,6 +1123,64 @@ gen_cmd["SetArr"] = function(self, cmd, _func)
...
+    local tab = self:c_value(cmd.src_tab)
+    local key = self:c_value(cmd.src_k)
+    local v = self:c_value(cmd.src_v)
+    local src_typ = cmd.src_typ
+    local line = C.integer(cmd.loc.line)
+    return (util.render([[
+        {
+            static int cache = 0;
+            TValue *slot = pallene_getstr($tab, $key, &cache);
+            if (PALLENE_UNLIKELY(isabstkey(slot))) {
```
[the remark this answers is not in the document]
gligneul: Well, it could be an error. Probably making it an error will bump the performance a little bit because the slow-path won't return. Can we go in this direction?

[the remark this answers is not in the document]
gligneul: I implemented it this way to be consistent with the behaviors of arrays, that allow you to insert keys in positions that are nil. Anyway, triggering an error is fine by me.

runtime/pallene_core.h
gligneul: I don't see the point.

runtime/pallene_core.h
gligneul: Why would pos be negative?

pallene/coder.lua
gligneul: I don't see the point, this makes the code more complicated since -1 isn't a valid position. I think we should reword this variable as `size_t` to make the code clearer.

gligneul:
> What do you think about using the types.T.Record type for both C and Lua records? I mean, add another field to the types.T.Record tagged-union that tells whether it is a C or Lua record, instead of having a separate Table and Record type. 

I'm not thrilled for that because they are different types. Adding a field in T.Record will just create another indirection in the remaining parts of the code. For instance, the creation of a table in `to_ir.lua` is totally different from the creation of a record.

> One thing I found a bit strange is that inside the checker there is a piece of code that checks if rec_type is a Table or a Record, and then if it is either it proceeds assuming that the field_types field is present, and means the same thing in both cases. I think this doesn't look like how we normally do things.

The coder can assume that because both types have the `field_types` field with the same semantic. This is polymorphism in a dynamic language; you could interpret that both these types implement a common interface. 

Moreover, we should think of an abstraction that also covers modules because they will use the dot notation as well. However, it still not clear to me the design of this abstraction.

pallene/coder.lua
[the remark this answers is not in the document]
hugomg: If we decide to make it an error, we'd have to be careful about correctly dealing with fields that have type `value` or `nil`. In these cases, a hole in the table might not be an error.

runtime/pallene_core.h
hugomg: As I suggested in other comments, we could use a negative pos to signify that the cache is invalid. But even if we didn't do that, using an unsigned comparison would be an extra layer of defensive programming, at no cost.

gligneul:
> One thing I found a bit strange is that inside the checker there is a piece of code that checks if rec_type is a Table or a Record,

I'm writing a proposal to improve the odd code, I will push it soon.

pallene/coder.lua
hugomg: If we choose to go the size_t route, -1 (aka UINT_MAX) could still be used as a marker of invalid cache position.

runtime/pallene_core.h
hugomg: Consider a program where the key is not found once. Odds are, that it will continue being not found once we go through this line again with other tables. If the cache remains a valid index, every time we're going to start the table access with an optimistic access to the cache, which is going to fail. It would be better to just skip and go straight to the for loop.

hugomg:
Looking good so far!

I still wonder about some those aspects of the cache heuristic that we discussed before:

1. Is it better to start the cache with a default of zero? Or should it use an invalid default value like `UINT_MAX` or `-1`?
2. When the key is not found, should the cache be kept unchanged, or should we invalidate it?

I still have a strong suspicion that the answer for question (1) should be to start with an invalid index, as there is no expectation that the value will likely be found at index 0. As for question (2), I don't really know. But I would like to see a stronger justification for the choice we make, whatever it might be.

gligneul:
> Is it better to start the cache with a default of zero? Or should it use an invalid default value like UINT_MAX or -1?

I don't think it makes a difference. I would just leave it as zero to be simpler.

> When the key is not found, should the cache be kept unchanged, or should we invalidate it?

Why bother to invalidate it? At best invalidating the cache will make the performance worse. 

> I still have a strong suspicion that the answer for question (1) should be to start with an invalid index, as there is no expectation that the value will likely be found at index 0.

But I can't see the benefit of initializing it with a invalid index. What is the rationale?

hugomg:
I think the rationale for all those questions is that it would be better to jump straight to the "cache miss" case instead of waste time making a wrong guess and then still fall through to the "cache miss" case.

gligneul:
> That is, it would be better to have no cache than to have a cache pointing to the wrong place.

That is fair. I will make the change because the rationale makes sense but I don't think the performance benefit would be noticeable.

gligneul:
Ready to review

hugomg:
When we try to access a `value` field that is not initialized, it runs into a similar problem that we had with arrays. `get_slot` returns an "empty" value, which is a special variant of nil that only exists inside tables and arrays, never in local variables or the Lua stack.

```lua
--- Pallene code
type T = { x: value }

function getx(t: T): value
    return t.x
end
```

Test case:

```
$ ./lua/src/lua -lfoo
Lua 5.4.0  Copyright (C) 1994-2019 Lua.org, PUC-Rio
> foo.getx({})
nil
> foo.getx({}) == nil
false
```

Currently the `"GetArr"` case includes some special logic to convert these "empty" nils into regular nils. We should use this same logic when reading from tables.

I think the more robust way to take the fix_nils" logic from GetArr and move it to the `get_slot` function. One way to do this would be to refactor `get_slot` into two separate functions, `get_stack_slot` and `get_heap_slot`. The current `get_slot` function would become `get_stack_slot`, and `get_heap_slot` would be the version of the function that also converts "empty" nils into regular nils when `typ` is `types.T.Value`.

gligneul:
We could also trigger a runtime error to be consistent with the set instruction:
```
> t.set({}, 10)
stdin:1: attempt to access nonexistent field 'x' in line 16
```

gligneul:
When assigning nil explicitly, the code works:
```
> t.get({x = nil}) == nil
true
```
The major issue with this approach is that the behavior becomes implementation defined. This is how the code behaves today:
```
> p = {}
> p.x = 10
> t.get(p) == 10
true
> p.x = nil
> t.get(p) == nil
true
> p.y = 10 --> it doesn't increase the table hash size and x loses its slot
> t.get(p) == nil
false
> t.set(p, 10) --> triggers an error even though we initialized x before
stdin:1: attempt to access nonexistent field 'x' in line 16
stack traceback:
	[C]: in function 't.set'
	stdin:1: in main chunk
	[C]: in ?
```

I would prefer to change the setter logic instead to increase the table size if the value is not found.

pallene/types.lua
```diff
@@ -98,6 +138,24 @@ local function equivalent(t1, t2, is_gradual)
...
+                return false
+            end
+        end
+
+        for name, typ in pairs(f2) do
+            if not f1[name] or not equivalent(f1[name], typ, is_gradual) then
+                return false
+            end
+        end
+
```
hugomg: What about rewriting this using 3 loops? This way we would only call `equivalent` once instead of twice.
1. Check that all fields in t1 are present in t2
2. Check that all fields in t2 are present in t1
3. Check that all fields in t1 are equivalent to the corresponding fields in t2

runtime/pallene_core.h
```diff
@@ -226,4 +230,46 @@ int pallene_is_truthy(const TValue *v)
...
+
+/* This function is a speciallization of luaH_getshortstr from ltable.c */
+static inline
+TValue *pallene_getstr(Table *t, TString *key, size_t *pos)
+{
+    if (PALLENE_LIKELY(*pos < sizenode(t))) {
+       Node *n = gnode(t, *pos);
+       if (PALLENE_LIKELY(keyisshrstr(n) && eqshrstr(keystrval(n), key)))
+           return gval(n);
+    }
```
hugomg: For `gcc` the LIKELY macro is interpreted to assume that the condition is true at least 90% of the time. In this case, using LIKELY for this cache signifies that we expect that the cache will be a hit at least 81% of the time.

Is it safe to assume that? If not, it would be more accurate to use a regular if-statement without the PALLENE_LIKELY annotations.

spec/coder_test_operators.lua
```diff
@@ -24,7 +24,10 @@ local values = {
     },
     ["{integer}"] = {
         {}, {17}, {17}, {17, 18},
-    }
+    },
+    ["{x: integer, y: integer}"] = {
+        {x = 0, y = 0}, {x = 1, y = 0}, {x = 0, y = 1}, {x = 1, y = 1},
```
hugomg: I think at least one test case should appear twice in the list, like we did with arrays just above this.

runtime/pallene_core.h
gligneul: That really depends on the benchmark that you are doing. In fact, any simple benchmark like ours will benefit this LIKELY case because it will always hit the cache eventually. Bigger programs might have an issue. I will remove the LIKELY for now to be more fair.

hugomg:
I wrote a small benchmark program to test the array access:

```
-- foo.pln
function test(xs: {value}, n: integer): value
    local x: value = nil
    for i = 1, n do
        x = xs[1]
    end
    return x
end
```
I then tested it with `time ./lua/src/lua -lfoo.pln -e "foo2.test({}, 1000000000)"`

I manually edited the generated C source code and tested three versions:

### Version 1, the default one

```
x4 = (isempty(slot) ? PALLENE_NILKEY : *(slot));
```

Run time: 1.64s

### Version 2: if statement instead of ?:

```
if (isempty(slot)) {
    x4 = PALLENE_NILKEY;
} else {
    x4 = *(slot);
}
```
Run time: 1.30s

### Version 3: same method as master branch

```
x4 = *(slot);
if (isempty(&x4)) {
    setnilvalue(&x4);
}
```
Run time: 0.82s

----------------------

With the caveat that this is only a very artificial microbenchmark this is what I have seemed to identify so far:

1. I only created version 2 to make the line numbers in the generated ASM easier to read, but for whatever reason version 2 is faster than version 1. Dunno why.
2. Version 2 seems to be able to optimize the fact that PALLENE_NILKEY is known at compile time. It doesn't actually read from the global variable in the top case of the branch.
3. Version 3 compiles down to a branchless version, using the `cmove` instruction.

runtime/pallene_core.h
```diff
@@ -226,4 +226,47 @@ int pallene_is_truthy(const TValue *v)
...
+    return t;
+}
+
+static const TValue PALLENE_ABSENTKEY = {ABSTKEYCONSTANT};
+static const TValue PALLENE_NILKEY = {{NULL}, LUA_TNIL};
+
+/* This function is a speciallization of luaH_getshortstr from ltable.c */
+static inline
+TValue *pallene_getstr(Table *t, TString *key, size_t *pos)
```
hugomg: What do you think about renaming this to pallene_getshortstr?

Issue #135, opened 2019-12-08
Standardize single quotation marks in checker error messages


gligneul:
Some messages use quotation marks to enclose names (functions, variables, or types); for instance:
```
    it("forbids variables from being used before they are defined", function()
        assert_error([[
            function fn(): nil
                x = 17
                local x = 18
            end
        ]],
            "variable 'x' is not declared")
    end)
```
However, in other cases there is no quotation marks, for instance:
```
    it("catches assignment to function", function ()
        assert_error([[
            function f()
            end

            function g()
                f = g
            end
        ]],
            "attempting to assign to toplevel constant function f")
    end)
```
We should decide what is the best and stick with it.

hugomg:
I vote on using single quotes, like `'f'`

1. Lua uses single quotes. the error message for calling a non-existing function is `attempt to call a nil value (global 'f')`
2. Clang seems to use single quotes too: `warning: implicit declaration of function 'f' is invalid in C99`
3. GCC uses fancy single quotes: `warning: implicit declaration of function ‘f’`

ahmedkrmn:
Will start working on this and submit a PR once finished.

hugomg:
Fixed by #142

PR #142, opened 2019-12-15
Standardize single quotation marks in checker error messages

ahmedkrmn:
Fixes issue #135.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -180,7 +180,7 @@ function Checker:check_program(prog_ast)
             local old_loc = names[name]
             if old_loc then
                 scope_error(loc,
-                    "duplicate toplevel declaration for %s, previous one at line %d",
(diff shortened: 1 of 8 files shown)


ahmedkrmn:
@hugomg and @gligneul I think Travis is failing because this commit changes the error messages, right ?!

hugomg:
Thanks for your PR!

I think Travis is working fine in this case. The problem is that some of the unit tests inside the `spec` directory stopped passing after you changed the error messages. Those unit tests will also need to be updated to match the changes made.

To run the unit tests on your machine, install and run the busted tool, as described in the [Running the test suite](https://github.com/pallene-lang/pallene#running-the-test-suite) section of the README.

ahmedkrmn:
@hugomg Done 👍

ahmedkrmn:
@hugomg Done 👍

PR #145, opened 2019-12-18
Add unit test for conflicting arguments when calling pallenec

### LINKED ISSUE #87: We should have a spec file that tests pallenec flags
Better pallenec test coverage would prevent bugs like the one fixed in #86

ahmedkrmn:
Fixes #87 
As suggested in #86, this PR adds an additional unit test to [`/spec/pallenec_spec.lua`](https://github.com/pallene-lang/pallene/blob/master/spec/pallenec_spec.lua) that checks against calling the compiler with conflicting arguments.
This utilizes the `pallenec_abort` function in the pallenec compiler.


ahmedkrmn:
Actually the `--emit-asm` argument is saved as `--emit-s` in the `flags` table in [`pallenec`](https://github.com/pallene-lang/pallene/blob/master/pallenec).
I can open another PR that fixes this typo and then rename the error message in this commit.

PR #157, opened 2019-12-24
Implement mutually-recursive functions

### LINKED ISSUE #84: Mutually-recursive functions
Currently, Pallene does not allow functions to call other functions that are defined further down in the file. I suggest that we allow consecutive function declarations to be mutually-recursive. That is, function declarations are mutually-recursive if there isn't an intervening variable or type declaration between them.

```
local function even(n : integer): boolean
    if n == 0 then
        return true
    else
        return odd(n-1)
    end
end

local function odd(n : integer): boolean
    if n == 0 then
        return false
    else
        return even(n-1)
    end
end
```

Current Lua idioms for mutually-recursive functions are unsatisfactory for Pallene. The first one is to forward-declare local variables:

```
local odd, even
odd = function(n) ... end
even = function(n) ... end
```

The first problem is that Pallene does not allow exporting local variables, so we would need to have  way to tell apart local variables being used in function forward-declarations, and other potentially-mutable local variables. Furthermore this requires adding a bunch of features to the Pallene compiler: uninitialized variables, an optimization pass to determine that odd and even are constants, and are always initialized by the time they are called, and so  on.

The other approach used in Lua is to indirect the function calls through a table:

```
local M = {}
function M.even(n) ... end
function M.odd(n) ... end
```

This is less than ideal in Pallene because it only works for "public" functions, and it requires storing the functions in a table.

hugomg:
Fixes #84. It turns out that one of the experiments I will want to run in the next paper needs mutually
recursive functions. So here they are.

This PR also paves the way for mutually-recursive types but doesn't implement
them yet.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -265,7 +265,8 @@ Parameters are a comma-separated list of `<name>: <type>`.
 Two parameters cannot have the same name.
 The function body is a sequence of statements.
 
-Pallene functions can refer to functions defined before them but not to functions defined further down in the file. A future version of Pallene will lift this restriction and allow for mutually-recursive functions.
+Pallene functions can be recursive.
+Blocks of mutually-recursive functions are also allowed, as long as the mutually-recursive functions are declared next to each other, without any type or variable declarations between them.
 
 ## Expressions and Statements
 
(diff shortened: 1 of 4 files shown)


gligneul:
The code is fine but we should add this to the manual.

hugomg:
Good catch. Updated the manual now.

PR #159, opened 2019-12-27
Implement '%' operator for floating point numbers

### LINKED ISSUE #149: Implement % operator for floating point numbers
Currently, Pallene implements the `%` operator for integers, but not for floating point numbers. The following test case currently crashes the compiler with a "not implemented" exception:

```
function f(a: float, b: float): float
    return a % b
end
```

To implement this we would need to add a case inside `gen_cmd["Binop"] ` in coder.lua to generate the appropriate C code for the FltMod operation. The simplest implementation would be to call the luaV_modf function from lvm.c

It might also be worthwhile to test the performance of re-exporting the luaV_modf function as an `static_inline` function in pallene_core.h

We should also remember to add the appropriate test case in `coder_spec.lua`. We should also add more non-integer floating point numbers to `coder_test_operators.lua` to ensure that the tests for the `%` operator cover all the cases.

ahmedkrmn:
Closes #149 

**As requested in the issue, this PR does the following:**
- Calls the `luaV_modf` function from `lvm.c` to calculate the modulus of float numbers in  `coder.lua`.
- Re-exports `luaV_modf` function as a `static inline pallene_float_modi()` in `pallene_core.h`.
- Adds 4 test cases of the how the function might be executed `coder_spec.lua`.
- Adds more floating point numbers to `coder_test_operators.lua`.


ahmedkrmn:
@hugomg Sorry for taking this long to implement it. I was busy with delivering some college projects last week!

I've followed your suggestions in the issue but I still have a question:
 - Why do we have to re-implement each function in `lvm.c` as a `static inline` function in `pallene_core.h`?

hugomg:
> Sorry for taking this long to implement it. I was busy with delivering some college projects last week!

Hahaha, no need to rush it. Help is always welcome at any time, and today is literally the day after Christmas :)

> Why do we have to re-implement each function in lvm.c as a static inline function in pallene_core.h?

That is an excellent question. The simple answer is that we create an inline function when it helps with optimization. The comments in pallene_core.h give some examples of this.

For example, consider the `<<` operator in Pallene. The Lua implementation of this operator is the `luaV_shiftl` function in lvm.c. We could just call that if we wanted to keep it simple. But in this case we instead call `pallene_shiftL`, which is a more efficient implementation. The main difference is that when the `y` is known at compilation time, the C compiler can inline the implementation of `pallene_shiftL` and throw away the dead parts of the if-statement. In those cases, GCC can optimize a call like `pallene_shiftL(x, 3)` down to a single left shift instruction. You can see this if you look at the generated assembly code with `objdump` or with `pallenec --emit-asm`.

In the case of `%`, the simplest implementation would be to just call luaV_modf, like you are already doing. We would only need to create a specialized implementation in pallene_core.h if it would make a big speed difference. Would you be able to run that experiment to measure if it helps or not? (I can help you do that if you want).

pallene/coder.lua
```diff
@@ -968,7 +968,7 @@ gen_cmd["Binop"] = function(self, cmd, _func)
     elseif op == "FltSub"    then return binop("-")
     elseif op == "FltMul"    then return binop("*")
     elseif op == "FltDivi"   then return flt_divi()
-    elseif op == "FltMod"    then return flt_mod()
+    elseif op == "FltMod"    then return flt_mod("pallene_float_modi")
```
hugomg: The flt_mod function doesn't receive any parameters, so there is no need to pass a parameter here.

runtime/pallene_core.h
```diff
@@ -155,6 +155,14 @@ lua_Integer pallene_int_modi(
...
 }
 
+static inline
+lua_Number pallene_float_modi(lua_State *L ,lua_Number m, lua_Number n)
+{
+    lua_Number r;
+    luai_nummod(L, m, n, r);
+    return r;
+}
+
```
hugomg: Currently, this `pallene_float_modi` function is being unused. We should either be use it instead of luaV_modf, or we should just delete it.

pallene/coder.lua
ahmedkrmn: Ah, yeah that's right. Things are starting to make more sense now!

runtime/pallene_core.h
ahmedkrmn: > We would only need to create a specialized implementation in pallene_core.h if it would make a big speed difference. Would you be able to run that experiment to measure if it helps or not? (I can help you do that if you want).

I'm actually very interested in implementing it from scratch and comparing the difference in performance. How do you suggest we proceed?

hugomg: What I like to do is to test the difference "by hand" before implementing it in the coder.lua

What I usually do is that I take a small Pallene file and compile it to C with `./pallenec --emit-c`. Then, I manually edit the part of the C file that I am interested in and compile the result with `./pallenec --compile-c` to test the result.

If you want to test multiple versions side-by-side you can copy the original C file. For example, if you have a `foo.pln` file and generate a `foo.c` file you can copy that `foo.c` to `foo1.c`, `foo2.c` and then compare the performance of `foo1.so` and `foo2.so`. The only thing to remember is that if you rename one of these generated C files you also need to rename the "luaopen" function that appears near the end of the file.

Let me know if you run into any problems. This is one of those areas that I still need to write more documentation about.

ahmedkrmn: I've been working on this for a bit now and I've two question:

1. What is a `lua_State`? Why do we have to pass it to `luai_nummod()`? And if I implement a new float modulus function in `pallene_core.h`, do I need to pass the `lua_State`?
2. I've implemented a function that computes the modulus in O(log N), how can I compare its perfomance programatically against `luai_nummod()`?

hugomg: > What is a lua_State

The Lua state stores all the information that the Lua interpreter needs to operate. The Lua interpreter does not use any global variables, it is all in the Lua state.

Usually, the place where programmers will encounter a lua_State is in the Lua-C API. Lua is designed to be an extensible and embeddable programming language, meaning that it is possible to write C functions that are called from Lua, and to call Lua code from C functions. The usual way to use this is through the [C-API](https://www.lua.org/work/doc/manual.html#4). Most of the functions in this API take a lua_State as a parameter. I think the best explanation of all of this can be found in the Programming in Lua book.

When using the normal Lua API, the lua_State is a black box. You pass a pointer to it around but you never look at what is inside. In Pallene we break this rule. There are some situations where we use private Lua APIs and peek directly inside the Lua state.

> Why do we have to pass it to luai_nummod()

That is a good question. As it is currently defined, luai_nummod never uses the lua_State so it could certainly have been defined without taking it as a parameter.

I'd have to ask Roberto why he did this but if I had to guess my guess would be that that during debugging he might have wanted to replace the definition of luai_nummod with an alternate definition that needed the lua_State parameter.

> And if I implement a new float modulus function in pallene_core.h, do I need to pass the lua_State?

The simplest implementation would be to call luai_nummod, which needs the lua_State (even though it doesn't actually use it).

> a function that computes the modulus in O(log N)

Is this function computing a floating-point modulus or an integer modulus?

One thing to be careful about here is that we want to be sure that the modulus operator in Pallene returns the exact same results as Lua. I think it would be difficult to do that with a completely different implementation.

> how can I compare its perfomance

The way I would normally do this is compile a Pallene file down to C with `./pallenec --compile-c foo.pln`, then manually edit the generated C file to put in the new implementation, then compile that C file with `./pallenec --compile-c` and then measure the performance with the `time` command.

For example, `time ./lua/src/lua -l foo -e "print(foo.f(3.14))"`

ahmedkrmn: Thanks for that thorough response @hugomg 🥇 
The function I implemented solves both float and integer modulus and it has nearly the same running time as Lua's `luai_nummod()`. So, I believe we can just go with using Lua's  for the time being. 
I will now delete the `pallene_float_modi()` function since it's unused.

hugomg:
Going with the simpler version was a wise choice.

PR #160, opened 2019-12-27
Create temp files for intermediate compilation steps

### LINKED ISSUE #71: Create temporary files with temporary names
Currently, if I have a `.c` file with the same name of the `.pallene` file, the compiler overwrites the `.c` file with a temporary C code.
Then, at the end of the compilation the file is wiped.
Instead of overwriting a possible existent file, the compiler should use something like mktmp to create a unique file.

ahmedkrmn:
Closes #71.

- Using Lua's `os.tmpname()`, a new temp file is created for each intermediate compilation step.
- The `-x` is added to the C compiler to force it to compile in the specified language since temp files won't have any extensions.

### DIFF
--- a/pallene/c_compiler.lua
+++ b/pallene/c_compiler.lua
@@ -56,6 +56,8 @@ function c_compiler.compile_s_to_o(in_filename, out_filename)
 end
 
 function c_compiler.compile_o_to_so(in_filename, out_filename)
+    -- There is no need to add the '-x' flag when compiling an object file without a '.o' extension.
+    -- According to GCC, any file name with no recognized suffix is treated as an object file.
     return run_cc({
         c_compiler.CFLAGS_SHARED,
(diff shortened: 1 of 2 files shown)


ahmedkrmn:
> Lovely!
> 
> I was surprised to learn that we did not need to pass a "-x" flag when linking the object files, even if they do not have a ".o" extension.

Yeah, I've also just discovered that when going through the man!
> Could you add a small comment to `compile_o_to_so` to clarify that there is a good reason why `compile_o_to_so` is different from `compile_c_to_s` and `compile_s_to_so`?

I'm not quite getting what you mean here. Do you mean specifying why we didn't use `-x` in `compile_o_to_so` or do you mean something else?

hugomg:
Exactly, we could add a comment saying why we don't need to use the -x flag.

PR #163, opened 2020-01-13
Update README

leokaplan:
Improved LuaRocks installation instructions for Bash
Added /usr/bin/time instalation instructions for Debian
Fix typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,9 +140,10 @@ the benchmark, without measuring anything.
 ```sh
 ./benchmarks/run benchmarks/sieve/pallene.pln --mode=none
 ```
-It may be the case that your terminal will come with a time program, but will lack the `/usr/bin/time`, which will raise an assertion error in lua when you try to run a benchmark. In that case install with `$ sudo apt install time` or equivalent.
 
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,14 +135,18 @@ script from the root project directory:
 By default, the benchmark runner just outputs the running time, as measured by
 `/usr/bin/time`, but it also supports other measurements. For example,
 `--mode=perf` shows perf output and `--mode=none` shows the stdout produced by
-the benchmark, withot measuring anything.
+the benchmark, without measuring anything.
 
 ```sh
 ./benchmarks/run benchmarks/sieve/pallene.pln --mode=none
 ```
+It may be the case that your terminal will come with a time program, but will lack the `/usr/bin/time`, which will raise an assertion error in lua when you try to run a benchmark. In that case install with `$ sudo apt install time` or equivalent.
```
hugomg: Bash's `time` is a builtin shell function, not a program. Suggestion:

To run Pallene's benchmarks you need to have `/usr/bin/time` installed in your system. Some Linux distributions may have only the Bash `time` builtin function but not the `/usr/bin/time` executable. If that is the case you will need to install `time` with `$ sudo apt install time` or equivalent.

PR #172, opened 2020-02-05
Implement Lua 5.4 for loop overflow behavior

### LINKED ISSUE #170: Update for loop semantics to Lua 5.4
Pallene's implementation of the for loop is not equivalent to Lua's in some corner cases:

The first is when the loop step is zero:
* Lua 5.3 loops infinitely if start ≥ limit, and skips the loop if start < limit
* Pallene loops infinitely if start > limit, and skips the loop if start ≤ limit.
* Lua 5.4 raises a runtime error.

The second is when the loop variable overflows:
* In Lua 5.3 and Pallene, the internal loop variable can overflow, causing an infinite loop.
* In Lua 5.4, a loop count is computed before the start of the loop, using integer division.

hugomg:
Fixes #170.

I disabled the -Wshadow flag to let me always use the same variable names in the for-loop macro. Do you remember if that warning was enabled for some important reason? If you can't remember, do you know if there is an easy way to check the git history to find that out?

### DIFF
--- a/runtime/pallene_core.h
+++ b/runtime/pallene_core.h
@@ -317,7 +317,7 @@ TValue *pallene_getshortstr(Table *t, TString *key, size_t * restrict pos)
         lua_Number _init  = A; \
         lua_Number _limit = B; \
         lua_Number _step  = C; \
-        if (_step == 0.0 ) { \
+        if (_step == 0.0) { \
             luaL_error(L, "'for' step is zero"); \
         } \
         for ( \
--- a/runtime/pallene_core.h
+++ b/runtime/pallene_core.h
@@ -270,4 +270,66 @@ TValue *pallene_getshortstr(Table *t, TString *key, size_t * restrict pos)
     }
 }
 
+/* To avoid looping infinitely due to integer overflow, Lua 5.4 carefully
+ * computes the number of iterations before starting the loop (see OP_FORPREP).
+ *
+ * The code that implements this behavior does not look like a regular C for
+ * loop, so to help improve the readability of the generated C code we hide it
+ * behind the following set of macros. Note that some of the open braces in the
+ * BEGIN macro are only closed in the END macro, so they must always be used
+ * together.
+ *
+ * We assume that the C compiler will be able to optimize the common case where
+ * the step parameter is a compile-time constant. */
+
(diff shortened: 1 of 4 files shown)


gligneul:
> Do you remember if that warning was enabled for some important reason? If you can't remember, do you know if there is an easy way to check the git history to find that out?

It was a coding styling decision. I prefer to avoid shadowing, but this is not that important.

runtime/pallene_core.h
```diff
@@ -266,4 +266,66 @@ TValue *pallene_getshortstr(Table *t, TString *key, size_t * restrict pos)
     }
 }
 
+/* To avoid looping infinitely due to integer overflow, Lua 5.4 carefully
```
gligneul: I think we should start to split this file into multiple files. We could have a single header called `pallene.h` that include all other headers.

runtime/pallene_core.h
```diff
@@ -266,4 +266,66 @@ TValue *pallene_getshortstr(Table *t, TString *key, size_t * restrict pos)
...
+ *
+ * We assume that the C compiler will be able to optimize the common case where
+ * the step parameter is a compile-time constant. */
+
+#define PALLENE_INT_FOR_LOOP_BEGIN(i, A, B, C) \
+    { \
+        lua_Integer _init  = A; \
+        lua_Integer _limit = B; \
+        lua_Integer _step  = C; \
+        if (_step == 0 ) { \
```
gligneul:
```suggestion
        if (_step == 0) { \
```

runtime/pallene_core.h
hugomg: Right now we only have slightly over 10 functions and macros in pallene_core.h, so I don't feel an urgent need to split it up. I think the coder.lua is a bigger offender in this sense.

But I wouldn't be against it if someone comes up with a reasonable way to split the pallene_core.h into smaller files.

gligneul: There are a lot of lengthy files in Pallene source code; this header file isn't that long, but surely it could be broken down. To newcomers (and even myself) it is hard to navigate through these files. And most of the time these files could be easily divided because the functions are not interdependent.

By the way, you can merge this PR.

hugomg: Certainly. I don't know what would be the best way to split this particular file but I am open to suggestions.

Issue #154, opened 2019-12-22
Implement table field names longer than 40 characters


hugomg:
The current implementation of table fields only works with table fields of up to `LUAI_MAXSHORTLEN` bytes (40).

Very long field names like this are exceedingly rare so this isn't a high-priority feature, and it isn't important that the implementation be very efficient either.

gligneul:
It is important that lengthy fields doesn't affect the performance of normal-sized fields; the differentiation should be solved during compile time.
Anyway, I don't think it is worth to implement lengthy fields due to the extra complexity in the compiler.

hugomg:
I'm not sure the complexity would be very high, because we woudn't be optimizing this implementation like we did with the short strings. But I agree with you that this is not a big priority.

To be honest, the thing that actually bothers me more is that currently the limit of 40 that Pallene uses does not necessarily match with the `LUA_MAXSHORTLEN` that the C code sees, as I discuss in #153.

gligneul:
> To be honest, the thing that actually bothers me more is that currently the limit of 40 that Pallene uses does not necessarily match with the LUA_MAXSHORTLEN that the C code sees, as I discuss in #153.

We would still have this problem because we would need to know this constant during the compilation of a Pallene module. If we have the path of the Lua source code inside the compiler, we could obtain this value appropriately.

hugomg:
> We would still have this problem because we would need to know this constant during the compilation of a Pallene module.

Not necessarily. We generate the code using ifdefs and let the C preprocessor take care of it.

ahmedkrmn:
@hugomg How do you think I approach this?

You said there's a problem with the Pallene implementation because the limit it uses does not necessarly match with `LUAI_MAXSHORTLEN`. 
Is this what the issue is aiming to fix? Or are you looking to remove that limit and put no bound on the table field size?

hugomg:
The first problem we are trying to fix is that Pallene currently outright rejects record fields with long names. If you look in checker.lua, line 165, you'll see that it is producing an error message in that case. 

The initial fix would be to move this if statement to the coder.lua. If the length of the field name is ≤ 40, then we use the current implementation, which is specialized for "short strings". It the length of the field name is > 40, then we need to create a new implementation, which should call luaH_getstr instead of pallene_getshortstr.

[...]

PR #176, opened 2020-02-16
Allow table field names longer than 40 character

ahmedkrmn:
- This PR implements issue #154.

- The code compiles and runs successfully for both long and short table field names, but with the following warning in case of long fields:

    warning
I tried some fixes for this warning but they involved modifying the source code of the used Lua version.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1202,7 +1202,6 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
     local tab = self:c_value(cmd.src_tab)
     local key = self:c_value(cmd.src_k)
     local dst_typ = cmd.dst_typ
-    local line = C.integer(cmd.loc.line)
 
     assert(cmd.src_k._tag == "ir.Value.String")
     local field_name = cmd.src_k.value
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1222,7 +1222,7 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
     -- Else, use Lua's default getStr method
     else
         find_slot = util.render([[
-            TValue *slot = luaH_getstr($tab, $key);
+            TValue *slot = cast(TValue *, luaH_getstr($tab, $key));
         ]], ops)
     end
 
@@ -1260,7 +1260,7 @@ gen_cmd["SetTable"] = function(self, cmd, _func)
     -- Else, use Lua's default getStr method
     else
         find_slot = util.render([[
-            TValue *slot = luaH_getstr($tab, $key);
+            TValue *slot = cast(TValue *, luaH_getstr($tab, $key));
         ]], ops)
     end
 
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1210,26 +1210,30 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
         get_slot = self:get_luatable_slot(dst_typ, dst, "slot", tab,
             cmd.loc, "table field"),
     }
+    local renderStr = [[
+        $get_slot
+    ]]
+
     -- If the table field size is smaller than 40, use the optimized getStr Pallene implementation
     assert(cmd.src_k._tag == "ir.Value.String")
     local field_name = cmd.src_k.value
     if #field_name < 40 then
-        return (util.render([[
+        return (util.render(string.format([[
             {
                 static size_t cache = UINT_MAX;
                 TValue *slot = pallene_getshortstr($tab, $key, &cache);
-                $get_slot
+                %s
             }
-        ]], ops))
+        ]], renderStr), ops))
 
     -- Else, use Lua's default getStr method
     else
-        return (util.render([[
+        return (util.render(string.format([[
             {
                 TValue *slot = luaH_getstr($tab, $key);
-                $get_slot
+                %s
             }
-        ]], ops))
+        ]], renderStr), ops))
     end
 end
 
@@ -1243,36 +1247,35 @@ gen_cmd["SetTable"] = function(self, cmd, _func)
         key = key,
         set_heap_slot = set_heap_slot(src_typ, "slot", v, tab),
     }
+    local renderStr = [[
+        if (PALLENE_UNLIKELY(isabstkey(slot))) {
+            TValue keyv;
+            setsvalue(L, &keyv, $key);
+            slot = luaH_newkey(L, $tab, &keyv);
+        }
+        ${set_heap_slot}
+    ]]
+    
     -- If the table field size is smaller than 40, use the optimized getStr Pallene implementation
     assert(cmd.src_k._tag == "ir.Value.String")
     local field_name = cmd.src_k.value
     if #field_name < 40 then
-        return (util.render([[
+        return (util.render(string.format([[
             {
                 static size_t cache = UINT_MAX;
                 TValue *slot = pallene_getshortstr($tab, $key, &cache);
-                if (PALLENE_UNLIKELY(isabstkey(slot))) {
-                    TValue keyv;
-                    setsvalue(L, &keyv, $key);
-                    slot = luaH_newkey(L, $tab, &keyv);
-                }
-                ${set_heap_slot}
+                %s
             }
-        ]], ops))
+        ]], renderStr), ops))
 
     -- Else, use Lua's default getStr method
     else
-        return (util.render([[
+        return (util.render(string.format([[
             {
                 TValue *slot = luaH_getstr($tab, $key);
-                if (PALLENE_UNLIKELY(isabstkey(slot))) {
-                    TValue keyv;
-                    setsvalue(L, &keyv, $key);
-                    slot = luaH_newkey(L, $tab, &keyv);
-                }
-                ${set_heap_slot}
+                %s
             }
-        ]], ops))
+        ]], renderStr), ops))
     end
 end
 
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1209,7 +1209,7 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
     return util.render([[
         {
             static size_t cache = UINT_MAX;
-            TValue *slot = pallene_get_fieldslot($field_len, $tab, $key, &cache);
+            TValue *slot = pallene_getstr($field_len, $tab, $key, &cache);
             ${get_slot}
         }
     ]], {
@@ -1232,7 +1232,7 @@ gen_cmd["SetTable"] = function(self, cmd, _func)
     return util.render([[
         {
             static size_t cache = UINT_MAX;
-            TValue *slot = pallene_get_fieldslot($field_len, $tab, $key, &cache);
+            TValue *slot = pallene_getstr($field_len, $tab, $key, &cache);
             if (PALLENE_UNLIKELY(isabstkey(slot))) {
                 TValue keyv;
                 setsvalue(L, &keyv, $key);
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1203,20 +1203,32 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
     local key = self:c_value(cmd.src_k)
     local dst_typ = cmd.dst_typ
     local line = C.integer(cmd.loc.line)
-
-    return (util.render([[
-        {
-            static size_t cache = UINT_MAX;
-            TValue *slot = pallene_getshortstr($tab, $key, &cache);
-            $get_slot
-        }
-    ]], {
+    local ops = {
         tab = tab,
         key = key,
         line = line,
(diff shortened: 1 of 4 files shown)


hugomg:
I'll take a closer look at this soon. Do you know if your C compiler is gcc or clang?

ahmedkrmn:
> I'll take a closer look at this soon. In the meantime, do you know if your C compiler is gcc or clang?

It's `gcc-8.3.0`.

pallene/coder.lua
```diff
@@ -1203,43 +1203,73 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
...
+                    TValue keyv;
+                    setsvalue(L, &keyv, $key);
+                    slot = luaH_newkey(L, $tab, &keyv);
+                }
+                ${set_heap_slot}
+            }
+        ]], ops))
+
+    -- Else, use Lua's default getStr method
+    else
```
hugomg: The body of the `else` should be indented.

spec/checker_spec.lua
```diff
@@ -130,14 +130,6 @@ describe("Pallene type checker", function()
...
     end)
 
-    it("prohibits tables with fields with more than LUAI_MAXSHORTLEN chars", function()
-        local field = string.rep('a', 41)
-        assert_error([[
-            function f(t: {]].. field ..[[: float}) end
-        ]],
-            "field name '".. field .. "' too long")
-    end)
-
```
hugomg: This test case should be moved to coder_spec. Except that instead of verifying that it produced an error, it should verify that it is working correctly.

hugomg:
This is looking promising. Once this first round of code review is done, we can discuss how to tackle the problem of the hardcoded `40`s

pallene/coder.lua
```diff
@@ -1203,43 +1203,73 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
...
         get_slot = self:get_luatable_slot(dst_typ, dst, "slot", tab,
             cmd.loc, "table field"),
-    }))
+    }
+    -- If the table field size is smaller than 40, use the optimized getStr Pallene implementation
+    if(#cmd.src_k.value < 40 ) then
+        return (util.render([[
+            {
+                static size_t cache = UINT_MAX;
+                TValue *slot = pallene_getshortstr($tab, $key, &cache);
```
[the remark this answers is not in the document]
ahmedkrmn: I tried this before. It shows a different warning:
warning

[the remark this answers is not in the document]
ahmedkrmn: I really can't find the macro in the source code. Can you please point out its location?

pallene/coder.lua
```diff
@@ -1203,43 +1203,77 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
...
+                if (PALLENE_UNLIKELY(isabstkey(slot))) {
+                    TValue keyv;
+                    setsvalue(L, &keyv, $key);
+                    slot = luaH_newkey(L, $tab, &keyv);
+                }
+                ${set_heap_slot}
```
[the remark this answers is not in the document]
ahmedkrmn: Can you check the new commit.
It declares a string `renderStr` with the repeated code in both the `util.render()` and use it in the function with the help of the `string.format()` method.

pallene/coder.lua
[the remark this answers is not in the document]
hugomg: Oops, I meant to say `luaV_finishfastset` in line 96 of file lvm.h

hugomg:
I think we can do simpler than that. What about just calling util.render twice, and then combining the results?

```
assert(cmd.src_k._tag == "ir.Value.String")
local field_name = cmd.src_k.value

local find_slot
if #field_name < 40 then
    find_slot = util.render(string.format([[
        static size_t cache = UINT_MAX;
        TValue *slot = pallene_getshortstr($tab, $key, &cache);
   ]], ops))
else
    find_slot = util.render(string.format([[
        TValue *slot = luaH_getstr($tab, $key);
    ]], ops))
end

return (util.render([[
    {
        ${find_slot}
        ${get_slot}
    }
]], {
    find_slot = find_slot,
    get_slot = self:get_luatable_slot(dst_typ, dst, "slot", tab, cmd.loc, "table field"),
})
```

For an example of this, see the check_tag variable in the get_stack_slot function.

ahmedkrmn:
> I think we can do simpler than that. What about just calling util.render twice, and then combining the results?
Very nice idea. I believe we don't need to use the `string.format()` function here.

pallene/coder.lua
```diff
@@ -1203,80 +1203,83 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
...
+            ${find_slot}
+            if (PALLENE_UNLIKELY(isabstkey(slot))) {
+                TValue keyv;
+                setsvalue(L, &keyv, $key);
+                slot = luaH_newkey(L, $tab, &keyv);
+            }
+            ${set_heap_slot}
+        }
+    ]], {
+        tab = tab,
```
ahmedkrmn: @hugomg Is there a way to unpack the contents of `ops` here instead of rewriting the `tab` and `key` keys and values again?

pallene/coder.lua
```diff
@@ -1203,80 +1203,83 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
     local key = self:c_value(cmd.src_k)
     local dst_typ = cmd.dst_typ
     local line = C.integer(cmd.loc.line)
```
ahmedkrmn: @hugomg Why do we also need to declare the `line` variable? It's used nowhere in the function.

hugomg: IIRC, GetTable needs the line info but SetTable doesn't. When we are reading a table field we also check if the type tag of the elemens is what we expected, and produce an error when it isn't.

pallene/coder.lua
hugomg: I think it is OK to repeat the `tab` and `key` fields in the two times that we call util.render. The alternatives would be worth the extra complexity.

hugomg:
I like where this is going. Next, we can try to get rid of the hardcoded `40` in our implementation.

One idea that might work would be to use an ifdef in the generated C code instead of an if statement in the compiler.

```c
#if field_len < LUAI_MAXSHORTLEN
    /* pallene_getshortstr implementation */
#else
    /* luaH_getstr implementation */
#endif
```

If this works, we could try creating a macro for it in `pallene_core.h`, to help keep the generated code more readable.

```c
#define pallene_getfieldslot(field_len, tab, key, slot) \
    #if field_len < LUAI_MAXSHORTLEN \
        /* pallene_getshortstr implementation */ \
    #else \
        /* luaH_getstr implementation */ \
    #endif \
```

pallene/coder.lua
ahmedkrmn: So you mean that the util.render() function always checks for line in its parameters in case of reading a table field?

hugomg: Oh, you are right. Now that the tag checking logic was moved to `get_luatable_slot`, the `util.render`  in `GetTable` is not using the `line` anymore. You can safely delete it.

ahmedkrmn: Got it. I Will remove it in a separate commit then start working on the other issue after this weekend.

runtime/pallene_core.h
```diff
@@ -332,4 +332,15 @@ TValue *pallene_getshortstr(Table *t, TString *key, size_t * restrict pos)
         } \
     }
 
+/* If the table field size is smaller than LUAI_MAXSHORTLEN, use the optimized getStr Pallene implementation.
+ * Else, use Lua's default getStr method. */
+#define pallene_getfieldslot(field_len, tab, key, slot) \
+    if (field_len < LUAI_MAXSHORTLEN) { \
```
hugomg: Do you know if it would be possible to use a preprocessor `#if` instead of a regular `if`? That way the cache variable would only be created if it were needed.

ahmedkrmn: My first attempt was to implement it like this as you proposed, but according to [this](https://stackoverflow.com/a/10074611/8831148), one can't embed a preprocessor directive in another preprocessor directive.

It showed that error message when I tried doing it:
`error: '#' is not followed by a macro parameter`

ahmedkrmn: Doing it as proposed in the link would cause a compilation error since the variable `field_len` isn't defined at this point.
```c
#if field_len < LUAI_MAXSHORTLE
#define /* pallene_getshortstr implementation */
#else
#define /* luaH_getstr implementation */
#endif
```

hugomg:
Hmm, I didn't know you couldn't put an `#if` inside another macro. I guess that I had never tried to do that...

The good news is that ran some experiments with GCC and Clang, and I think we can solve this problem without needing to use macros! Apparently, if we implement `pallene_getfieldslot` as an inline function, which receives a pointer to the cache, then both GCC and Clang optimize away the static cache variable depending on the field_len, which is know at compile time.

That is, we can change the generated code so it looks like this:

```
static size_t cache = UINT_MAX;
TValue *slot = pallene_getfieldslot($field_len, $tab, $key, &cache);
```
And we can change `pallene_getfieldslot` so that it is a `static inline` function instead of a macro.

```
static inline
TValue *pallene_get_field_slot(
        size_t field_len, Table *tab, TString *key, size_t *restrict cache)
{
    ...
}
```

At least in the case of GCC and Clang, the generated assembly is just as good as the one that we can get with a macro. The `cache` variable is optimized away if it is not actually needed.

hugomg:
Other than that, the only other comment I'd made are two small stylistic issues:

1. The comment above pallene_getfieldslot goes over our 80-column limit. 
2. In Pallene, we format else statemtents with the brackets surrounding the else: `} else {`. This is a bit different from the Lua source code, where the `}` and the `else {` are on two separate lines.

ahmedkrmn:
What if we instead use this for the generated code:
```lua
TValue *slot = pallene_getfieldslot($field_len, $tab, $key);
```
and this for the implementation of the function `pallene_get_field_slot()`:
```c
static inline
TValue *pallene_get_fieldslot(size_t field_len, Table *tab, TString *key)
{
    if (field_len < LUAI_MAXSHORTLEN) { 
        static size_t cache = UINT_MAX;
        return pallene_getshortstr(tab, key, &cache);
    } else {
        return cast(TValue *, luaH_getstr(tab, key));
    }
}
```
Will the location of declaring the cache variable effect the compiler optimization?

hugomg:
>Will the location of declaring the cache variable effect the compiler
>optimization?

Declaring the cache variable inside getfieldslot would change the meaning of the function. We need to use a different cache for each different call site.

What we are doing here is called an "inline cache" and is useful when  we access the same kind of field for many similar tables.

For example, suppose we have a loop like this:

```
for i = 1, N do
    local p = ps[i]
    sum = sum + p.x
    sum = sum + p.y
```

We expect that all the ps might have been created in the same place, and that the internal placement of the fields inside the hash table will be the same. So we cache the internal index of where we expect the xs to be, and we also keep another cache for where we expect the ys to be.

-----

By the way, one thing that I also noticed now is that pallene_getshortstr is only called from pallene_getfieldslot. Do you think it is clearer to keep them separate, or should we just put all the code inside getfieldslot?

ahmedkrmn:
@hugomg But isn't that what we were going to do if we used preprocessor directives:
```c
#define pallene_getfieldslot(field_len, tab, key, slot) \
    if (field_len < LUAI_MAXSHORTLEN) { \
        static size_t cache = UINT_MAX; \
        slot = pallene_getshortstr(tab, key, &cache); \
    } \
    else { \
        slot = cast(TValue *, luaH_getstr(tab, key)); \
    }
```

hugomg:
Functions which are inlined and preprocessor macros are similar in many ways, but `static` variables is one aspect where they behave differently.

ahmedkrmn:
@hugomg Interesting!

ahmedkrmn:
Done.
I've thought about putting the complete code of `pallene_getshortstr()` inside `pallene_get_fieldslot()` as you've proposed, but I believe that having them separated is a nicer idea for better code abstraction and maintenance.

hugomg:
> but I believe that having them separated is a nicer idea for better code abstraction and maintenance.

That is a good point. I agree that it did look cleaner in a separate function.

What do you think about renaming `pallene_get_fieldslot` to `pallene_getstr`, and moving it next to  after `pallene_getshortstr`? I think that could make it even clearer. Then we could move on to merging this.

ahmedkrmn:
> pallene_getshortstr

Perfect!

ahmedkrmn:
> > but I believe that having them separated is a nicer idea for better code abstraction and maintenance.
> 
> That is a good point. I agree that it did look cleaner in a separate function.
> 
> What do you think about renaming `pallene_get_fieldslot` to `pallene_getstr`, and moving it next to `pallene_getshortstr`? I think that could make it even clearer. Then we could move on to merging this.

Perfect!

PR #177, opened 2020-02-18
Multiple return values

### LINKED ISSUE #152: Implement functions with multiple return values
Currently Pallene only allows functions that return a single value. We should lift this restriction at some point

- [x] Parser: return statements need to allow 2+ return values separated by commas
- [x] Parser: variable declarations and assignments need to allow 2+ return values
- [ ] Parser: toplevel variable declarations need to allow 2+ variables. This has the ripple effect that the ast.toplevel_name gets a bit weird. I'd suggest getting rid of that function by changing the "dubplicate toplevel names" error to be more specific and separate the "duplicate names in letrec group" and "two exported functions with the same name".
- [ ] Type checker: many parts of the type checker will need to be refactored to use lists of expressions instead of a single expression. This should allow us to remove the `types.T.Void` type, which is currently a hack we use for functions without a return type.
- [ ] Coder: the Pallene function calling convention will need to be extended to allow functions with multiple return values.

Conceptually this should all be simple, but actually implementing this is going to be a bit more challenging because it will require refactoring a lot of code and it it is likely that we will find lots of little places in the code that currently only work because they assume that functions return a single value.

mambrozio:
Work in progress. Fixes #152


pallene/to_ir.lua
```diff
@@ -89,52 +89,56 @@ function ToIR:convert_stat(cmds, stat)
...
-                error("impossible")
-            end
-
-        elseif var._tag == "ast.Var.Bracket" then
-            local arr = self:exp_to_value(cmds, var.t)
-            local i   = self:exp_to_value(cmds, var.k)
-            local v   = self:exp_to_value(cmds, exp)
-            local src_typ = exp._type
-            table.insert(cmds, ir.Cmd.SetArr(stat.loc, src_typ, arr, i, v))
+        for i = 1, #stat.vars do
```
hugomg: We should be careful to keep the order of evaluation the same as Lua. https://www.lua.org/manual/5.3/manual.html#3.3.3

pallene/ast.lua
```diff
@@ -72,6 +72,7 @@ declare_type("Exp", {
     Binop      = {"loc", "lhs", "op", "rhs"},
     Cast       = {"loc", "exp", "target"},
     Paren      = {"loc", "exp"},
+    ExtraRet   = {"loc", "call_exp", "i", "n", "_type"}, -- See checker.lua
```
hugomg, line 75: Does `_type` need to be a parameter to ExtraRet's constructor? In all other nodes, the `_type` attribute is separately added by checker.lua and is not a parameter to the constructor.

mambrozio: It should be possible to remove it from the constructor and add `_type` to the new node immediately after the call to the constructor. It sure makes more sense now that you asked. I'll change this in the next commit.

pallene/to_ir.lua
mambrozio: It should be working the same as Lua. I wrote some tests to verify that too, with similar examples to the ones in Lua's manual.

pallene/checker.lua
```diff
@@ -391,14 +463,52 @@ function FunChecker:check_function(lambda, func_typ)
...
 end
 
+function FunChecker:check_and_declare_var(decl, exp)
+    local typ
+    typ, exp = self:check_initializer_exp(decl,
+        exp, "declaration of local variable %s",
+        decl.name)
+    self:add_local(decl.name, typ)
+    decl._name = self.p.symbol_table:find_symbol(decl.name)
+    return exp
```
hugomg, line 473: This code isn't implementing the scope correctly. The variables being declared should NOT be in scope for subsequent initializers.

Please consider the following test case. It should be allowed but currently it is gives a type error:

```
function test()
    local a: integer = 10
    local a: float, b: integer = 3.14, a
end
```

pallene/checker.lua
```diff
@@ -274,16 +319,43 @@ function Checker:check_program(prog_ast)
...
+                local exps = {}
+                for i = 1, #tl_var.decls do
+                    local decl = tl_var.decls[i]
+                    local exp = tl_var.values[i]
+                    local name = decl.name
+                    local typ
+                    typ, exp = toplevel_fun_checker:check_initializer_exp(
+                                    decl, exp,
+                                    "declaration of module variable %s", name)
+                    local _ = self:add_global(name, typ)
```
hugomg, line 352: This is implementing the scopes incorrectly. It is basically the same problem that is happening in Stat.Decl.

The current code accepts the following test case but it should actualy give an error saying that "variable 'a' is not declared"
```
local a: integer, b: integer = 10, a
```

pallene/checker.lua
```diff
@@ -391,14 +463,52 @@ function FunChecker:check_function(lambda, func_typ)
...
+                    end
+                    table.insert(stat.exps, exp)
+                end
+            end
+            if #stat.decls ~= #stat.exps then
+                type_error(stat.loc,
+                    "left-hand side expects %d value(s) but right-hand " ..
+                    "side produces %d value(s)", #stat.decls, #stat.exps)
+            end
+        end
```
hugomg, line 511: I wonder if there would be a way to implement Stat.Decl and Toplevel.Var in a similar manner. They both do basically the same thing so having different implementations smells funny.

BTW, one of the prerequisites for implementing closures in the future will be to change the Toplevel.Var to be Stat.Decls instead. Having a similar implementation for both will also help with that.

pallene/checker.lua
```diff
@@ -765,19 +913,53 @@ function FunChecker:check_exp_synthesize(exp)
...
-            if #f_type.ret_types >= 1 then
+
+            if #f_type.ret_types == 1 or
+               (#f_type.ret_types > 1 and exp._single_ret)
+            then
                 exp._type = f_type.ret_types[1]
+
+            elseif #f_type.ret_types > 1 and not exp._single_ret then
+                exp._type  = f_type.ret_types[1]
+                exp._types = f_type.ret_types
```
hugomg, line 961: Would there be a problem if we got rid of the `_single_ret` and always set the `_types` for all CallFuncs?

mambrozio: Currently this is used to know, in `to_ir`, if `ast.Exp.CallFunc` is the last call and we need to create the `ir.Cmd.CallStatic`. For example, if you write `return f()` and `f()` returns an integer, `_types` will not be set and this is the information we need in the node to see that we need to create the call in that evaluation. Same thing happens with a function call inside an `ast.Exp.Paren`. If we put `_types` in every function call we'll miss that information in `to_ir` and I couldn't think of another way to do that. This was the way I got it to work, but I understand that it is not very clean, especially in `ast.Exp.Paren`. I'm thinking about a new way to do this while I'm fixing the other issues that you pointed out.

pallene/checker.lua
mambrozio: I didn't think about this case. Already fixed it.

pallene/checker.lua
mambrozio: I didn't think about this case. Already fixed it.

pallene/checker.lua
```diff
@@ -869,11 +1063,19 @@ function FunChecker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
     elseif tag == "ast.Exp.Lambda" then
         error("not implemented yet")
 
+    elseif tag == "ast.Exp.Paren" then
+        if exp.exp._tag == "ast.Exp.CallFunc" or
+           exp.exp._tag == "ast.Exp.CallMethod"
+        then
+            exp.exp._single_ret = true
+        end
```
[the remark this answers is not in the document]
mambrozio: I agree. I'm thinking on how to change this behaviour.

pallene/checker.lua
```diff
@@ -274,16 +319,43 @@ function Checker:check_program(prog_ast)
...
-                local var = ast.Var.Name(loc, name)
-                toplevel_fun_checker:check_var(var)
-                table.insert(toplevel_stats,
-                    ast.Stat.Assign(loc, var, exp))
+
+                local last_val = tl_var.values[#tl_var.values]
+                if last_val._tag == "ast.Exp.CallFunc" then
+                    -- print(ii(last_val))
+                    last_val = toplevel_fun_checker:check_exp_synthesize(last_val)
+                    if last_val._types and #last_val._types > 1 then
```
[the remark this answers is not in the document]
mambrozio: 1 - Makes sense, but currently `_types` not being set has a meaning (see other comments)
2 - I didn't bother because I saw that it actually doesn't, but fixed it
3 - Absolutely! I forgot to remove those from an older code I was working on, sorry!

pallene/checker.lua
```diff
@@ -274,16 +319,43 @@ function Checker:check_program(prog_ast)
...
-                    "declaration of module variable %s", name)
-                local _ = self:add_global(name, typ)
-                local var = ast.Var.Name(loc, name)
-                toplevel_fun_checker:check_var(var)
-                table.insert(toplevel_stats,
-                    ast.Stat.Assign(loc, var, exp))
+
+                local last_val = tl_var.values[#tl_var.values]
+                if last_val._tag == "ast.Exp.CallFunc" then
+                    -- print(ii(last_val))
```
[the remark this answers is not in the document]
mambrozio: Sorry!

pallene/checker.lua
```diff
@@ -274,16 +319,43 @@ function Checker:check_program(prog_ast)
...
+                    last_val = toplevel_fun_checker:check_exp_synthesize(last_val)
+                    if last_val._types and #last_val._types > 1 then
+                        for i = 2, #last_val._types do
+                            local nval= ast.Exp.ExtraRet(last_val.loc, last_val, i,
+                                            #last_val._types)
+                            nval._type = last_val._types[i]
```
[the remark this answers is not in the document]
mambrozio: Changed it.

pallene/checker.lua
```diff
@@ -17,6 +17,23 @@ local checker = {}
...
+--
+--
+
+-- ast.Exp.ExtraRet is a node that represents an extra return value of a
+--     function call. This node is added to keep information about which return
+--     value must be used.
+-- @loc: This is the location of the function call
+-- @call_exp: This is the actual function call
+-- @i: This is the index of the return value of this node
+-- @n: This is the total number of return values of the function
```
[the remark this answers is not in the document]
mambrozio: In this current setting it is the total number of return values. I'm changing this to be only the number of values needed.

doc/manual.md
```diff
@@ -367,14 +369,16 @@ As usual, {A} means 0 or more As, and \[A\] means an optional A.
...
 
-    toplevelvar ::= local NAME [':' type] '=' exp
+    toplevelvar ::= local NAME [':' type] {',' NAME [':' type]} '=' explist
 
-    toplevelfunc ::= [local] function NAME '(' [paramlists] ')'  [':' type] block end
+    toplevelfunc ::= [local] function NAME '(' [paramlists] ')'  [':' ['('] typelist [')']] block end
 
     paramlist ::= NAME ':' type {',' NAME ':' type}
 
     type ::= nil | integer | float | boolean | string | any | '{' type '}' | NAME
```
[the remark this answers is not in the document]
mambrozio: I'm not sure I understood

doc/manual.md
```diff
@@ -367,14 +369,16 @@ As usual, {A} means 0 or more As, and \[A\] means an optional A.
...
 
+    typelist ::= type {',' type}
```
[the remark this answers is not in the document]
mambrozio: Much better, thanks! Fixed it

doc/manual.md
```diff
@@ -242,24 +240,28 @@ end
...
 end
 ```
 
 A `local` function is only visible inside the module it is defined.
-Non-local functions exported, which means that they are accessible to Lua if it requires the Pallene module.
+Non-local functions are exported, which means that they are accessible to Lua if it requires the Pallene module.
 
 As with variables, `<name>` can be any valid identifier, but it is a compile-time error to declare two functions with the same name, or a function with the same name as a module variable.
-The return types `<rettypes>` are optional, and if not given it is assumed that the function does not return anything.
+The return types `<rettypes>` are optional, and if not given it is assumed that the function does not return anything. If two or more return types are present, a parenthesis surrounding `<rettypes>` is obligatory.
```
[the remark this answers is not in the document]
mambrozio: Fixed

pallene/checker.lua
```diff
@@ -193,21 +207,52 @@ local letrec_groups = {
...
+            elseif tag == "ast.Toplevel.Import" then
+                self:check_top_level_name(names, ast.toplevel_name(tl_node),
+                    tl_node.loc)
+
+            elseif tag == "ast.Toplevel.Builtin" then
+                self:check_top_level_name(names, ast.toplevel_name(tl_node),
+                    tl_node.loc)
+
+            else
+                error("impossible")
```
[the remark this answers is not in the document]
mambrozio: Fixed.

pallene/checker.lua
mambrozio: Makes sense. Fixed.

pallene/to_ir.lua
```diff
@@ -87,52 +90,115 @@ function ToIR:convert_stat(cmds, stat)
...
-        local var = stat.var
-        local exp = stat.exp
-        if     var._tag == "ast.Var.Name" then
-            local cname = var._name
-            if     cname._tag == "checker.Name.Local" then
-                self:exp_to_assignment(cmds, cname.id, exp)
-            elseif cname._tag == "checker.Name.Global" then
-                local v = self:exp_to_value(cmds, exp)
-                table.insert(cmds, ir.Cmd.SetGlobal(stat.loc, cname.id, v))
+        local lvars = { locals = {}, globals = {}, brackets = {}, dots = {} }
```
hugomg, line 93: Unfortunately, I think this code still has a few bugs. Firstly, there is the big problem of using `pairs` instead of `ipairs`, which means that this algorithm is non-deterministic. Secondly, there are some situations where it should be saving things in temporary variables but it isn't (which can be hard to reproduce due to the pairs nondeterminism). Finally, it also generates temporary variables where it isn't necessary, which boats the generated code.

For an example of the first two problems, consider the following example:

```
function f(a: {integer}, b:{integer})
    a, a[1] = b, 42
end
```

Depending on how the `pairs` orders things, this can results in the following IR, which does the wrong thing. (It assigns to the `b` array while it should be assigning to the `a` array). To reproduce, run `./pallenec --emit-c` multiple times until the `pairs` orders things the buggy and then compile it with `./pallenec --compile-c`.

```
tmp1 = b
tmp2 = 1
tmp3 = 30
a = tmp1
a[tmp2] = tmp3
```

The other more minor problem is that the new version of the code always generates temporary variables, even when not strictly necessary. For example, a simple assignment like `a = b` becomes

```
tmp1 = b
a = tmp1
```

------------------------------------------------------

But I have good  news! I have investigated how Lua implements multiple assignments and figured out what we need to do to generate code that is both correct and without unnecessary temporary variables. I also implemented and uploaded the code that does that to the "hugo-multiple" branch. Check it out! I figured that it would be clearer to just write down the code instead of giving detailed explanation of the algorithm. Feel free to use that code in the PR but if you do please look carefully through it as if you were the one that were doing the code review this time. :)

One thing that I did not do is add extra tests. I would appreciate if you could write some additional tests to cover all the places where we call the `save_if_necessary` function. For example, we would need tests for `a, a[i] = ` and `a[i], a = `. 

We might also want to add a test for `x, x = 10, 20` to verify that it returns 10 just like Lua does.

pallene/ir.lua
```diff
@@ -165,11 +165,11 @@ local ir_cmd_constructors = {
...
-    BuiltinMathSqrt   = {"loc", "dst", "src"},
-    BuiltinStringChar = {"loc", "dst", "src"},
-    BuiltinStringSub  = {"loc", "dst", "src1", "src2", "src3"},
-    BuiltinToFloat    = {"loc", "dst", "src"},
-    BuiltinType       = {"loc", "dst", "src"},
+    BuiltinMathSqrt   = {"loc", "dsts", "src"},
+    BuiltinStringChar = {"loc", "dsts", "src"},
+    BuiltinStringSub  = {"loc", "dsts", "src1", "src2", "src3"},
+    BuiltinToFloat    = {"loc", "dsts", "src"},
+    BuiltinType       = {"loc", "dsts", "src"},
```
hugomg, line 172: If we are going to change the dsts we might as well change the src1 .. srcN to srcs as well. (And while you are at it, also remove the references to those field named from the `get_srcs` function).

pallene/checker.lua
```diff
@@ -765,19 +860,33 @@ function FunChecker:check_exp_synthesize(exp)
...
-                    exp.args[i], f_type.arg_types[i],
-                    "argument %d of call to function", i)
-            end
-            assert(#f_type.ret_types <= 1)
-            if #f_type.ret_types >= 1 then
-                exp._type = f_type.ret_types[1]
+
+            if #f_type.ret_types > 0 then
+                exp._type  = f_type.ret_types[1]
+                exp._types = f_type.ret_types
```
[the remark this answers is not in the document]
hugomg: I think this comment needs a clarification because it wasn't 100% accurate:

* Every function call expression should receive a `._types`, even when `#ret_types == 0`.
* There should be no need to test `if fexp._types` if we know that `fexp` is a function call expression. Every function call expression should have a `_types` now.

pallene/checker.lua
```diff
@@ -668,7 +668,14 @@ end
...
 function FunChecker:check_exp_synthesize(exp)
     local tag = exp._tag
-    if     tag == "ast.Exp.Nil" then
+
+    if exp._type then
+        -- Fallthrough
+        -- Expression already synthesized. This occurs when
+        -- expand_function_returns is called. This ensures the work is done
+        -- only once for call function and call method expressions.
+
```
hugomg, line 677: I'd rather leave this as a separate if statement. All the other parts of the if-elseif are checking the tag except this one. Suggestion:
```
if exp._type then
    -- This expression was already type-checked before. Probably because of
    -- expand_function_returns.
    return exp
end

local tag = exp._tag
if     tag == "ast.Exp.Nil" then
```

pallene/ir.lua
```diff
@@ -196,7 +196,7 @@ declare_type("Seq", {
 })
 
 local  src_fields = {
-    "src", "src1", "src2", "src3",
+    "src", "src1", "src2", "srcs",
```
hugomg, line 187: "srcs" belongs to the srcs_fields list. No need to add it here.

PR #187, opened 2020-03-09
Create type builtin function

### LINKED ISSUE #186: Create `type` builtin function
To ease the conversion of Lua code to Pallene, we might want to create a `type` builtin function, that receives an `any` and returns a `string`.

```
    function is_boolean(x: any)
        return type(x) == "boolean"
    end
```

mambrozio:
Fix #186

### DIFF
--- a/pallene/builtins.lua
+++ b/pallene/builtins.lua
@@ -15,6 +15,7 @@ for lua_name, typ in pairs({
     ["string.char"]  = T.Function({T.Integer()}, {T.String()}),
     ["string.sub"]   = T.Function({T.String(), T.Integer(), T.Integer()}, {T.String()}),
     ["tofloat"]      = T.Function({T.Integer()}, {T.Float()}),
+    ["btype"]         = T.Function({T.Any()}, {T.String()}),
 }) do
     local pallene_name = string.gsub(lua_name, "%.", "_")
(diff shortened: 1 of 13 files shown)


pallene/builtins.lua
```diff
@@ -15,6 +15,7 @@ for lua_name, typ in pairs({
     ["string.char"]  = T.Function({T.Integer()}, {T.String()}),
     ["string.sub"]   = T.Function({T.String(), T.Integer(), T.Integer()}, {T.String()}),
     ["tofloat"]      = T.Function({T.Integer()}, {T.Float()}),
+    ["btype"]         = T.Function({T.Any()}, {T.String()}),
```
hugomg: Will it be possible to call this function "type" instead of "btype"?

gligneul: I would like it to be called `type` as well.

gligneul: There is no problem renaming the current `type` to `typealias` or even `alias`.

hugomg: Agreed. I would prefer renaming the current `type` to `typealias`.

mambrozio: Changed to `typealias` in the new commit.

mambrozio:
Updated manual with the last commit.

PR #188, opened 2020-03-11
Updated README.md and implemented three examples in Pallene

itssamuelrowe:
I've added a note on the LuaRocks, Lua, and Pallene version dependency problem users may encounter when building Pallene.

Further, I've also added three examples to the repository and a shell script for executing them.


itssamuelrowe:
I've also added three examples to the repository and a shell script for executing them.

README.md
```diff
@@ -26,6 +26,12 @@ the [LuaRocks](http://luarocks.org) package manager:
 $ luarocks install --local --only-deps pallene-dev-1.rockspec
 ```
 
+> As of now, LuaRocks is configured to use Lua 5.1 by default. However, Pallene
+> requires Lua 5.3 to run. Therefore, please configure LuaRocks to use Lua 5.3.
+> You can use the following command to configure LuaRocks to use Lua 5.3 when
+> compiling it:
+> `./configure --lua-version=5.3`
+
```
gligneul: Instead of this quote-like paragraph, you could write a section called `### Prerequisites`. In this section, you should mention that the user needs Lua 5.3 installed and then he/she should compile LuaRocks. It would be nice to provide links to where you can get Lua and LuaRocks. Only after that you should explain the issue with Lua 5.1. It is also important to notice that this issue should only happen when you have Lua 5.1 installed, which probably is the case for Ubuntu users.

examples/arithmetic_main.lua
```diff
@@ -0,0 +1,4 @@
+arithmetic = require "examples.arithmetic"
```
gligneul: Each example could be inside its own dir. For instance
```
examples/arithmetic/arithmetic.pln
examples/arithmetic/main.lua
```
And so on.

examples/arithmetic.pln
```diff
@@ -0,0 +1,7 @@
+function add(a: any, b: any): integer
```
gligneul: Could you write a arithmetic example that shows a little more of Pallene? Something like factorial or Fibonacci would be nice.

examples/rectangle.pln
```diff
@@ -0,0 +1,10 @@
+-- The following program demonstrates using table types in Pallene.
```
gligneul: Please add the copyright notice to all files.

run_examples
```diff
@@ -0,0 +1,11 @@
+echo "[info] Running sum of array"
```
gligneul: Instead of this script, you could add a read-me to the examples directory.

examples/arithmetic_main.lua
itssamuelrowe: Okay, done.

examples/arithmetic.pln
itssamuelrowe: Yeah, sure. I'll add a few more examples shortly.

PS: The arithmetic.pln program was written to demonstrate the `any` keyword. I'll try to add comments which describe the purpose of each example.

examples/rectangle.pln
itssamuelrowe: Okay, done.

README.md
itssamuelrowe: I've made the requested changes.

run_examples
itssamuelrowe: I can add the README file. But removing the script means the user has to manually execute each example. Wouldn't that be inconvenient?

examples/arithmetic.pln
gligneul: A example with any will make more sense if you use the recently-added `type` function.

run_examples
gligneul: I believe that the purpose of the examples would be to check out the code, change a little and then execute it. I don't see the reason to execute all the examples at once like they were tests or benchmarks.

itssamuelrowe: Fair enough. I'll remove the script.

examples/arithmetic.pln
itssamuelrowe: Where can I find the documentation for the type function? I am not able to find it in the manual.

gligneul: We could add this later.

README.md
```diff
@@ -17,6 +17,36 @@ Compared to LuaJIT, Pallene aims to offer more predictable run-time performance.
 In order to use this source distribution of the Pallene compiler, you need to
 install its Lua library dependencies and compile its run-time library.
 
+### Prerequisites
+
+Before you build Pallene, you need to install `gcc` and `make` on your system.
```
gligneul:
```suggestion
Before you build Pallene, you need to install a C compler (e.g. `gcc`) and `make` on your system.
```

README.md
```diff
@@ -17,6 +17,36 @@ Compared to LuaJIT, Pallene aims to offer more predictable run-time performance.
...
+sudo apt-get update
+sudo apt-get install build-essential
+```
+
+Linux users using pacman can run the following commands.
+```
+pacman -Syyu
+pacman -S make
+pacman -S gcc
+```
```
gligneul: I think you can remove this part, we could leave the installation instructions just for Ubuntu as an illustration. Specially because Arch users are usually more advanced and know how to install packages by their own.

README.md
```diff
@@ -17,6 +17,36 @@ Compared to LuaJIT, Pallene aims to offer more predictable run-time performance.
...
+
+After Lua 5.3 is installed, download the source code of LuaRocks from 
+[https://github.com/luarocks/luarocks/releases](https://github.com/luarocks/luarocks/releases). Follow
+the build instructions appropriate for your platform and install it on your system.
+
+If LuaRocks is configured to use older versions of Lua, you may not be able to
+install the dependencies as described in the next section. Therefore, please
+configure LuaRocks to use Lua 5.3. You can use the following command to configure
+LuaRocks to use Lua 5.3 when compiling it:
+`./configure --lua-version=5.3`
```
gligneul: Very nice!

examples/arithmetic.pln
itssamuelrowe: Yeah, alright.

examples/arithmetic/arithmetic.pln
```diff
@@ -0,0 +1,16 @@
+-- Copyright (c) 2020, The Pallene Developers
+-- Pallene is licensed under the MIT license.
+-- Please refer to the LICENSE and AUTHORS files for details
+-- SPDX-License-Identifier: MIT
+
+function add(a: any, b: any): integer
+    if a.type() != "integer" then
```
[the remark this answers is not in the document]
itssamuelrowe: I actually tried `type(a)` and `a.type()`. Both did not work. I forgot to remove the reference to `type()`. Will remove it.

README.md
itssamuelrowe: I'll remove it.

PS: You make Arch Linux users look cool! :P

ahmedkrmn:
@gligneul and @itssamuelrowe, there is a typo in **compler** in line 22 of the README.

gligneul:
@ahmedkrmn this PR is merged, feel free to create another PR or an issue.

ahmedkrmn:
@gligneul #192

PR #190, opened 2020-03-17
Compiler options

mambrozio:
This PR adds the following compiler options:

- --dump-ast: Print ast from a .pln file and exit
- --dump-checker: Print checker's ast from a .pln file and exit
- --dump-ir: Print ir's ast from a .pln file and exit
- --dump-uninitialized: Print uninitialized's ast from a .pln file and exit
- --dump-constant_propagation: Print constant_propagation's ast from a .pln file and exit

I believe this will be useful when developing new features in the compiler.


pallenec
```diff
@@ -28,10 +34,24 @@ local function compile(in_ext, out_ext)
     end
 end
 
+local function compile_internal(filename, stop_after)
+    local ok, errs = driver.compile_internal(filename, stop_after, true)
+    if not ok then
+        util.abort(table.concat(errs, "\n"))
+    end
+end
```
gligneul: Couldn't you add the `inspect` call inside this function instead of adding a new argument?

pallenec
```diff
@@ -10,6 +10,12 @@ p:argument("source_file", "File to compile")
 p:flag("--emit-c", "Generate a .c file instead of an .so file")
 p:flag("--emit-asm", "Generate a .s file instead of an .so file")
 p:flag("--compile-c", "Compile a .c file generated by --emit-c")
+p:flag("--dump-ast", "Print ast from a .pln file")
```
gligneul: This could be an options instead of multiple flags. For instance, `--dump=ast`.

pallenec
mambrozio: Sure. I thought of doing it this way, but did not know which was best, requiring `inspect` in pallenec or in the driver. I chose the driver to keep the pallenec script cleaner.

pallenec
mambrozio: At first I thought of doing with options, but when talking to Hugo about this he said to do it like this. I will read the documentation on options and make the change.

gligneul: Ask Hugo before changing to option, if he agrees then make the chage. Else keep like this.

pallenec
gligneul: I think it is better to add to pallenec because it will only be used there. Also, there will be less code repetiton.

mambrozio: Ok, I'll prepare this change.

pallenec
hugomg: One nice thing about having  a single `--dump` flag is that it would make it clearer that it is only allowed to dump one thing at a time.

But there should be a way for an user running `pallenec --help` to find out what are the valid values that can be passed to `--dump`.

mambrozio: It is possible to limit choices for the --dump option, but this is only possible on the latest version of argparse, meaning that I would have to bump the requirement on the rockspec to at version 0.7.0 or up. If it's ok to bump the requirement, I'll make the change.

gligneul: Upgrading the lib is not a problem

PR #193, opened 2020-03-30
Mention in readme that we need the Lua headers

hugomg:


ahmedkrmn:
Can I add this?

gligneul:
Do we really need the headers? Don't we use the headers from our version of Lua?

hugomg:
Exactly. This is for the dependencies we install with luarocks.

PR #196, opened 2020-04-01
Run luacheck before busted in travis.yml

hugomg:
Running Luacheck is faster than running the whole test suite, so we might as
well do it first.


hugomg:
Thanks!

PR #199, opened 2020-04-15
slot obtention could be decoupled from Get/SetArr

leokaplan:


hugomg:
I think it is a bit strange to open a PR if the only code is a pair of TODO comments that will be removed by the time the PR is complete. This could habe been opened, an issue instead.

But back to the topic...

1. Are we also going to replace SetTable and GetTable?
2. How do we ensure that slots are valid, and pointers are not used after a table is reallocated?

hugomg:
I'm closing this to clean up the list of PRs. We can create a new PR for this later.

PR #201, opened 2020-04-23
Prettier pointer variable declarations in C

hugomg:
We now do not leave a space after the `*` when declaring a pointer variable.

Before: `TString * x1;`
After:  `TString *x1;`


hugomg:
FYI, the build error is the same `luarocks install busted` problem that we are encountering in PR #177.

PR #204, opened 2020-04-24
Update Lua to version 5.4-rc1

### LINKED ISSUE #200: Update Pallene to Lua 5.4 rc-1
It looks like Lua 5.4 rc-1 changed a bunch of stuff. I expect that we'll need to update Pallene accordingly.

hugomg:
Updates our version of Lua from 5.4-beta to 5.4-rc1. Fixes #200.

The main change is that now the type variants are named LUA_VFOO instead of LUA_TFOO. For example LUA_VNUMINT instead of LUA_TNUMINT.

The other important change is that now the boolean type has two variants, LUA_VTRUE and LUA_VFALSE. The value of the boolean is now completely encoded by the type tag.


hugomg:
Don't forget to recompile your Lua after this gets merged :)

PR #216, opened 2020-05-19
New IR pretty printer

hugomg:
Add an ir.pretty_print function that outputs a human-readable version of
the IR. This representation may be helpful for debugging because it is
significantly more concise than --dump=ir or --emit-c.

The main limitation is that we don't show any types. I couldn't think of a
clean way to show them so I just left them out.


hugomg:
I'm opening this as a draft PR because I'd like to give people a chance to play around with it. I'm specially interested in knowing if it is really worth adding this to help with development of optimization at the IR level. Is this genuinely useful for people working with Pallene's IR? I'm looking at you, @leokaplan :)

If we do conclude that this is worth adding then there are some things that would need to be ironed out before this can be merged:

1. We'd have to add a tests. The pretty printer will need to be updated whenever there are changes to the IR so it would be important to have tests with good coverage.
2. We'd have to decide if the current output format is sufficient or if we also need to find a way to show other things (such as the types of variables).
3. I exposed this functionality as a `--print-ir` flag but I am not sure that is necessarily the best way to do it.

hugomg:
I'd also appreciate any suggestions in the direction of simplifying the code and making it more maintainable. This feature would be purely to help Pallene's development so I don't want the cost of maintaining it to be higher than the benefit that it provides.

hugomg:
Matheus asked about showing the original variable names instead of showing the variable identifiers (x1, x2, x3)... Could be something worth experimenting with.

hugomg:
I haven't felt the need to change this. I'm open to merging this in to put it in the master branch, where it is easier to use. We can always update or even delete it later if we change our minds.

hugomg:
Added tests. They increase the time to run the test suite from 7 seconds to 8 seconds but might be a worthwhile tradeoff in order to get good coverage.

PR #217, opened 2020-05-19
Add nbody to the cbenchmarks makefile

hugomg:
When we added the nbody benchmark in PR #127 we added it to the makefile but forgot to add it to the default make target. This means that `make` and `make clean` do not work as it should.

This commit fixes that oversight.


hugomg:
While I was testing the nbody problem I also found other problems in the code... So more commits to fix those as well.

PR #218, opened 2020-05-30
Increase column limit from 80 to 100

hugomg:
I was wondering what things would look if we had an 100 column limit instead of an 80 column one so I created this PR. I'd like to hear your opinion on it, specially @gqligneul, who I know cared about the previous column limit.

My impression of the change was that a wider limit of 100 reduces the number of times where we need to do weird contortions to fit inside the column limit, while still being narrow enough to fit multiple windows side-by-side and to keep block comments easy to read. I also noticed that there were a bunch of situations where we were already going over the 80 column limit but few situations where we got close to 100.

The aggregate diff is quite large but most of it are just comments being reformatted. It might be clearer to look at the diffs for the individual commits.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
-max_line_length = 80
+max_line_length = 100
 
 [*.md]
 max_line_length = off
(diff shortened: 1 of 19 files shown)


gligneul:
I still prefer the 80 column limit but given that I'm not that active on the project anymore you should increase the limit if you prefer.

hugomg:
I'll take that as an "OK" then :)

PR #223, opened 2020-06-02
Create repeated string literals only once

mambrozio:
The current code generates multiple userdata for strings that are equal. This PR changes that to
reuse the userdata string when it has the same value.

For example, in the following code:

```
function g()
  ... "var" ...
  ... "function" ...
end

function f()
  ...
  
  if     tag == "var"
    ...
  elseif tag == "function"
    ...
  end
  
  ...
end
```

Before this commit the result would have been:

```
int luaopen_a(lua_State *L)
{
  ...
  
  lua_newuserdatauv(L, 0, 12);
  int globals = lua_gettop(L);
  
  ...
  
  lua_pushstring(L, "var");

  lua_pushstring(L, "function");
  lua_setiuservalue(L, globals, 2);

  lua_pushstring(L, "var");          -- repeated string literal
  lua_setiuservalue(L, globals, 3);

  lua_pushstring(L, "function");     -- repeated string literal
  lua_setiuservalue(L, globals, 4);
  
  ...
```

The result with this PR would be:

```
int luaopen_a(lua_State *L)
{
  ...
  
  lua_newuserdatauv(L, 0, 12);
  int globals = lua_gettop(L);
  
  ...
  
  lua_pushstring(L, "var");

  lua_pushstring(L, "function");
  lua_setiuservalue(L, globals, 2);
  
  ...
```

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -652,8 +652,10 @@ function Coder:init_upvalues()
             for _, v in ipairs(ir.get_srcs(cmd)) do
                 if v._tag == "ir.Value.String" then
                     local str = v.value
-                    table.insert(self.upvalues, coder.Upvalue.String(str))
-                    self.upvalue_of_string[str] = #self.upvalues
+                    if not self.upvalue_of_string[str] then
+                        table.insert(self.upvalues, coder.Upvalue.String(str))
+                        self.upvalue_of_string[str] = #self.upvalues
+                    end
                 end
             end
         end


mambrozio:
@hugomg, I believe that there are no side effects with this change, but I only tested scenarios that I could think of. If you think of one that this breaks, please let me know so I can fix it.

hugomg:
I'm pretty sure that we used to reuse string literals but it seems that something broke along the way! Nice catch!

pallene/coder.lua
```diff
@@ -647,13 +647,17 @@ function Coder:init_upvalues()
...
     -- String Literals
+    local literals = {}
     for _, func in ipairs(self.module.functions) do
         for cmd in ir.iter(func.body) do
             for _, v in ipairs(ir.get_srcs(cmd)) do
                 if v._tag == "ir.Value.String" then
                     local str = v.value
-                    table.insert(self.upvalues, coder.Upvalue.String(str))
-                    self.upvalue_of_string[str] = #self.upvalues
+                    if not literals[str] then
```
hugomg, line 656: Do we need this `literals` table? Would it be possible to just use `self.upvalue_of_string` instead?

Issue #184, opened 2020-03-06
Should we use an explicit export statement instead of local/global to define what functions are exported?


hugomg:
For example, instead of 

```
function foo() end
function bar() end
```

we could use 

```
local function foo() end
local function bar() end
export foo, bar
```

The "local" specifier for would now be required for toplevel functions, making clear that they are not assigned to a global variable. To export the functions, we would use an explicit `export` list at the end of the module. 

The main thing that draws me to this idea is that it would allow function declaration statements to work the same in the toplevel and inside other functions, when we implement closures.  I also think the export statement would bring Pallene a bit closer to Lua. Pallene would stop giving special treatment to toplevel globals, and the export statement at the end of the module is similar to the `return` statement that Lua modules already use. What do you think?

ahmedkrmn:
I believe this is a very appropriate change that should be made.

gligneul:
> The "local" specifier for would now be required for toplevel functions, making clear that they are not assigned to a global variable. 

What happens when the user declares a function without the `local` modifier? I expect that the compiler creates a closure and sets it to the appropriate global variable. Are we implementing this behavior? Or will it be an error? 

---

> To export the functions, we would use an explicit export list at the end of the module.

I like the idea since it covers one way of implementing modules in Lua. A to-Lua compiler could just replace the export with a return statement. However, are we going to allow the export only as the last statement in the top level? What happens when the user write more than one export? For instance:
```
local function foo() end
export foo
local function bar() end
export bar
```

What happens when the user doesn't write the export statement? Will the compiler create an empty module? Will it be an error?

---

One more thing: when I'm programming in Lua I prefer to declare the module at the top of the file and declare the functions with the module name:
```
local my_module = {}
function my_module.f() end
return my_module
```
How is this going to be translated to Pallene? Even if we don't implement this feature at first, it is important to consider this case and preferably design how it is going to be. My suggestion is something like:
```
module my_module
function my_module.f() end
```

If the `export` statement is going to be a special statement that can only be written in the end of the top-level, the `module` statement could also be a special statement that is written in the beginning of the top-level.

gligneul:
I suggest that we decide what we are going to do, create the appropriate issues and then close this one.

hugomg:
> I expect that the compiler creates a closure and sets it to the appropriate global variable. Are we implementing this behavior? Or will it be an error?

[...]

PR #224, opened 2020-06-03
In-built modules can be indexed with the dot operator

itssamuelrowe:


.gitignore
```diff
@@ -9,3 +9,7 @@ lua/src/luac
 cbenchmarks/*
 !cbenchmarks/*.c
 !cbenchmarks/makefile
+
+test.pln
+test.so
+test.lua
```
hugomg, line 15: We already have a gitignore rule to ignore ".so" files.

pallene/ast.lua
```diff
@@ -83,7 +83,15 @@ declare_type("Field", {
 -- note: the following functions are why we need `if type(conss) == "table"` in parser.lua
 --
 
--- Return the variables names declared by a given toplevel node
+-- Returns a sequence containing the variable names declared by the specified
+-- toplevel node.
```
hugomg, line 87: It looks like your text editor is configured with lines of length 80?

pallene/ast.lua
```diff
@@ -83,7 +83,15 @@ declare_type("Field", {
...
+--
+-- The following nodes will result in a sequence with one element.
+--     ast.Toplevel.Func, ast.Toplevel.Typealias, ast.Toplevel.Typealias,
+--     ast.Toplevel.Record, ast.Toplevel.Import, and ast.Toplevel.Builtin 
+-- The following nodes will result in a sequence with one or more elements.
+--     ast.Toplevel.Var
+--
```
hugomg, line 94: I don't know if we need this part of the comment. In the places that use ast.toplevel_names the user should not care whether each case is returning one or more names.

pallene/checker.lua
```diff
@@ -53,25 +54,34 @@ function checker.check(prog_ast)
...
+-- and report a Pallene compilation error.
+local function checker_error(error_location, format, ...)
+    local error_message = location.format_error(error_location, format, ...)
+    coroutine.yield(error_message)
 end
 
-local function scope_error(loc, fmt, ...)
-    return checker_error(loc, ("scope error: " .. fmt), ...)
+local function scope_error(error_location, format, ...)
+    return checker_error(error_location, ("scope error: " .. format), ...)
```
hugomg, line 72: We are already using `loc` and `fmt` in other places in the code. I think it would be more consistent to keep the old names.

pallene/checker.lua
```diff
@@ -198,31 +212,63 @@ local letrec_groups = {
...
+--    a flat `ast.Var.Name`.
+--
+--
+-- In the AST, unions tagged with `the ast.Var` refer to things that can appear
+-- on the left-hand side of an assignment. That is, variable names such as
+-- `foo` (`ast.Var.Name`), bracketed expressions such as `foo[i]` (`ast.Var.Bracket`)
+-- and qualified names such as `foo.bar` (`ast.Var.Dot`).
+-- In a union tagged as `ast.Stat.Assign`, the `vars` field refers to the `ast.Var`
+-- nodes in the left-hand-side of the assignment statement. Since Pallene has multiple
+-- assignments there may be one or more of those nodes.
```
hugomg, line 236: The right place to put comments about the meaning of the `ast.*` data structures would be in the ast.lua file. That said, I am not sure that this particular paragraph is saying something that is not already being said by the grammar itself.

pallene/checker.lua
```diff
@@ -198,31 +212,63 @@ local letrec_groups = {
...
 function Checker:check_program(prog_ast)
 
     do
-        -- Forbid toplevel duplicates
+        -- Forbid top-level duplicates
+        --
+        -- Retrieve the names for all the top-level components such as functions,
+        -- imports, builtins and variables. Construct an auxillary table with name and location
+        -- as key and value, respectively. If the current name already exists in the auxillary
+        -- table, report an error.
```
hugomg, line 246: You could change this comment to say the  "why" instead of the "how".  Instead of talking about the auxiliary table, which is an implementation detail, it should say that we block duplicates to avoid ambiguities when exporting functions. We could also add a link to issue #184.

pallene/checker.lua
```diff
@@ -587,8 +637,11 @@ function FunChecker:check_var(var)
             var._type = self.p.module.functions[cname.id].typ
         elseif cname._tag == "checker.Name.Builtin" then
             var._type = builtins[cname.name].typ
-               else
-            error("impossible")
+        elseif cname._tag == "checker.Name.Module" then
+            -- The type for a module is a table.
+            var._type = builtin_modules[cname.name]
+        else
+            error("Unexpected condition: cname._tag == " .. cname._tag)
```
hugomg, line 644: Let's leave this as `error("impossible")` for now. I think it will be cleaner if we make this kind of change in a separate PR and if we change all the `error("imposible")`s at once.

pallene/ast.lua
hugomg: (Adding editorconfig support to your editor should take care of this)

.gitignore
itssamuelrowe: I added these lines for development. I'll remove them before we merge.

pallene/ast.lua
itssamuelrowe: Fair enough. I'll remove it.

pallene/checker.lua
itssamuelrowe: Alright. I'll revert the changes I made.

pallene/checker.lua
itssamuelrowe: > The right place to put comments about the meaning of the ast.* data structures would be in the ast.lua file.
I'll move it.

> That said, I am not sure that this particular paragraph is saying something that is not already being said by the grammar itself.
WIth that logic, our code wouldn't need any comments at all. :P
I'll remove it if you want though.

pallene/checker.lua
itssamuelrowe: Alright.

pallene/checker.lua
itssamuelrowe: Okay. I'll revert it.

.gitignore
hugomg: You could consider using `git add -u` instead of plain `git add`. That way it only adds changes to files that are already in the repository and ignores the extra test files that you create. No need to change the gitignore file.

pallene/checker.lua
```diff
@@ -11,7 +11,7 @@ local symtab = require "pallene.symtab"
 local types = require "pallene.types"
 local typedecl = require "pallene.typedecl"
 local util = require "pallene.util"
-local builtin_modules = require "pallene.builtin_modules"
+local inspect = require "inspect"
```
hugomg, line 14: Tip: you can require inspect in the same line where you use it. This way you don't forget about the `require "inspect"` at the top of the file after you delete the debugging print statement.

     print(require("inspect")(something))

pallene/builtins.lua
```diff
@@ -4,24 +4,45 @@
...
-        typ = typ,
+builtins.functions = {
+    io_write = {
+        name = "io.write",
+        typ = T.Function({ T.String() }, {})
+    },
+    tofloat = {
+        name = "tofloat",
+        typ = T.Function({ T.Integer() }, { T.Float() }),
+    },
```
hugomg, line 19: Do we need this `name` field? I suspect that we could just use key=name and value=type, as in the old builtins table:

    ["tofloat"] = T.Function({ T.Integer() }, { T.Float() })

pallene/checker.lua
```diff
@@ -621,6 +621,8 @@ end
 function FunChecker:check_var(var)
     local tag = var._tag
     if     tag == "ast.Var.Name" then
+        -- print(inspect(var))
+        -- print('------------')
```
hugomg, line 625: Please delete these comments.

pallene/checker.lua
```diff
@@ -636,29 +638,47 @@ function FunChecker:check_var(var)
         elseif cname._tag == "checker.Name.Function" then
             var._type = self.p.module.functions[cname.id].typ
         elseif cname._tag == "checker.Name.Builtin" then
-            var._type = builtins[cname.name].typ
+            var._type = builtins.functions[cname.name].typ
         elseif cname._tag == "checker.Name.Module" then
             -- The type for a module is a table.
-            var._type = builtin_modules[cname.name]
+           var._type = builtins.modules[cname.name]
```
hugomg, line 644: This case should be a type error, just like `checker.Name.Type`. This way, we will disallow things such as

    local x = io

By the way, don't forget to add a test case for that.

pallene/checker.lua
```diff
@@ -636,29 +638,47 @@ function FunChecker:check_var(var)
...
         else
-            error("Unexpected condition: cname._tag == " .. cname._tag)
+            error("impossible")
         end
 
+        -- print(inspect(var))
```
hugomg, line 649: More print comments. Please remove all of them. (Consider using `git add -p`, which makes it possible to leave them in the text file but not include them in the commit)

pallene/checker.lua
```diff
@@ -636,29 +638,47 @@ function FunChecker:check_var(var)
...
+            local function_name = var.name
+            local internal_name = module_name .. "." .. function_name
+            local cname = self.p.symbol_table:find_symbol(internal_name)
+            return {
+                _name = cname,
+                _tag = "ast.Var.Name",
+                loc = var.exp.loc,
+                name = var.exp.var.name .. '.' .. var.name,
+                _type = builtins.functions[internal_name].typ
+            }
```
hugomg, line 665: We should use the `ast.Var.Name` constructor instead of building this table by hand:

     local flat_var = ast.Var.Name(var.exp.loc, internal_name)
     flat_var._name = ...
     flat_var._type = ...

pallene/checker.lua
```diff
@@ -636,29 +638,47 @@ function FunChecker:check_var(var)
...
-                types.tostring(ind_type))
-        end
-        local field_type = types.indices(ind_type)[var.name]
-        if not field_type then
-            type_error(var.loc,
-                "field '%s' not found in type '%s'",
-                var.name, types.tostring(ind_type))
+        -- print(inspect(var))
+        -- print('---------------------')
+        if builtins.modules[var.exp.var.name] ~= nil then
```
hugomg, line 654: Before you access `exp.var` you need to check that the `exp` in question is an `ast.Exp.Var`. This line of code will blow up if the expression on the left of the dot is not a variable name. For example `xs[i].name`

By the way, if the existing test suite is not catching this bug then please add a test case that ensures that it does :)

pallene/types.lua
```diff
@@ -206,7 +206,9 @@ function types.consistent(t1, t2)
     return equivalent(t1, t2, true)
 end
 
+local inspect = require 'inspect'
 function types.tostring(t)
+    print(inspect(t))
```
hugomg, line 211: This inspect is printing a lot of junk when we run `busted`.

pallene/checker.lua
hugomg: Another thing: you can leave out the `~= nil`. In the rest of the code base we just use `if set[key] then` instead of `if set[key] ~= nil then`

pallene/ast.lua
itssamuelrowe: I manually formatted it to 80 columns. I'll add the editorconfig support ASAP.

pallene/checker.lua
itssamuelrowe: Alright. Will make the changes.

pallene/checker.lua
itssamuelrowe: Sure.

pallene/checker.lua
itssamuelrowe: Alright. Will start working on it tomorrow.

pallene/checker.lua
itssamuelrowe: Done.

pallene/types.lua
itssamuelrowe: I must have pushed it without commenting. Sorry. Will remove it.

pallene/checker.lua
```diff
@@ -586,27 +612,50 @@ function FunChecker:check_var(var)
...
-            type_error(var.loc,
-                "trying to access a member of value of type '%s'",
-                types.tostring(ind_type))
-        end
-        local field_type = types.indices(ind_type)[var.name]
-        if not field_type then
-            type_error(var.loc,
-                "field '%s' not found in type '%s'",
-                var.name, types.tostring(ind_type))
+        if var.exp._tag == "ast.Exp.Var" and builtins.modules[var.exp.var.name] then
```
hugomg, line 627: This also needs to check if `var.exp.var` is a `ast.Var.Name`. It turns out that the example I had previously provided where `var.exp.var` is a `var.Bracket` works by chance because `builtins.modules[nil]` is `nil`. However, this code will do weird things if `var.exp.var` is an `ast.Var.Dot` because that case also has a `.name` field.

pallene/checker.lua
```diff
@@ -529,6 +554,7 @@ function FunChecker:check_stat(stat)
                         "attempting to assign to builtin function %s",
                         stat.vars[i].name)
                 end
+                -- TODO: Add checker.Name.Module to prevent module assignment
```
hugomg, line 557: Do we still need this TODO?

pallene/checker.lua
```diff
@@ -407,10 +429,13 @@ function FunChecker:expand_function_returns(lhs, rhs)
...
 end
 
-
 function FunChecker:check_function(lambda, func_typ)
     assert(lambda._tag == "ast.Exp.Lambda")
 
+    -- Check for duplicate parameter names.
+    --
+    -- Add each parameter name to the `names` set. If the current name already
+    -- exists in the set, report an error.
```
hugomg, line 438: We don't need to add a comment here because PR #222 got rid of this check for duplicate parameter names. I'm worried that git will gladly merge this comment into the master branch even though the block of code that it is commenting won't exist anymore.

.gitignore
```diff
@@ -8,4 +8,4 @@ lua/src/luac
 
 cbenchmarks/*
 !cbenchmarks/*.c
-!cbenchmarks/makefile
+!cbenchmarks/makefile
```
hugomg, line 11: We can make the diff a bit cleaner if we add back the newline.

PR #228, opened 2020-06-04
Parallelize make linux-readline in travis

hugomg:
This could potentially save about 3 seconds from the CI builds. Let's see...


hugomg:
No, it didn't make a difference.  ¯\\\_(ツ)\_/¯

PR #244, opened 2020-06-12
Export top-level components from Pallene module

itssamuelrowe:
This pull request implements the `export` modifier which allows top-level functions and variables to be exported from a Pallene module.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -297,7 +297,7 @@ end
 
 Unlike languages like C or Java, Pallene does not require type annotations on every variable.
 It uses a bidirectional type-checking system that is able to infer the types of almost all variables and expressions.
-Roughly speaking, you must include type annotations for the parameters and return types of toplevel functions, and almost everything else can be inferred from that.
+Roughly speaking, you must include type annotations for the parameters and return types of top-level functions, and almost everything else can be inferred from that.
 For example, notice how the `sum_floats` from the Brief Overview section does not include a type annotation for the `result` and `i` variables.
 
 If a local variable declaration doesn't have an initializer, it must have a type annotation:
@@ -357,6 +357,24 @@ insert(ns, "boom!")
 local x1 : integer = ns[1]
 local x2 : integer = ns[4] -- run-time error
 ```
+
+## Exporting top-level components
+
+The `export` keyword can be used to export top-level components such as variables and functions from a Pallene module.
+The keyword `export` can be used in the same places where the `local` keyword can be used.
+Every top-level component declaration should have either `local` or `export` modifier because Pallene does not support global variables.
+
+In the following example, we export `foo` and `name` from our module.
+```lua
+export function foo(x: integer): integer
+    return x + 1
+end
+
+export name: string = "Hello"
+```
+
+You can export on top-level components from a Pallene module.
+Further, it is an error to export the same name twice.
 
 ## The Complete Syntax of Pallene
 
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -249,7 +249,8 @@ local grammar = re.compile([[
     toplevelrecord  <- (P  RECORD NAME^NameRecord recordfields
                            END^EndRecord)                        -> ToplevelRecord
 
-    modifier        <- (LOCAL / EXPORT)                            -> opt_bool
+    modifier        <- LOCAL -> to_true
+                     / EXPORT -> to_false
 
     import          <- (P  LOCAL NAME^NameImport ASSIGN^AssignImport
                            IMPORT^ImportImport
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -236,11 +236,11 @@ local grammar = re.compile([[
                            / toplevelrecord
                            / import )* |} !.
 
-    toplevelfunc    <- (P  modifier FUNCTION NAME^NameFunc
+    toplevelfunc    <- (P  export_or_local FUNCTION NAME^NameFunc
                            LPAREN^LParPList paramlist RPAREN^RParPList
                            rettypeopt block END^EndFunc)         -> toplevel_func
 
-    toplevelvar     <- (P  modifier decllist ASSIGN^AssignVar
+    toplevelvar     <- (P  export_or_local decllist ASSIGN^AssignVar
                            !IMPORT explist1^ExpVarDec)           -> ToplevelVar
 
     typealias       <- (P  TYPEALIAS NAME^NameTypeAlias ASSIGN^AssignTypeAlias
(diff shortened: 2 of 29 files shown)


doc/manual.md
```diff
@@ -358,21 +358,40 @@ local x1 : integer = ns[1]
 local x2 : integer = ns[4] -- run-time error
 ```
 
+## Exporting top-level components
+
+The `export` keyword can be used to export top-level components such as variables
+and functions from a Pallene module. The keyword `export` can be used in the
+same places where the `local` keyword can be used. It should be noted that, every
+top-level component declaration should have either `local` or `export` modifier.
```
hugomg, line 366: To help keep the git diffs easy to follow, we try to put each sentence on a separate line. It is OK if the line is longer than 80 characters. Our editor config should have disabled to max_line_length for markdown files, let me know if that isn't working for you.

doc/manual.md
```diff
@@ -358,21 +358,40 @@ local x1 : integer = ns[1]
...
+You can export on top-level components from a Pallene module. Further, it is an
+error to export the same name twice.
+
 ## The Complete Syntax of Pallene
 
 Here is the complete syntax of Pallene in extended BNF.
 As usual, {A} means 0 or more As, and \[A\] means an optional A.
 
-    program ::= {toplevelrecord | topleveltypealias | toplevelvar | toplevelfunc}
+    program ::= {top-levelrecord | top-leveltypealias | top-levelvar | top-levelfunc}
```
hugomg, line 385: I'd prefer if the names of the non-terminals matched the ones that we are using in parser.lua. Unfortunately, I don't think we can use hyphens there.

itssamuelrowe: I accidentally replaced toplevel with top-level in the grammar.

hugomg: Yeah... Let's keep using `toplevel` everywhere.

spec/checker_spec.lua
```diff
@@ -55,20 +55,21 @@ describe("Scope analysis: ", function()
...
-        ]],
-            "duplicate toplevel declaration for 'a'")
-    end)
+    -- Shouldn't this be removed?
+    -- it("forbids multiple toplevel variable declarations with the same name", function()
+    --     assert_error([[
+    --         local a, a = 1, 2
+    --     ]],
+    --         "duplicate toplevel declaration for 'a'")
+    -- end)
```
hugomg, line 72: Instead of removing it, I would move this test to coder_spec, the "Corner cases of scoping" section.

Please also add a test there to check what happens if you have two local functions with the same name.

spec/checker_spec.lua
```diff
@@ -55,20 +55,21 @@ describe("Scope analysis: ", function()
...
-    it("forbids multiple toplevel declarations with the same name", function()
+    it("forbids multiple toplevel declarations with the same name for exported functions", function()
         assert_error([[
-            local function f() end
-            local function f() end
+            export function f() end
+            export function f() end
         ]],
-            "duplicate toplevel declaration for 'f'")
+            "duplicate export 'f', previous one at line 1")
```
hugomg, line 63: Could you please add another test case for an exported function that has the same name as an exported variable? Just so that we don't forget to add that test case later. You can use `pending()` for now.

pallene/parser.lua
```diff
@@ -249,7 +249,8 @@ local grammar = re.compile([[
     toplevelrecord  <- (P  RECORD NAME^NameRecord recordfields
                            END^EndRecord)                        -> ToplevelRecord
 
-    localopt        <- (LOCAL)?                                  -> opt_bool
+    modifier        <- LOCAL -> to_true
+                     / EXPORT -> to_false
```
hugomg, line 253: What do you think about using a more specific name such as `export_modifier` instead of `modifier`?

I also wonder if the rest of the code would become clearer if we changed `is_local` to `is_exported`.

itssamuelrowe: >What do you think about using a more specific name such as export_modifier instead of modifier?
The modifier rule has both the `local` and `export` keywords. It is somewhat confusing to have them both under a rule called `export_modifier`.

> I also wonder if the rest of the code would become clearer if we changed `is_local` to `is_exported`.
As far as I can tell, it is clear. I think we could use the time and energy on something else. :P

hugomg: What about `export_or_local`? What I don't like about the "modifier" name is that it is too vague.

doc/manual.md
hugomg: By the way, please add the new export keyword to the "Complete Syntax of Pallene" section in the manual.

spec/checker_spec.lua
```diff
@@ -63,12 +63,13 @@ describe("Scope analysis: ", function()
...
     end)
 
-    -- Shouldn't this be removed?
-    -- it("forbids multiple toplevel variable declarations with the same name", function()
+    pending("forbids multiple toplevel declarations with the same name for exported function and variable")
+    -- it("forbids multiple toplevel declarations with the same name for exported function and variable", function()
     --     assert_error([[
-    --         local a, a = 1, 2
+    --         export function f() end
+    --         export variable = 1
```
hugomg, line 70: The variable and the function should have the same name in this pending test.

spec/coder_spec.lua
```diff
@@ -1719,6 +1719,15 @@ describe("Pallene coder /", function()
         end)
     end)
 
+
+    describe("Toplevel corner cases", function ()
```
hugomg, line 1723: I think you could put these tests in the existing "Corner cases of scoping" section.

spec/checker_spec.lua
```diff
@@ -55,19 +55,20 @@ describe("Scope analysis: ", function()
...
-            "duplicate toplevel declaration for 'f'")
+            "duplicate export 'f', previous one at line 1")
     end)
 
-    it("forbids multiple toplevel variable declarations with the same name", function()
+    pending("forbids multiple toplevel declarations with the same name for exported function and variable", function()
         assert_error([[
-            local a, a = 1, 2
+            export function f() end
+            export variable = 1
```
hugomg, line 69: This variable needs to have the same name as the function.

spec/coder_spec.lua
```diff
@@ -1608,44 +1609,54 @@ describe("Pallene coder /", function()
 
             typealias y = integer
 
+            local a, a = 1, 2
+
             ------
 
-            function local_type(): integer
+            export function local_a(): integer
```
hugomg, line 1616: In the other tests next to this we are using the name "local variable" to refer to local variables inside a function. It might be better to mention "toplevel variable" instead.

spec/coder_spec.lua
```diff
@@ -1608,44 +1609,54 @@ describe("Pallene coder /", function()
...
                 return (x + tofloat)
             end
 
-            function duplicate_parameter(x: integer, x:integer) : integer
+            export function duplicate_parameter(x: integer, x:integer) : integer
                 return x
             end
         ]]))
 
+        it("local variable is shadowed by the latest declaration", function ()
```
hugomg, line 1656: Also update this description to mention toplevel variables.

hugomg:
I think it is important to add test cases for those things that we discussed yesterday such as a local function shadowing an exported function or an exported function shadowing a local variable. For example, look at the code that you wrote in checker.lua for the "forbid toplevel duplicates" part and think of what test cases would be necessary to cover all of the tricky things.

These test cases will help a lot in the future. Remember those times when you ran into a tricky bug because it was something that wasn't previously covered by the test suite? :)

PR #253, opened 2020-06-26
Top-level declarations cannot have duplicates.

itssamuelrowe:
This pull-request fixes two issues:
1. In groups of function definitions that are defined next to each other, Pallene (unlike Lua) allows all the functions to see each other. We need to ensure that in each group of functions there are never two functions with the same name.
2. The second issue has to do with exported functions and variables. We need to ensure that those names are not shadowed at the end of the file, which is where we will insert the return statement.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -233,9 +233,9 @@ function Checker:check_program(prog_ast)
     do
         -- Forbid top-level duplicates
         --
-        -- To avoid ambiguities that could happen if the programmer tried to export multiple
-        -- functions with the same name, we give a compilation error if there is more than one
-        -- definition in the toplevel with the same name.
+        -- To avoid ambiguities that could happen if the programmer tried to declare
+        -- multiple toplevel entities with the same name, we give a compilation
+        -- error.
         local names = {}
         for _, top_level_node in ipairs(prog_ast) do
             local top_level_names = ast.toplevel_names(top_level_node)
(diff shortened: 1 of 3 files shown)


pallene/checker.lua
```diff
@@ -243,14 +243,9 @@ function Checker:check_program(prog_ast)
...
-                        top_level_node._tag == "ast.Toplevel.Var" then
-                        if not top_level_node.is_local then
-                            scope_error(node_location,
-                                "duplicate export '%s', previous one at line %d",
-                                name, old_location.line)
-                        end
-                    end
+                    scope_error(node_location,
+                        "duplicate '%s', previous one at line %d",
+                        name, old_location.line)
```
hugomg, line 248: Let's use the old error message: `duplicate toplevel declaration for '%s'`.

We should also update the comment again to mention that, in addition to avoiding confusion with `export`, this also avoids confusion with the mutually recursive functions.

spec/coder_spec.lua
```diff
@@ -1599,7 +1599,6 @@ describe("Pallene coder /", function()
         compile([[
             local z = 10
             export x = 2000
-            local x = 300
```
hugomg, line 1602: Let's continue testing the situation where there is a deleted variable after the one that we are testing. This test cared more about the `global_id` than the name of the variables... While we are at it, we could also slightly improve the test by testing both at the middle and at the end of the list of variables.

```
local  a = 10
export b = 20
local  c = 30
export d = 40
```

```
assert(nil == test.a)
assert(20 == test.b)
assert(nil == test.c)
assert(40 == test.d)
```

PR #254, opened 2020-06-29
Implement the Pallene to Lua translator

itssamuelrowe:
Pallene is a statically typed, ahead-of-time-compiled sister language to ​ Lua​ , with a focus on performance. Programmers can use Pallene instead of C modules and LuaJIT to improve and gain a predictable run-time performance. Pallene achieves this by translating a statically typed compilation unit into a C source file, which is implicitly compiled in the background. However, there are situations where removal of type annotations are useful.

 * The main premise of Pallene is that it is compatible with Lua. This compatibility is defined as removing type annotations from a 
 Pallene program results in transforming it into a Lua program. A Pallene to Lua translator will allow us to check whether this property is still valid.
 * Provides greater portability, interoperability and integration with existing Lua codebase and tools.
 * The Pallene developers could check if the unit tests are obeying the "gradual guarantee".
 * In the benchmarks that do not use LuaJIT features, we could generate the Lua version of the code from the Pallene version. With the help of the Pallene to Lua translator, users can remove Pallene type annotations to generate plain Lua. Pallene type annotations can be removed by extending the backend of the Pallene compiler to generate Lua.

Consider the following example written in Pallene.
```
function sum(values : { float }): float
    local s : float = 0.0
    for i = 1, #values do
         s = s + values[i]
    end
    return s
end
```

The Pallene to Lua translator should produce the following Lua program.
```
function sum(values)
    local s = 0.0
    for i = 1, #values do
        s = s + values[i]
    end
    return s
end
```


pallene/driver.lua
```diff
@@ -133,33 +133,45 @@ function driver.compile(argv0, input_ext, output_ext, input_file_name)
...
-        local step = compiler_steps[i]
-        if (i == first_step or i == last_step) then
-            file_names[i] = base_name .. "." .. step.name
-        else
-            file_names[i] = os.tmpname()
+    local ok, errs
+    if output_ext == "lua" then
+        print(input_file_name)
+        local input = assert(io.open(input_file_name, "r"))
+        local content = input:read("*all")
```
hugomg, line 140: We can use `util.get_file_contents` and `util.set_file_contents` here.

hugomg:
I'm unlinking issue 173. Let's leave that until the end of the project :)

spec/translator_spec.lua
```diff
@@ -0,0 +1,307 @@
...
+        ]],
+        [[
+            local function print_strings(strings)
+                for s in strings do
+                    io.write(s .. '\n')
+                end
+            end
+        ]])
+    end)
+end)
```
hugomg, line 307: Please add a newline here.

The weird thing is that our editorconfig is supposed to be telling the editor to add this newline. I don't know why sublime text didn't do it.

spec/translator_spec.lua
```diff
@@ -0,0 +1,307 @@
...
+            end
+        ]],
+        [[
+            function print(text: string)
+                io.write(text .. "\n")
+            end
+        ]])
+    end)
+
+    -- TODO: records, export/local shadows
```
hugomg, line 43: Please create an issue on github instead of the TODO comment.

spec/translator_spec.lua
itssamuelrowe: I was not able to setup the editorconfig.

hugomg: What was the problem that you encountered? I don't use sublime text myself but the editor config website says that there is a plugin for it.

https://github.com/sindresorhus/editorconfig-sublime#readme

itssamuelrowe: I installed the plugin and luacheck. But I don't get any warnings.

pallene/translator.lua
```diff
@@ -0,0 +1,36 @@
...
+        io.write(' = ')
+        for _, value in pairs(node.values) do
+            translate_expression(value)
+        end
+    end
+
+    io.write('\n')
+end
+
+function translator.translate(input, ast)
```
hugomg, line 25: In the rest of the program we have been using the name `prog_ast` instead of `ast`, to avoid clashing with the name of the `ast` module.

pallene/driver.lua
```diff
@@ -51,32 +52,32 @@ function driver.compile_internal(filename, stop_after)
     local prog_ast
     prog_ast, errs = parser.parse(filename, input)
     if stop_after == "ast" or not prog_ast then
-        return prog_ast, errs
+        return prog_ast, errs, input
```
hugomg, line 55: Having everything return an `input` as a third argument smells a bit fishy.

What do you think about splitting `compile_internal` into two functions, one that reads the source code (going up until the part that says "util.get_file_contents") and another one that calls the various compiler steps, starting with parser.parse?

pallene/driver.lua
```diff
@@ -133,33 +134,42 @@ function driver.compile(argv0, input_ext, output_ext, input_file_name)
...
-    for i = first_step, last_step do
-        local step = compiler_steps[i]
-        if (i == first_step or i == last_step) then
-            file_names[i] = base_name .. "." .. step.name
-        else
-            file_names[i] = os.tmpname()
+    local ok, errs
+    if output_ext == "lua" then
+        assert(input_ext == "pln")
+        local ast, errs, input = driver.compile_internal(input_file_name, "checker")
```
hugomg, line 140: This `errs` is shadowing the other errs so when you do a return `ok, errs` at the end of the function we won't have the actual error message.

This reminds me that we probably should add a test case to ensure that type errors are properly reported to the user if we call `--emit-lua` in a file that has type errors.

pallene/driver.lua
```diff
@@ -83,26 +81,27 @@ function driver.compile_internal(filename, stop_after)
...
-    local c_code
-    c_code, errs = coder.generate(module, mod_name)
-    if not c_code then
-        return c_code, errs
-    end
-
-    ok, err = util.set_file_contents(c_filename, c_code)
-    if not ok then
-        return ok, {err}
+    local ok, err, errs, input = false, nil, {}, nil
```
hugomg, line 84: I think the old version of this function was easier to read than the new one. 

1. Please use the early exit pattern instead of nested if statements. `if not blah then return blah, errs`. This way the program reads as a linear sequence instead of a big triangle.
2. Please declare the variables close to where they are used instead of all at the top of the function.
3. We don't need to initialize the variables in the foward declaration, since the value is going to be overwritten anyway.

pallene/driver.lua
```diff
@@ -29,27 +30,24 @@ local function check_source_filename(argv0, file_name, expected_ext)
...
 end
 
---
--- Run AST and IR passes, up-to and including the specified pass. This is meant for unit tests.
---
-function driver.compile_internal(filename, stop_after)
-    stop_after = stop_after or "optimize"
-    local err, errs
+function driver.load_input(filename)
+    local input, err = false, nil
```
hugomg, line 34: We might want to use the early-exit pattern here to be more consistent with the other functions. In this particular case we can write it in a way that doesn't need a forward declaration for input and err:

```
local base_name, err = check_source(...)
if not base_name then
    return base_name, err
end
return util.get_file_contents(...)
```

pallene/translator.lua
```diff
@@ -0,0 +1,81 @@
+local translator = {}
+
+local function translate_expression(expression)
+    if expression._tag == "ast.Exp.Integer" then
+        io.write(expression.value)
+    end
+end
+
+local function translate_tl_var(node)
+    io.write('local ')
```
hugomg, line 10: When we print the `"local "`, the `declaration.name` and the and `" = "` I think we won't be preserving the the whitespace that might have been present around those tokens in the original input. Would it be possible to implement these parts similarly to what we have been doing with the type annotations? Just find the location of the `local` or `export` token and replace just that substring.

pallene/translator.lua
```diff
@@ -0,0 +1,81 @@
...
+        io.write(' = ')
+        for _, value in pairs(node.values) do
+            translate_expression(value)
+        end
+    end
+
+    io.write('\n')
+end
+
+local last_index = 1
```
hugomg, line 25: This could be an intance variable, as we discussed.

pallene/translator.lua
```diff
@@ -0,0 +1,81 @@
...
+
+local function add_previous(input, partials, stop_index)
+    local partial = input:sub(last_index, stop_index)
+    table.insert(partials, partial)
+
+    -- Do not update the last index if this is the last call to add_previous.
```
hugomg, line 31: If the last call passes in the string's length instead of `nil` then I think we wouldn't need this special case.

pallene/translator.lua
```diff
@@ -0,0 +1,81 @@
...
+        if input:sub(q, q) == "\n" then
+            local partial = string.rep(" ", q - p - 1)
+            table.insert(partials, partial)
+            table.insert(partials, "\n")
+            p = q + 2
+        end
+        q = q + 1
+    end
+    local final_partial = string.rep(" ", q - p)
+    table.insert(partials, final_partial)
```
hugomg, line 54: What do you think about using `string.gsub` here? Replace any non-whitespace characters with a space `string.replace(str, "%S", " ").

By the way, this reminds me that we probably should also be careful about preserving tabs, for proper indentation, and `\r`, for people who are programming on Windows. Could you please add a test case that uses hard tabs?

pallene/translator.lua
```diff
@@ -0,0 +1,81 @@
...
+                local start = decl.type_start.pos
+                local stop = decl.type_end.pos
+                
+                add_whitespace(input, partials, start, stop - 1)
+            end
+        end
+    end
+    -- Whatever characters that were not included in the partials should be added.
+    -- local final_partial = input:sub(last_index)
+    -- table.insert(partials, final_partial)
```
hugomg, line 75: Please remove the commented-out code.

pallene/typedecl.lua
```diff
@@ -68,6 +68,7 @@ function typedecl.declare(module, mod_name, type_name, constructors)
         local function cons(...)
             local args = table.pack(...)
             if args.n ~= #fields then
+                print(require('inspect')(args))
```
hugomg, line 71: Please remove extra calls to inspect.

pallene/parser.lua
```diff
@@ -85,7 +85,7 @@ function defs.toplevel_func(loc, is_local, name, params, ret_types, block)
     local func_typ = ast.Type.Function(loc, arg_types, ret_types)
     return ast.Toplevel.Func(
         loc, is_local,
-        ast.Decl.Decl(loc, name, func_typ),
+        ast.Decl.Decl(loc, name, nil, func_typ, nil),
```
hugomg, line 88: It might be better to use `false` instead of `nil`. It works better with things like `inspect`.

pallene/translator.lua
itssamuelrowe: I actually have a list of things to test, tabs are included. But then I am waiting to implement the translator for the top-level variables completely before writing them.

pallene/translator.lua
itssamuelrowe: Oh okay.

pallene/ast.lua
```diff
@@ -33,7 +33,7 @@ declare_type("Toplevel", {
 })
 
 declare_type("Decl", {
-    Decl = {"loc", "name", "type"},
+    Decl = {"loc", "name", "type_start", "type", "type_end"},
```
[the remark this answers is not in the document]
itssamuelrowe: `col_loc` is a palindrome though XD

pallene/translator.lua
```diff
@@ -0,0 +1,62 @@
...
+    return self
+end
+
+function Translator:add_previous(input, stop_index)
+    local partial = input:sub(self.last_index, stop_index)
+    table.insert(self.partials, partial)
+    self.last_index = stop_index + 1
+end
+
+-- TODO: Added test case for newlines, space, comments, and tabs.
```
hugomg, line 19: Please create a github issue instead of a comment.

pallene/translator.lua
```diff
@@ -0,0 +1,62 @@
...
+        q = q + 1
+    end
+    local final_partial = string.rep(" ", q - p)
+    table.insert(self.partials, final_partial)
+
+    self.last_index = stop_index + 1
+end
+
+function translator.translate(input, prog_ast)        
+    -- print(require('inspect')(prog_ast))
```
hugomg, line 42: You forgot this comment.

pallene/translator.lua
```diff
@@ -0,0 +1,62 @@
...
+
+    local instance = Translator:new()
+
+    for _, node in ipairs(prog_ast) do
+        if node._tag == "ast.Toplevel.Var" then
+            for _, decl in ipairs(node.decls) do
+                local start = decl.col_loc.pos
+                local stop = decl.end_loc.pos
+                
+                instance:add_whitespace(input, start, stop - 1)
```
hugomg, line 52: We might want to call add_whitespace twice, once to remove the colon and once to remove the type. This way, if the programmer is a lunatic that left a comment between the colon and the type then the comment will be preserved.

pallene/translator.lua
```diff
@@ -0,0 +1,62 @@
...
+        if input:sub(q, q) == "\n" then
+            local partial = string.rep(" ", q - p - 1)
+            table.insert(self.partials, partial)
+            table.insert(self.partials, "\n")
+            p = q + 2
+        end
+        q = q + 1
+    end
+    local final_partial = string.rep(" ", q - p)
+    table.insert(self.partials, final_partial)
```
hugomg, line 36: I still think we should be able to replace this whole block of code by a call to `string.sub` followed by a call to `string.gsub`.

pallenec
```diff
@@ -49,8 +55,14 @@ local function compile_internal(filename, stop_after)
 end
 
 local function pretty_print_ir(filename)
-    local module, errs = driver.compile_internal(filename, "ir")
-    if not module then util.abort(table.concat(errs, "\n")) end
+    local input, err = driver.load_input(filename)
+    if err then
```
hugomg, line 59: To be consistent with the rest of the code, please use `if not input`.

spec/parser_spec.lua
```diff
@@ -33,9 +33,9 @@ end
 -- Assertions for full programs
 --
 
-local function parse(program_str)
-    assert(util.set_file_contents("__test__.pln", program_str))
-    return driver.compile_internal("__test__.pln", "ast")
+local function parse(code)
+    assert(util.set_file_contents("__test__.pln", code))
+    return driver.compile_internal("__test__.pln", code, "ast")
```
hugomg, line 38: Can we get away with removing the call to `set_file_contents`?

spec/translator_spec.lua
```diff
@@ -0,0 +1,341 @@
...
+    local ok, _, _, error_message = util.outputs_of_execute("./pallenec __test__.pln --emit-lua")
+    if not ok then
+        error(error_message)
+    end
+end
+
+local function assert_translation(pallene_code, expected)
+    compile(pallene_code)
+    local contents = util.get_file_contents("__test__.lua")
+    assert.are.same(contents, expected)
```
hugomg, line 14: IIRC, the expected "should" come before the "contents". This makes a difference in the error message that the asser generates. Please check to see if they are the right way around.

spec/translator_spec.lua
```diff
@@ -0,0 +1,341 @@
...
+    it("Remove type annotations from top-level variables", function ()
+        assert_translation([[
+            local xs: integer = 10
+        ]],
+        [[
+            local xs          = 10
+        ]])
+    end)
+
+    it("Keep newlines that appear inside type annotations", function ()
```
hugomg, line 57: The newlines in this particular test case are between the colon and the type. Could you add another test case where the newlines are inside the type itself? You could use an array, table or function type to do this.

pallene/translator.lua
```diff
@@ -0,0 +1,62 @@
...
+
+    self.last_index = stop_index + 1
+end
+
+function translator.translate(input, prog_ast)        
+    print(require('inspect')(prog_ast))
```
hugomg, line 42: We needed to delete this, not uncomment it 😅

spec/parser_spec.lua
itssamuelrowe: I thought we were doing in this in a different PR. :/

pallene/translator.lua
itssamuelrowe: Don't `sub` and `gsub` and create new strings every time we call them? Won't that be inefficient? I adopted this technique based on what is suggested in the Lua book.

pallene/driver.lua
```diff
@@ -83,26 +78,29 @@ function driver.compile_internal(filename, stop_after)
...
     if not c_code then
-        return c_code, errs
+        return false, errs
     end
 
+    local ok
     ok, err = util.set_file_contents(c_filename, c_code)
     if not ok then
-        return ok, {err}
+        errs = { err }
```
hugomg, line 100: This should be `return false, {err}`

pallene/driver.lua
```diff
@@ -83,26 +78,29 @@ function driver.compile_internal(filename, stop_after)
...
     end
 
-    return true, {}
+    return ok, errs
```
hugomg, line 103: If we are using `return false, ...`  in the other ones, we should use `return true,  {}` here.

pallene/driver.lua
```diff
@@ -133,33 +131,49 @@ function driver.compile(argv0, input_ext, output_ext, input_file_name)
...
+        assert(input_ext == "pln")
 
-    local file_names = {}
-    for i = first_step, last_step do
-        local step = compiler_steps[i]
-        if (i == first_step or i == last_step) then
-            file_names[i] = base_name .. "." .. step.name
+        local input, err = driver.load_input(input_file_name)
+        if err then
+            errs = { err }
```
hugomg, line 140: I think the basic ideas we used in compile_to_c also apply to this function. Try declaring variables like `ok` and `errs` closer to where we need them and see if you can use `return false, errs` or `return false, {err}` instead of nested `else` statements.

pallene/translator.lua
hugomg: String.sub creates new strings but I think it should be OK because this is only for the type annotations, which are relatively small.

This is much less string creation and concatenation than we have in the coder.lua, which is a big collection of table.concat and util.render (which calls gsub behind the scenes). I suspect that we might even be doing some O(N^2) stuff over there. 😅

spec/parser_spec.lua
hugomg: Sure. We can do it in another PR. There is already an issue for it so we won't forget it.

pallene/driver.lua
```diff
@@ -29,27 +30,21 @@ local function check_source_filename(argv0, file_name, expected_ext)
     return name
 end
 
+function driver.load_input(path)
+    local base_name, err = check_source_filename("pallenec test", path, "pln")
+    if not base_name then
+        return base_name, err
```
hugomg, line 36: We should probably choose between `if not foo then return foo` or `if not foo then return false` and use the same pattern everywhere in the file.

pallene/translator.lua
hugomg: On second thought, let's not bother with this for now. We can come back to this later if it is a problem in practice.

spec/translator_spec.lua
```diff
@@ -0,0 +1,394 @@
...
+            local xs 
+                       
+              = 10
+        ]])
+    end)
+
+    pending("Keep tabs that appear in a top-level variable type annotation", function ()
+        assert_translation(
+            "    local xs:\t\n" ..
+            "    \t    integer = 10\n",
```
hugomg, line 93: I think it is a bit hard to read where one string ends and another starts. What about

```
assert_translation(
    "aaaa"..
    "aaaa"
    ,
    "bbbb"..
    "bbbb")

or 

assert_translation(
    "aaaa"..
    "aaaa",

    "bbbb"..
    "bbbb")

?

pallene/driver.lua
itssamuelrowe: Okay.

But then if `not ok` evaluates to true, doesn't that mean  `ok` is `false`, or at least a falsy value?

pallene/driver.lua
itssamuelrowe: Alright.

pallene/driver.lua
itssamuelrowe: I can work on that. The problem is, I am not sure if these are falsy values or an actual `false`.

PR #257, opened 2020-07-02
Updated typedecl.declare to include expected and received argument co…

itssamuelrowe:
…unts in the error message.

Debugging things in the parser is a bit tricky. One thing that might help is editing the error
message in typedecl.lua to say what was the expected number and how many arguments were received.

### DIFF
--- a/pallene/typedecl.lua
+++ b/pallene/typedecl.lua
@@ -17,7 +17,7 @@
 -- })
 -- ```
 -- and what it does is create three functions, `ast.Var.Name`, `ast.Var.Bracket`, and `ast.Var.Dot`.
--- 
+--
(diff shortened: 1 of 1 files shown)


pallene/typedecl.lua
```diff
@@ -68,7 +68,8 @@ function typedecl.declare(module, mod_name, type_name, constructors)
         local function cons(...)
             local args = table.pack(...)
             if args.n ~= #fields then
-                error("wrong number of arguments for " .. cons_name)
+                error("wrong number of arguments for " .. cons_name
+                    .. ". Expected " .. #fields .. " but received " .. args.n .. ".")
```
hugomg, line 72: What do you think about using `string.format` here?

PR #268, opened 2020-07-13
Updated the test specs to avoid creating "__test__.pln".

itssamuelrowe:
Since `compile_internal` now accepts the Pallene code, the utility functions in the spec files were
updated to take advantage of this. In other words, a temporary file does not have to be created for
each test thus reducing the overall time taken for running the tests.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -2,7 +2,7 @@ local driver = require 'pallene.driver'
 local util = require 'pallene.util'
 
 local function run_checker(code)
-    assert(util.set_file_contents("__test__.pln", code))
+    -- "__test__.pln" does not exist on disk. The name is only used for error messages.
     local module, errs = driver.compile_internal("__test__.pln", code, "checker")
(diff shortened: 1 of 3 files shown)


spec/checker_spec.lua
```diff
@@ -2,7 +2,8 @@ local driver = require 'pallene.driver'
 local util = require 'pallene.util'
 
 local function run_checker(code)
-    assert(util.set_file_contents("__test__.pln", code))
+    -- "__test__.pln" does not exist on disk. The compiler expects
+    -- a file with the ".pln" extension.
```
hugomg, line 6: This isn't actually true... The name is used in error messages but Pallene does not complain if we use a name that does not end in pln. What about "\_\_test\_\_.pln does not actually exist. The name is only used for error messages" ?

By the way, that is a weird place to put a line break. Usually we break after a sentence or at the 100th column.

itssamuelrowe: I am sure why I added a line break there. Will update.

Issue #261, opened 2020-07-09
Add test cases for inputs with errors when the --emit-lua flag is enabled.


itssamuelrowe:
The translator spec should be updated to ensure that input with syntax and semantic errors are properly reported when --emit-lua is specified.

PR #269, opened 2020-07-14
Added test cases for inputs with errors during translation

itssamuelrowe:
The translator spec was updated to ensure that input with syntax and semantic errors are properly reported when `--emit-lua` is specified. It should be noted that the test spec does not actually invoke the compiler with the flag, instead it invokes an internal
function to simulate it.

### DIFF
--- a/pallene/driver.lua
+++ b/pallene/driver.lua
@@ -125,6 +125,18 @@ end
 --    compile("c", "so", "foo.c)      --> outputs "foo.so"
 --
 
+function driver.translate(input_file_name, input)
+    local prog_ast, errs = driver.compile_internal(input_file_name, input, "checker")
+
+    if not prog_ast then
+        return false, errs
+    end
+
+    local translation = translator.translate(input, prog_ast)
+
+    return translation, {}
+end
+
 local function compile_pln_to_lua(input_ext, output_ext, input_file_name, base_name)
     assert(input_ext == "pln")
 
@@ -133,14 +145,12 @@ local function compile_pln_to_lua(input_ext, output_ext, input_file_name, base_n
         return false, { err }
     end
 
-    local prog_ast, errs = driver.compile_internal(input_file_name, input, "checker")
-    if not prog_ast then
+    local translation, errs = driver.translate(input_file_name, input)
+    if not translation then
         return false, errs
     end
 
-    local translation = translator.translate(input, prog_ast)
     assert(util.set_file_contents(base_name .. "." .. output_ext, translation))
-
     return true, {}
 end
 
--- a/pallene/driver.lua
+++ b/pallene/driver.lua
@@ -134,16 +134,11 @@ local function compile_pln_to_lua(input_ext, output_ext, input_file_name, base_n
     end
 
     local prog_ast, errs = driver.compile_internal(input_file_name, input, "checker")
-
     if not prog_ast then
         return false, errs
     end
 
     local translation = translator.translate(input, prog_ast)
-
-    if not translation then
-        return false, errs
-    end
 
     assert(util.set_file_contents(base_name .. "." .. output_ext, translation))
     return true, {}
--- a/pallene/driver.lua
+++ b/pallene/driver.lua
@@ -125,18 +125,6 @@ end
 --    compile("c", "so", "foo.c)      --> outputs "foo.so"
 --
 
-function driver.translate(input_file_name, input)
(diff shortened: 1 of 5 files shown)


pallene/parser.lua
```diff
@@ -249,8 +249,8 @@ local grammar = re.compile([[
     toplevelrecord  <- (P  RECORD NAME^NameRecord recordfields
                            END^EndRecord)                        -> ToplevelRecord
 
-    export_or_local <- LOCAL -> to_true
-                     / EXPORT -> to_false
+    export_or_local <- (LOCAL -> to_true
+                     / EXPORT -> to_false)
```
hugomg, line 253: Why are we adding `()` here?

pallene/driver.lua
```diff
@@ -125,6 +126,18 @@ end
 --    compile("c", "so", "foo.c)      --> outputs "foo.so"
 --
 
+function driver.translate(input_file_name, input)
```
hugomg, line 129: I think it is a bit strange that the test suite is using different code paths for the cases that give errors and the cases that don't. In the former we invoke pallenec and in the latter we call a special function from driver.lua.

What do you think about calling pallenec in both cases?

```
local function compile(pallene_code)
    assert(util.set_file_contents("__test__.pln", pallene_code))
    local ok, _, _, error_message = util.outputs_of_execute("./pallenec __test__.pln --emit-lua")
    return ok, error_message
end

local function assert_translation(pallene_code, expected)
    assert(compile(pallene_code))
    local contents = util.get_file_contents("__test__.lua")
    assert.are.same(expected, contents)
end

local function assert_translation_error(pallene_code, expected)
    local ok, err = compile("__test__.pln", pallene_code)
    assert.is.falsy(ok)
    assert.match(expected, err, 1, true)
end
```

pallene/driver.lua
```diff
@@ -50,6 +50,7 @@ function driver.compile_internal(filename, input, stop_after)
     end
 
     prog_ast, errs = checker.check(prog_ast)
+
```
hugomg, line 53: Do we need this empty line?

pallene/syntax_errors.lua
```diff
@@ -175,6 +175,8 @@ local errors = {
     AssignNotToVar = "Expected a valid lvalue in the left side of assignment but found a regular expression.",
 
     CastMissingType = "Expected a type for the cast expression.",
+
+    LocalOrExportRequired = "Function definition must include 'local' or 'export' modifier."
```
hugomg, line 179: Is this related to #261 or is it something else?

itssamuelrowe: No, it's not related to #261. I thought it would be nice to have a custom message is all. I've removed it.

pallene/driver.lua
```diff
@@ -133,14 +145,12 @@ local function compile_pln_to_lua(input_ext, output_ext, input_file_name, base_n
         return false, { err }
     end
 
-    local prog_ast, errs = driver.compile_internal(input_file_name, input, "checker")
-    if not prog_ast then
+    local translation, errs = driver.translate(input_file_name, input)
+    if not translation then
```
hugomg, line 149: I think we can keep all the logic inside compile_pln_to_lua like it was before. No need to split part of it into an `driver.translate` exported function.

spec/translator_spec.lua
```diff
@@ -1,4 +1,5 @@
 local util = require "pallene.util"
+local driver = require "pallene.driver"
```
hugomg, line 2: The linter is compaining about this unused variable.

spec/translator_spec.lua
```diff
@@ -14,11 +15,58 @@ local function assert_translation(pallene_code, expected)
...
+    assert.is_false(ok)
+    assert.match(expected, actual, 1, true)
+end
+
 local function cleanup()
     os.remove("__test__.pln")
     os.remove("__test__.lua")
 end
 
+describe("Syntax errors during Pallene to Lua translation", function ()
```
hugomg, line 30: We could replace each of these describe blocks with a single test, as discussed on Slack.

pallene/driver.lua
```diff
@@ -134,13 +134,14 @@ local function compile_pln_to_lua(input_ext, output_ext, input_file_name, base_n
     end
 
     local prog_ast, errs = driver.compile_internal(input_file_name, input, "checker")
+
```
hugomg: Can we remove this line?

PR #271, opened 2020-07-17
Added documentation for the Pallene to Lua translator.

itssamuelrowe:


pallene/translator.lua
```diff
@@ -5,6 +5,24 @@
...
+--
+-- Both these ends are decoupled, this provides us with the flexibility to integrate another backend
+-- that generates Lua. The users can run the compiler with `--emit-lua` trigger the translator to
+-- generate plain Lua instead of C.
+--
+-- The generation of Lua is performed by a different backend (implemented here). It accepts input
+-- string and the AST generated by the parser. The generator then walks over the AST to replacing
+-- type annotations with white space. Interestingly spaces, newlines, comments and pretty much
+-- everything else other than type annotations are retained in the translated code. Thus, the
+-- formatting in the original input is preserved.
```
hugomg, line 23: Good explanation. Could you please also mention that the line numbers are preserved, so the error messages point to the right place?

doc/manual.md
```diff
@@ -410,6 +410,49 @@ return m
 
 But we do not use this pattern in Pallene because it is harder to optimize function calls if they can be monkey-patched.
 
+## Pallene to Lua translator
```
hugomg, line 413: In this part we don't need to repeat the explanation of what Pallene is. We also don't need to explain how it is implemented -- the manual is primarily intended for end users.

PR #275, opened 2020-07-21
Update the translator to translate the remaining constructs

itssamuelrowe:
The following tasks were achieved in this PR:
 * Update the test cases to enable or add new tests.
 * Refactor the translator to remove redundant code.
 * Remove type annotation from for parameter.
 * Generate a return statement at the end to export symbols.
 * Convert exported symbols to local symbols to prevent globals symbols from being created.

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -245,7 +245,7 @@ local grammar = re.compile([[
                            LPAREN^LParPList paramlist RPAREN^RParPList
                            P rettypeopt P block END^EndFunc)         -> toplevel_func
 
-    toplevelvar     <- (P  export_or_local decllist ASSIGN^AssignVar
+    toplevelvar     <- (P  export_or_local P decllist ASSIGN^AssignVar
                            !IMPORT explist1^ExpVarDec)           -> ToplevelVar
 
     typealias       <- (P  TYPEALIAS NAME^NameTypeAlias ASSIGN^AssignTypeAlias
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -77,7 +77,7 @@ function defs.opt_list(x)
     end
 end
 
-function defs.toplevel_func(loc, is_local, mod_end_loc, name, params, rt_col_loc, ret_types,
+function defs.toplevel_func(loc, is_local, name, params, rt_col_loc, ret_types,
     rt_end_loc, block)
     local arg_types = {}
     for i, decl in ipairs(params) do
@@ -85,7 +85,7 @@ function defs.toplevel_func(loc, is_local, mod_end_loc, name, params, rt_col_loc
     end
     local func_typ = ast.Type.Function(loc, arg_types, ret_types)
     return ast.Toplevel.Func(
-        loc, is_local, mod_end_loc,
+        loc, is_local,
         ast.Decl.Decl(loc, name, false, func_typ, false),
         ast.Exp.Lambda(loc, params, block), rt_col_loc, rt_end_loc)
 end
@@ -241,11 +241,11 @@ local grammar = re.compile([[
                            / NAME (ASSIGN / COMMA) %{LocalOrExportRequiredVariable} )*
                         |} !.
 
-    toplevelfunc    <- (P  export_or_local P FUNCTION NAME^NameFunc
+    toplevelfunc    <- (P  export_or_local FUNCTION NAME^NameFunc
                            LPAREN^LParPList paramlist RPAREN^RParPList
                            P rettypeopt P block END^EndFunc)         -> toplevel_func
 
-    toplevelvar     <- (P  export_or_local P decllist ASSIGN^AssignVar
+    toplevelvar     <- (P  export_or_local decllist ASSIGN^AssignVar
                            !IMPORT explist1^ExpVarDec)           -> ToplevelVar
 
     typealias       <- (P  TYPEALIAS NAME^NameTypeAlias ASSIGN^AssignTypeAlias
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -77,14 +77,15 @@ function defs.opt_list(x)
     end
 end
 
-function defs.toplevel_func(loc, is_local, name, params, rt_col_loc, ret_types, rt_end_loc, block)
+function defs.toplevel_func(loc, is_local, mod_end_loc, name, params, rt_col_loc, ret_types,
+    rt_end_loc, block)
     local arg_types = {}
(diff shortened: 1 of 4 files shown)


pallene/parser.lua
```diff
@@ -240,11 +241,11 @@ local grammar = re.compile([[
                            / NAME (ASSIGN / COMMA) %{LocalOrExportRequiredVariable} )*
                         |} !.
 
-    toplevelfunc    <- (P  export_or_local FUNCTION NAME^NameFunc
+    toplevelfunc    <- (P  export_or_local P FUNCTION NAME^NameFunc
```
hugomg, line 244: When we removed the `:` in the decls we didn't need to add a P after the `COLON`. Would it be possible to do the same here?

pallene/translator.lua
```diff
@@ -31,77 +31,124 @@ local Translator = util.Class()
 function Translator:init()
     self.last_index = 1
     self.partials = {}
+    self.exports = {}
+    self.input = false
```
hugomg, line 35: Please put the declaration of `self.input` next to the declaration of `last_index`. It would also be a good idea to add comments next to each of these instance variables, describing their types. (The coder.lua should have some examples of that).

pallene/translator.lua
```diff
@@ -31,77 +31,124 @@ local Translator = util.Class()
...
+    self:add_whitespace(node.rt_col_loc.pos, node.rt_end_loc.pos - 1)
+
+    self:translate_block(node.value.body)
+end
+
 function translator.translate(input, prog_ast)
     local instance = Translator:new()
-
-    print(require("inspect")(prog_ast))
+    instance.input = input
```
hugomg, line 134: Two things here:

1. We should be using `Translator.new()` instead of `Translator:new()`. Our util.class expects the former.
2. We could pass the input as a parameter to the constructor instead of initializing it separately.

Fixing problem number 1, which we had overlooked before, is a prerequisite for fixing problem number 2.

pallene/translator.lua
```diff
@@ -31,77 +31,124 @@ local Translator = util.Class()
...
+        table.insert(self.partials, "\nreturn {\n")
+        for _, export in ipairs(self.exports) do
+            local pair = string.format("    %s = %s,\n", export, export)
+            table.insert(self.partials, pair)
+        end
+        table.insert(self.partials, "}\n")
+    end
+end
+
+function Translator:translate_var_decl(decl)
```
hugomg, line 76: To keep the same pattern as the `checker.lua`, what do you think about naming these functions after the type that they are translating. That is, `translate_decl`, `translate_toplevel`, `translate_stat`, etc.

pallene/translator.lua
```diff
@@ -31,77 +31,124 @@ local Translator = util.Class()
...
+    if decl.type then
+        -- Remove the colon but retain any adjacent comment to the right.
+        self:add_whitespace(decl.col_loc.pos, decl.col_loc.pos)
+        -- Remove the type annotation but exclude the next token.
+        self:add_whitespace(decl.type.loc.pos, decl.end_loc.pos - 1)
+    end
+end
+
+function Translator:translate_tl_variables(node)
```
hugomg, line 85: We could have a single `translate_toplevel` function instead of one function for each kind of toplevel thing.

pallene/translator.lua
```diff
@@ -31,77 +31,124 @@ local Translator = util.Class()
...
+            table.insert(self.exports, decl.name)
+        end
+    end
+
+    for _, decl in ipairs(node.decls) do
+        self:translate_var_decl(decl)
+    end
+end
+
+function Translator:translate_block(block)
```
hugomg, line 100: Similarly to the toplevel functions, we should have a translate_stat function instead of a translate_block function that only works on `ast.Stat.Block`.

pallene/parser.lua
itssamuelrowe: The `P` after `:` in `decl` points exactly to the `:` token. With the modifier we need to capture the range for `KEYWORD WHITESPACE*`, if I am not wrong.

pallene/translator.lua
```diff
@@ -28,9 +28,11 @@ local translator = {}
...
 local Translator = util.Class()
 
-function Translator:init()
-    self.last_index = 1
-    self.partials = {}
+function Translator:init(input)
+    self.input = input -- string
+    self.last_index = 1 -- integer
+    self.partials = {} -- string
+    self.exports = {} -- string
```
hugomg, line 35: Both partials and exports are lists of strings, not a string.

pallene/translator.lua
```diff
@@ -46,65 +48,105 @@ function Translator:add_whitespace(input, start_index, stop_index)
...
+function Translator:translate_decl(decl)
+    if decl.type then
+        -- Remove the colon but retain any adjacent comment to the right.
+        self:add_whitespace(decl.col_loc.pos, decl.col_loc.pos)
+        -- Remove the type annotation but exclude the next token.
+        self:add_whitespace(decl.type.loc.pos, decl.end_loc.pos - 1)
+    end
+end
+
+function Translator:translate_stat(block)
```
hugomg, line 88: Could you make this function similar to the other ones?

There is a big if-else-if for the possible statement types. If it is an ast.Stat.Block then it calls itself recursively.

pallene/translator.lua
```diff
@@ -46,65 +48,105 @@ function Translator:add_whitespace(input, start_index, stop_index)
...
-                    for _, decl in ipairs(statement.decls) do
-                        if decl.type then
-                            -- Remove the colon but retain any adjacent comment to the right.
-                            instance:add_whitespace(input, decl.col_loc.pos, decl.col_loc.pos)
-                            -- Remove the type annotation but exclude the next token.
-                            instance:add_whitespace(input, decl.type.loc.pos, decl.end_loc.pos - 1)
-                        end
-                    end
-                end
+function Translator:translate_tl(node)
```
hugomg, line 102: To be consistent with the other files, please call this function `translate_toplevel`.

pallene/translator.lua
```diff
@@ -46,65 +48,105 @@ function Translator:add_whitespace(input, start_index, stop_index)
...
+
+    for _, node in ipairs(prog_ast) do
+        instance:translate_tl(node)
     end
     -- Whatever characters that were not included in the partials should be added.
-    instance:add_previous(input, #input)
+    instance:add_previous(#input)
+    instance:add_exports()
 
     return table.concat(instance.partials, "")
```
hugomg, line 151: We don't need the `, "")` here.

PR #278, opened 2020-07-21
Assert that last_index is increasing.

hugomg:
The add_previous and add_whitespaces assume that they are called on increasing indices. This should
be checked by an assertion.

### DIFF
--- a/pallene/translator.lua
+++ b/pallene/translator.lua
@@ -35,12 +35,15 @@ function Translator:init()
 end
 
 function Translator:add_previous(input, stop_index)
+    assert(self.last_index <= stop_index + 1)
     local partial = input:sub(self.last_index, stop_index)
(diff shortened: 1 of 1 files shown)


hugomg:
It is possible that there could be a ±1 error here so I'd appreciate if you could double check the math.

pallene/translator.lua
```diff
@@ -35,12 +35,15 @@ function Translator:init()
 end
 
 function Translator:add_previous(input, stop_index)
+    assert(self.last_index <= stop_index + 1)
```
itssamuelrowe, line 38: Why are we adding 1 here?

itssamuelrowe:
> It is possible that there could be a ±1 error here so I'd appreciate if you could double check the math.

Seems correct :sweat_smile:

Issue #280, opened 2020-07-22
Reduce the amount of space used by locations when we dump an AST


hugomg:
Each location node adds 6 lines to the output of --dump=ast or --dump=checker. 

```
        col_loc = {
          col = 10,
          file_name = "foo.pln",
          line = 1,
          pos = 10
        },
        end_loc = {
          col = 10,
          file_name = "foo.pln",
          line = 1,
          pos = 10
        },
```

We should use a location-specific printinf function for all the fields that end with "loc". For example, something like 

     col_loc = L:1:10,
     end_loc = L:1:10,

hugomg:
Fixed by #281.

Github is a bit obtuse. It doesn't understand `Fixes issue #NNN`. It has to be `Fixes #NNN`.

PR #281, opened 2020-07-23
Reduce the amount of space used by locations when dumping AST.

itssamuelrowe:
Fixes issue #280.

Each location node adds 6 lines to the output of `--dump=ast` or `--dump=checker`.
To avaoid this, we format the location when dumping the AST using the `string.format`
function. To be precise, we try to format for all the fields that end with "loc".
For example, something like:

```
    col_loc = {
      col = 10,
      file_name = "foo.pln",
      line = 1,
      pos = 10
    },
    end_loc = {
      col = 10,
      file_name = "foo.pln",
      line = 1,
      pos = 10
    },
```

becomes:

```
	col_loc = "L:1 C:10 P:10"
	end_loc = "L:1 C:10 P:10"
```


pallenec
```diff
@@ -35,6 +35,11 @@ local function compile(in_ext, out_ext)
 end
 
 local function process_item(item, path)
+    print(inspect(path))
```
hugomg, line 38: 👀

pallenec
```diff
@@ -35,6 +35,11 @@ local function compile(in_ext, out_ext)
...
+    if type(path[#path]) == 'string' and string.match(".*loc$", path[#path]) then
```
hugomg, line 39: The parameters to string.match are switched. This is not recognizing end_loc.

PR #289, opened 2020-07-25
Compiler optimization flag

### LINKED ISSUE #248: Add compiler flags to enable or disable optimizations
In PR #245 we found a bug in the constant_propagation optimization, where that optimization was doing the wrong thing. I expect that we might find more problems like this in the future, as we add new optimizations.

One thing that might help is to add compiler flags to enable or disable optimizations. 

Another thing that might help would be to teach the test suite to run programs both with and without optimizations.

mambrozio:
Fixes #248 

This PR introduces the flag `-O` in pallenec. This flag accepts the same values as a C compiler accepts for the `-O` flag. For example:

```bash
#execute no optimization (Pallene and C compiler)
$ ./pallenec test.pln -O0

#execute Pallene compiler optimizations and C compiler level 3 optimizations
$ ./pallenec test.pln -O3

#execute no optimizations for Pallene compiler but executes C compiler level 2 optimizations
$ env CFLAGS="-O2" ./pallenec test.pln -O0

#execute all default optimizations (same as -O2)
$ ./pallenec test.pln
```

It also changes how pallenec handles mutually exclusive options. Now `argparse` treats this, instead of manually doing it. This was done by moving such options to a `mutex` which is `argparse`'s way of handling mutually exclusive options. The test that covers this was also updated.


pallenec
```diff
@@ -70,21 +76,27 @@ local function pretty_print_ir(filename)
...
+    end
+else
+    for _, opt in ipairs(all_opt_passes) do
+        opt_passes[opt] = true
+    end
 end
 
-if     #flags == 0    then compile("pln", "so")
-elseif args.emit_c    then compile("pln", "c")
+if     flags == 0     then compile("pln", "so", opt_passes)
```
hugomg, line 98: If we put this in the `else`, would this let us get rid of the `flags` variable?

pallene/driver.lua
```diff
@@ -60,14 +60,18 @@ function driver.compile_internal(filename, input, stop_after)
...
     end
 
-    module, errs = uninitialized.verify_variables(module)
-    if stop_after == "uninitialized" or not module then
-        return module, errs
+    if stop_after ~= "optimize" or opt_passes["uninitialized"] then
+        module, errs = uninitialized.verify_variables(module)
+        if stop_after == "uninitialized" or not module then
+            return module, errs
+        end
```
hugomg, line 67: The "uninitialized" pass is not an optimization. It is an analysis pass to see if there are any variables that are used before they are initialized. Therefore, it should not be optional.

pallene/driver.lua
```diff
@@ -60,14 +60,18 @@ function driver.compile_internal(filename, input, stop_after)
...
     end
 
-    module, errs = constant_propagation.run(module)
-    if stop_after == "constant_propagation" or not module then
-        return module, errs
+    if stop_after ~= "optimize" or opt_passes["constant_propagation"] then
```
hugomg, line 70: Do we need this `stop_after ~= "optimize"`?

pallene/driver.lua
```diff
@@ -41,7 +41,7 @@ end
 --
 -- Run AST and IR passes, up-to and including the specified pass. This is meant for unit tests.
 --
-function driver.compile_internal(filename, input, stop_after)
+function driver.compile_internal(filename, input, stop_after, opt_passes)
```
hugomg, line 44: Now might be a good time to expand the documentation commend to describe what the `stop_after` and `opt_passes` parameters are. We at least want to mention their types.

spec/pallenec_spec.lua
```diff
@@ -46,6 +46,12 @@ describe("pallenec", function()
     it("Can detect conflicting arguments", function()
         local ok, err, _, abortMsg = util.outputs_of_execute("./pallenec --emit-c --emit-asm __test__.pln")
         assert.is_false(ok, err)
-        assert.equals("./pallenec: flags --emit-c and --emit-asm are mutually exclusive\n", abortMsg)
+        assert.equals([[Usage: pallenec
```
hugomg, line 49: This assert.equals will break whenever we change the set of flags. What do you think about doing a substring search instead?

    assert(string.find(err, "Error: option '--emit-asm' can not be used together with option '--emit-c'", nil, true))

pallenec
mambrozio: I thought about doing this, but I was afraid to lose the `error("impossible")`. If that is not a problem, I'll get rid of the `flag` variable.

hugomg: I think it is OK. I think we should have been using an `else` all along.

pallenec
```diff
@@ -7,28 +7,34 @@ local driver = require "pallene.driver"
...
 -- Inspired by gcc, eg. "gcc: fatal error: no input files".
 local compiler_name = arg[0]
 
-local function pallenec_abort(fmt, ...)
-    return util.abort(compiler_name .. ": " .. string.format(fmt, ...))
-end
-
-local function compile(in_ext, out_ext)
-    local ok, errs = driver.compile(compiler_name, in_ext, out_ext, args.source_file)
+local function compile(in_ext, out_ext, opt_passes)
```
hugomg, line 36: This function doesn't need to receive opt_passes as a parameter, because it is already in scope.

hugomg:
Actually, I think you are right about this fixing issue 248. I was confusing it with 252...

That said, I still think it makes sense to try to tackle both of those issues at once. Or at least make sure that whatever we do here will play nicely with the solution for 252.

pallene/driver.lua
```diff
@@ -65,9 +70,11 @@ function driver.compile_internal(filename, input, stop_after)
         return module, errs
     end
 
-    module, errs = constant_propagation.run(module)
-    if stop_after == "constant_propagation" or not module then
-        return module, errs
+    if opt_level ~= "0" then
```
hugomg, line 73: I think it would be nice if we could use `opt_level > 0` here. What do you think about removing -Os, -Ofast and -Og, allowing only -O0, -O1, -O2, and -O3? If the programmer really wants those other optimization flags they could still use CFLAGS.

mambrozio: I believe you are right. Since there is the option to override the C compiler optimization level with `CFLAGS`, -Os, fast, and g are really not needed. Already changed this. Now argparse ensures that the `-O` option only accepts numbers 0, 1, 2, and 3.

PR #290, opened 2020-07-27
Translate mutually recursive functions

itssamuelrowe:


pallene/translator.lua
```diff
@@ -190,6 +190,22 @@ function Translator:translate_stat(stat)
     end
 end
 
+function Translator:erase_mrf_modifier(start_index, is_local)
+    self:add_previous(start_index - 1)
+    local new_index
+    if is_local then
+        new_index = self.last_index + 5
```
hugomg, line 197: Could we use `string.find` to compute the end of the modifier? I'm not a fan of the hardcoded lengths.

PR #293, opened 2020-07-29
Add test cases to verify the resulting Lua program produced by the translator

itssamuelrowe:
The main premise of Pallene is that it is compatible with Lua. This compatibility is defined
as removing type annotations from a Pallene program results in transforming it into a Lua
program. The Pallene to Lua translator will allow us to check whether this property is still
valid. We need to automate this validation through unit tests that execute the resulting Lua
programs produced by the translator.

The utility functions introduced in this PR will help us design test cases to verify the
resulting Lua program produced by the translator to work as expected.


spec/execution_tests.lua
```diff
@@ -0,0 +1,2163 @@
...
+            local function assert_pallene_type(expected, x, ...)
+                if [===[$backend]===] == "c" then
+                    assert(expected == type(x))
+                end
+            end
+
+            ${TEST_SCRIPT}
+        ]], {
+            TEST_SCRIPT = test_script,
+            backend = backend
```
hugomg, line 32: If we use `string.format("%q", backend)` I think we would be able to get rid of the `[===[` quotes.

spec/execution_tests.lua
```diff
@@ -0,0 +1,2163 @@
...
+                    local ok, err = pcall(target, ...)
+                    assert(not ok)
+                    assert(string.find(err, message, nil, true))
+                end
+            end
+
+            local function assert_pallene_type(expected, x, ...)
+                if [===[$backend]===] == "c" then
+                    assert(expected == type(x))
+                end
```
hugomg, line 26: We could change this function into `assert_is_pallene_record`. On the C backend it tests if the thing is userdata and on the Lua backend it could test if the thing is  a  table.

spec/execution_tests.lua
```diff
@@ -0,0 +1,2163 @@
...
+
+            local function assert_pallene_error(message, target, ...)
+                if [===[$backend]===] == "c" then
+                    local ok, err = pcall(target, ...)
+                    assert(not ok)
+                    assert(string.find(err, message, nil, true))
+                end
+            end
+
+            local function assert_pallene_type(expected, x, ...)
```
hugomg, line 23: Do we need the `...`?

Issue #262, opened 2020-07-09
Should we replace the LPeg parser with a recursive descent parser?


itssamuelrowe:
This issue lists out all the pain points we have encountered so far with the parser. If we ever do rewrite the parser, knowing exactly what are the pain points that we want to solve would let us make more informed decisions later.

Some pain points currently include:
 * Contributors have to learn LPeg.
 * The parser and the lexer are hard to debug.
 * The lexer is embedded in the parser.
 * Some tokens such as comments are discarded. If we ever implement document generators, linters, and other such tools we could reuse existing code base if the lexer emitted hidden tokens on a different channel, in other words, a multi-channel lexer.  
 * Finding locations for tokens, sometimes, means we have to edit the parser.
 * Pallene exits after encountering a single syntax error. With a handwritten parser, we could implement error recovery strategies. However, we should also take into consideration that we are using an extended version of LPeg called lpeglabel, which has support for error recovery. But currently we are only generating errors and never recovering from them. We have not explored the question of whether it would be easier to implement error recovery using lpeglabel's error recovery or doing it by hand.

hugomg:
Implemented in #294

PR #294, opened 2020-07-29
Recursive descent parser

hugomg:
Inspired by #262, I've been exploring what it would mean to replace the current LPeg-based parser with a recursive-descent parser written in Lua. I still haven't started working on the actual parser but I have already finished rewriting the lexer. Now the main lexer exports a "next token" function instead of just being a collection of LPeg patterns. The new lexer architecture let's us simplify some things:

1. Since the lexer rules are explicitly ordered, we can get rid of most negative lookaheads. We don't need to worry about LPeg  backtracking in a way that violates the longest match rule.
2. Lua loops and if statements are easier to understand than fancy LPeg combinators, in my opinion.
3. The lexer is now able to compute line and column numbers without resorting to binary search.
4. The lexer can now be expressed using vanilla LPeg.

So far I think I like the new lexer more than the old one but the real question is going to be the parser. If the recursive descent parser isn't better than the old one then changing the lexer would not be worth it.


hugomg:
Highlights from the rewritten parser:

* Now we use the almost the same grammar as the original Lua parser from lparser.c
* The new parser no longer backtracks. The control flow is more explicit.
* We don't need to use global variables for THIS_FILENAME
* We don't need to use a separate pass to detect break statements outside a loop.
* Error messages are now standardized, using a format similar to what the Lua parser uses.
  We no longer have to come up with a new error message for each possible error.
* If there is an exception during parsing we now get a useful stack trace.

Along the way we have also made some small changes to the AST and the grammar

* The ast.Field is now split into ast.Field.Rec and ast.Field.List
* We got rid of the grammar rules for `import`
* The empty table type {} is now allowed.

The new version of the parser is a bit more verbose than the LPeg version but in the end it balanced out by reducing the size of things around the parser, like the parser.defs and the syntax_errors.lua. (Although to be fair, it would also be possible to change the error messages while maintaining a LPeg-based architecture).

The thing that feels most clunky at the moment are the various self.prev and self.next we use to compute the positions. I couldn't think of a better way but any suggestions would be appreciated.

Issue #277, opened 2020-07-21
In the translator, consider using gsub("%S", "") instead of gsub("%S", " ")


hugomg:
If we remove the non-whitespace characters instead of simply replacing them with spaces it might produce more pleasant-looking output. For example,

    function f(x: integer)

could become

    function (x)

instead of 

    function f(x         )

Column numbers would not be preserved but line numbers would still be preserved. I think this might be OK. IIRC, Lua error messages don't show column numbers, they only show line numbers.

hugomg:
Fixed by #303.

bjornbm:
> Column numbers would not be preserved but line numbers would still be preserved. I think this might be OK. IIRC, Lua error messages don't show column numbers, they only show line numbers

I was about to write an issue suggesting the original behavior, to get the column numbers. For “cheap” editor support, I am considering running `--emit-lua` and then, for example, run _luacheck_ on the lua file(s) and use this to drive the editor (for example, [neo]vim quickfix functionality). Maybe _lua-ls_ (sumneko) could be jury-rigged to be useful for Pallene in a similar fashion (roundtripping via `--emit-lua`).

Also, the fact that Lua itself doesn't show column numbers  is perhaps an implementation detail that could change in the future?

Anyway, in my opinion not much was won by removing the spaces, while potential for some tooling reuse was lost. If the more pleasant-looking output is important maybe keep `--emit-lua` as it is and add `--emit-lua-preserve-columns`?

hugomg:
Hmm, now that's an use case I something I hadn't thought about!

I suppose that a command flag to tell pallenec to only type-check the program, without emitting a C or Lua file, might be helpful. I cooked up a quick one over on PR #566. Could you please check if it is useful for your purposes?

As for the column numbers in --emit-lua, maybe it's useful if there are extra error messages you can get from luacheck but not from pallenec. Please feel free to experiment with it and create a PR if it turns out to be a good idea. (I do like the current pleasant-looking output though, so my vote is for a  separate flag or configuration option). I only ask that if you do so, open a new issue instead of commenting on this one. It's easier for me to find the discussion if it's under a recent and open issue.

PR #303, opened 2020-08-06
Updated the translator to use `gsub("%S", "")` instead of `gsub("%S",…

itssamuelrowe:
… " ")`

Fixes issue #277. If we remove the non-whitespace characters instead of simply replacing them with
spaces it produces a more pleasant-looking output. For example,

```
function f(x: integer)
```

becomes

```
function (x)
```

instead of

```
function f(x         )
```

Column numbers are not be preserved but line numbers are. This is okay because Lua error messages
do not show column numbers, they only show line numbers.

### DIFF
--- a/pallene/translator.lua
+++ b/pallene/translator.lua
@@ -50,12 +50,26 @@ end
 
 function Translator:add_whitespace(start_index, stop_index)
     assert(self.last_index <= start_index)
-    assert(start_index <= stop_index+1)
+    assert(start_index <= stop_index + 1)
     self:add_previous(start_index - 1)
 
     local region = self.input:sub(start_index, stop_index)
-    local partial = region:gsub("%S", " ")
-    table.insert(self.partials, partial)
+    local start_pos = region:find("[\n\r][ \t]*$")
+    if start_pos then
+        local first_region = region:sub(1, start_pos)
+        local first_partial = first_region:gsub("[^\n\r]", "")
+        local last_partial = region:sub(start_pos + 1)
+
+        table.insert(self.partials, first_partial)
+        table.insert(self.partials, last_partial)
+    else
+        local partial = region:gsub("[^\n\r]", "")
+        table.insert(self.partials, partial)
+    end
+
+    if self.input:sub(stop_index + 1, stop_index + 1) == "=" then
+        table.insert(self.partials, " ")
+    end
 
     self.last_index = stop_index + 1
 end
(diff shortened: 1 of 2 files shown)


pallene/translator.lua
```diff
@@ -50,12 +50,24 @@ end
...
 function Translator:add_whitespace(start_index, stop_index)
     assert(self.last_index <= start_index)
-    assert(start_index <= stop_index+1)
+    assert(start_index <= stop_index + 1)
     self:add_previous(start_index - 1)
 
     local region = self.input:sub(start_index, stop_index)
-    local partial = region:gsub("%S", " ")
-    table.insert(self.partials, partial)
+    local start_pos = region:find("[\n\r]%s*$")
```
hugomg, line 57: What do you think about using `[ \t]` instead of `%s`? I just noticed that this `%s` will also match newlines, which is not what we want.

pallene/translator.lua
```diff
@@ -50,12 +50,24 @@ end
...
+    if start_pos then
+        local first_region = region:sub(1, start_pos)
+        local first_partial = first_region:gsub("[^\n\r]", "")
+
+        local last_region = region:sub(start_pos + 1)
+        local last_partial = last_region:gsub("%S", "")
```
hugomg, line 63: Do we need to gsub the last region?

itssamuelrowe: I'll remove it.

pallene/translator.lua
```diff
@@ -83,10 +97,11 @@ end
 
 function Translator:translate_decl(decl)
     if decl.type then
+        print(require('inspect')(decl))
```
hugomg, line 100: Please review the PR diff to avoid leaving these debugging print statements behind..

pallene/translator.lua
```diff
@@ -83,10 +97,11 @@ end
...
         -- Remove the colon but retain any adjacent comment to the right.
-        self:add_whitespace(decl.col_loc.pos, decl.col_loc.pos)
+        self:add_whitespace(decl.col_loc.pos, decl.end_loc.pos - 1)
         -- Remove the type annotation but exclude the next token.
-        self:add_whitespace(decl.type.loc.pos, decl.end_loc.pos - 1)
+        -- self:add_whitespace(decl.type.loc.pos, decl.end_loc.pos - 1)
```
hugomg, line 104: These 3 comments are now outdated. We could just remove them.

PR #304, opened 2020-08-08
Updated the Pallene runtime to report error locations consistently.

itssamuelrowe:


pallene/coder.lua
```diff
@@ -967,8 +969,15 @@ gen_cmd["Binop"] = function(self, cmd, _func)
     -- For integer division and modulus:
     local function int_division(fname)
         local line = cmd.loc.line
-        return (util.render([[ $dst = $fname(L, $x, $y, $line); ]], {
-            fname = fname, dst = dst, x = x, y = y, line = C.integer(line) }))
+        print(require("inspect")(cmd))
```
hugomg, line 972: Please review the diff when you create the PR to avoid leaving behind these print statements.

pallene/coder.lua
```diff
@@ -196,10 +196,11 @@ end
 local function check_no_metatable(src, loc)
     return (util.render([[
         if ($src->metatable) {
-            pallene_runtime_array_metatable_error(L, $line);
+            pallene_runtime_array_metatable_error(L, $file, $line);
         }
     ]], {
         src = src,
+        file = C.string(loc.file_name),
```
hugomg, line 203: It is a bit distracting to see these string literals repeated at every syntax error. Would it look better if we created a  `#define` at the top of the file to hold the file name?

runtime/pallene_core.h
```diff
@@ -57,20 +57,20 @@ int pallene_runtime_record_index_error(
...
-    lua_State *L, int line)
+    lua_State *L, const char* file, int line)
     PALLENE_NORETURN;
 
 TString *pallene_string_concatN(
     lua_State *L, size_t n, TString **ss);
 
 void pallene_grow_array(
-    lua_State *L, Table *arr, unsigned int ui, int line);
+    lua_State *L, Table *arr, unsigned int ui, const char* file, int line);
```
hugomg, line 67: In some of the functions the file+line are the first arguments while in others they are the last. I think this might be because in the past the argument order reflected the order that these things appeared in the error message.

We could take this opportunity to make things more consistent by always putting the file+line at the start. (Since the pallene_runtime_tag_check_error is a vararg function, putting it at the end would not be an option in that case.)

pallene/coder.lua
itssamuelrowe: I remove the print statements before we merge, I find them useful. :sweat_smile:

PR #305, opened 2020-08-10
Removed the extra location fields from the AST.

itssamuelrowe:
In this PR, we got rid of all the extra location fields from the AST. Instead of adding the
start and end locations of type annotations to the AST, we could have the parser return a list
of type annotation regions in addition to returning the AST. This helps us get rid of the location
fields that are cluttering the AST. Further, this will allow us to get rid of most of the recursive
functions in `translator.lua`.


pallene/parser.lua
```diff
@@ -17,6 +17,7 @@ function Parser:init(lexer)
     self.loop_depth = 0
     self.next = false -- Token
     self.look = false -- Token
+    self.regions = {} -- Sequence of tables
```
hugomg: We could be a bit more explicit in describing what those tables are.

pallenec
```diff
@@ -21,6 +22,7 @@ p:mutex(
     p:flag("--emit-lua",  "Generate a .lua file instead of an .so file"),
     p:flag("--compile-c", "Compile a .c file generated by --emit-c"),
     p:flag("--print-ir",  "Print a human-readable version of the program's IR"),
+    p:flag("--tokens", "Print the tokens recognized by the lexer"),
```
hugomg, line 25: This is a nice feature! But could we change it into `--dump=lexer`, to be more consistent  with the others?

pallene/parser.lua
```diff
@@ -88,18 +116,24 @@ function Parser:Program()
     while not self:peek("EOF") do
         table.insert(tls, self:Toplevel())
     end
-    return tls
+    return {
+        tls = tls,
+        regions = self.regions
+    }
```
hugomg, line 122: What do you think about creating a new kind of AST node, ast.Program.Program? It would be nice to have all the field names listed in ast.lua.

pallene/parser.lua
```diff
@@ -79,6 +80,33 @@ function Parser:loop_end()
...
+function Parser:is_inside_region()
+    local result = false
+    if #self.regions > 0 then
+        local region = self.regions[#self.regions]
+        result = not region[2]
+    end
+    return result
+end
+
+-- The ranges are inclusive.
```
hugomg, line 92: I think it would be nice if the comment made it clear that the purpose of these functions is to implement the Pallene to Lua translator. Similartly to how the comment before loop_begin and loop_end says that the purpose is to catch invalid break statements.

pallene/parser.lua
```diff
@@ -79,6 +81,24 @@ function Parser:loop_end()
...
+end
+
+function Parser:region_end()
+    assert(self.region_depth > 0)
+    self.region_depth = self.region_depth - 1
+    if self.region_depth == 0 then
+        local region = self.regions[#self.regions]
+        region[2] = self.next.loc.pos - 1
+    end
+
```
hugomg, line 99: We can remove this space.

pallene/parser.lua
```diff
@@ -126,9 +153,15 @@ function Parser:Toplevel()
             local oparen   = self:e("(")
             local params   = self:DeclList()
             local _        = self:e(")", oparen)
-            local rt_colon = self.next.loc
-            local rt_types = self:try(":") and self:RetTypes() or {}
-            local rt_end   = self.next.loc
+
+            local rt_types = {}
+            if self.next.name == ":" then
```
hugomg, line 158: Please use self:peek() here.

pallene/parser.lua
```diff
@@ -578,9 +612,11 @@ end
 function Parser:CastExp()
     local exp = self:SimpleExp()
     while self:peek("as") do
-        local op  = self:e()
+        local op = self:e()
+        self:region_begin()
```
hugomg, line 616: I think we might want to put the region_begin before the self:e().

pallenec
```diff
@@ -74,6 +76,29 @@ local function dump(filename, stop_after)
...
+            local loc = token.loc
+            local end_loc = token.end_loc
+            local value = ""
+            if token.value then
+                value = string.format(" (%s)", token.value)
+            end
+            local s = string.format("%d:%d-%d:%d %s%s", loc.line, loc.col,
+                end_loc.line, end_loc.col, token.name, value)
+            print(s)
+            token = lexer:next()
```
hugomg, line 95: We can remove the duplicated calls to lexer:next() if we rewrite the loop as 

```
while true do
     local token = lexer:next()
     if token.name == "EOF" then break end
     --- ...
end
```

pallenec
```diff
@@ -74,6 +76,29 @@ local function dump(filename, stop_after)
...
+        while token.name ~= "EOF" do
+            local loc = token.loc
+            local end_loc = token.end_loc
+            local value = ""
+            if token.value then
+                value = string.format(" (%s)", token.value)
+            end
+            local s = string.format("%d:%d-%d:%d %s%s", loc.line, loc.col,
+                end_loc.line, end_loc.col, token.name, value)
+            print(s)
```
hugomg, line 94: I think we could simplify this to `print(string.format( ))`

pallene/parser.lua
```diff
@@ -17,6 +17,8 @@ function Parser:init(lexer)
     self.loop_depth = 0
     self.next = false -- Token
     self.look = false -- Token
+    self.regions = {} -- Sequence of sequence
+    self.region_depth = 0
```
hugomg, line 21: I think it is better if the `self.next` and `self.look` are next to the calls to `self:_advance()`. Maybe we should reorder  the variables to be the same order as the methods.

```
self.next = false -- Token
self.look = false -- Token
self:_advance(); self:_advance()
self.loop_depth = 0
self.region_depth = 0
self.regions = {}
```

pallene/checker.lua
```diff
@@ -226,15 +225,15 @@ local letrec_groups = {
 }
 
 function Checker:check_program(prog_ast)
-
+    local tls = prog_ast.tls
```
hugomg, line 228: Please add an `assert(prog_ast._tag == "ast.Progra.Program")` before this line. Generally speking, whenever you have a piece of code that assumes that the tag must be a certain type, we should add an assert to help catch it.

For example, if someone passes a list of toplevel nodes instead of a prog_ast then this assignment will silently succeed, returning nil... The error will only be caught further down when we do an ipairs on it.

pallene/to_ir.lua
```diff
@@ -21,7 +21,7 @@ typedecl.declare(to_ir, "to_ir", "LHS", {
 local ToIR = util.Class()
 
 function to_ir.convert(prog_ast)
-    local module = ToIR.new():convert_toplevel(prog_ast)
+    local module = ToIR.new():convert_toplevel(prog_ast.tls)
```
hugomg, line 24: Like the checker.convert_program, this should also have an assertion that the `prog_ast._tag` is what you expect it to be.

pallene/translator.lua
```diff
@@ -95,205 +85,37 @@ function Translator:add_exports()
...
-    else
-        error("impossible")
-    end
-end
+    local names = {}
+    for _, node in ipairs(prog_ast.tls) do
+        -- Build the exports and forward declaration table.
+        if node._tag == "ast.Toplevel.Var" then
+            -- Add the variables to the export sequence if they are declared with the `export`
+            -- modifier.
```
hugomg, line 96: I think we can remove this comment.

pallenec
```diff
@@ -63,15 +64,42 @@ local inspect_opts = {
...
     local input, err = driver.load_input(filename)
     if err then util.abort(err) end
 
-    local out, errs = driver.compile_internal(filename, input, stop_after, opt_level)
-    if not out then util.abort(table.concat(errs, "\n")) end
+    if stop_after == "lexer" then
+        local lexer = driver.compile_internal(filename, input, stop_after, opt_level)
+        print_tokens(lexer)
+    else
+        local out, errs = driver.compile_internal(filename, input, stop_after, opt_level)
```
hugomg, line 97: The local `out, errs` is the same in the two branches of the `if`. We could put it before the `if`.

spec/parser_spec.lua
```diff
@@ -47,6 +47,8 @@ end
 
 local function assert_program_ast(program_str, expected_ast)
     local prog_ast = assert_parses_successfuly(program_str)
+    print(require('inspect')(prog_ast))
+    print(require('inspect')(expected_ast))
```
hugomg, line 51: ...

spec/parser_spec.lua
```diff
@@ -153,14 +153,22 @@ describe("Pallene parser", function()
...
-        assert.are.same({}, prog_ast)
+        assert.are.same({}, prog_ast.tls)
     end)
 
     it("can parse toplevel var declarations", function()
-        assert_program_ast([[ local x=17 ]], {
-            { _tag = "ast.Toplevel.Var",
-                decls = { { name = "x", type = false } } }
-        })
+        assert_program_ast([[ local x=17 ]],
```
hugomg, line 160: I think it would be more readable if the assert_program_ast continued to receive a list of tls as a parameter. We we could  add the ast.Program.Program tag in the call to `assert_is_subset`.

pallenec
```diff
@@ -63,15 +64,42 @@ local inspect_opts = {
...
+        local loc = token.loc
+        local end_loc = token.end_loc
+        local value = ""
+        if token.value then
+            local s = token.value
+            if type(token.value) == "string" then
+                s = C.string(token.value)
+            end
+            value = string.format(" %s", s)
+        end
```
hugomg, line 83: Suggestion: 

```
local value
if token.value then
    if type(token.value) == "string" then
        value = " " .. C.string(token.value)
    else
        value = " " .. tostring(token.value)
    end
else
    value = ""
end
```

I think an if with an else is easier to understand what is going on than an if without an else, where the value of the variable is overwritten.

pallenec
```diff
@@ -63,15 +64,42 @@ local inspect_opts = {
...
+        print(string.format("%d:%d,%d:%d,%d:%d %s%s", loc.line, loc.col,
+            end_loc.line, end_loc.col, loc.pos, end_loc.pos, token.name, value))
```
hugomg, line 85: It isn't super clear that the first two `%d:%d` are line and column, but the third `%d:%d` are absolute positions.  Can you think of a way to make it clearer that they are different things?

pallene/parser.lua
```diff
@@ -14,17 +14,28 @@ local Parser = util.Class()
 
 function Parser:init(lexer)
     self.lexer = lexer
-    self.loop_depth = 0
     self.next = false -- Token
     self.look = false -- Token
+    self.comment_regions = {} -- Sequence of sequence
```
hugomg, line 19: I've been thinking... What do you think about computing the list of comment regions in the lexer instead of in the parser? It would avoid adding extra complexity to the parser.

pallene/Lexer.lua
```diff
@@ -266,8 +267,19 @@ function Lexer:next()
         if not name then
             return false, val
         end
-        if name ~= "SPACE" and name ~= "COMMENT" then
-            return { name = name, value = val, loc = loc }
+
+        local end_loc = self:loc()
+        if name == "COMMENT" then
+            -- For a comment, `end_loc.pos` points at the starting position of the next token.
+            -- Therefore, we subtract 1 from the position to adjust this anamoly.
```
hugomg, line 274: I think we can get rid of this comment. This isn't an anomaly. `self:loc()` points to the start of the next token, meaning that `end_loc.pos` pointing to one character past the end of the comment is to be expected.

pallene/parser.lua
```diff
@@ -14,17 +14,20 @@ local Parser = util.Class()
 
 function Parser:init(lexer)
     self.lexer = lexer
-    self.loop_depth = 0
     self.next = false -- Token
     self.look = false -- Token
     self:_advance(); self:_advance()
+    self.loop_depth = 0
+    self.region_depth = 0
+    self.regions = {} -- Sequence of sequence. Contains the ranges where type annotations span.
```
hugomg, line 22: Technically, this is a list of pairs, not a list of sequence.

pallene/parser.lua
```diff
@@ -79,6 +82,24 @@ function Parser:loop_end()
     self.loop_depth = self.loop_depth - 1
 end
 
+-- The region_begin() and region_end() methods are used to mark the regions which the translator
+-- removes. The regions are packaged as part of the AST. The ranges of the regions are inclusive.
```
hugomg, line 86: Can we call it the Pallene to Lua translator instead of just "the translator"?

pallenec
```diff
@@ -63,15 +64,43 @@ local inspect_opts = {
...
 local function dump(filename, stop_after)
     local input, err = driver.load_input(filename)
     if err then util.abort(err) end
 
     local out, errs = driver.compile_internal(filename, input, stop_after, opt_level)
-    if not out then util.abort(table.concat(errs, "\n")) end
+    if stop_after == "lexer" then
+        print_tokens(out)
+    else
+        if not out then util.abort(table.concat(errs, "\n")) end
```
hugomg, line 99: Should this error handling be right after the call to `compile_internal`, before the `if stop_after`?

spec/lexer_spec.lua
```diff
@@ -73,8 +73,8 @@ describe("Pallene lexer", function()
         assert_lex("~=",  {"~="},          {})
         assert_lex("~~=", {"~", "~="},     {})
         assert_lex("->-", {"->", "-"},     {})
-        assert_lex("---", {},              {})
-        assert_lex("-->", {},              {})
+        assert_lex("---", {},     {})
+        assert_lex("-->", {},     {})
```
hugomg, line 77: Can we reindent these back to where they were before, to clean up the diff?

spec/parser_spec.lua
```diff
@@ -219,65 +228,83 @@ describe("Pallene parser", function()
...
-                            { _tag = "ast.Type.Float" }, }, } },
-                value = {
-                    _tag = "ast.Exp.Lambda",
-                    arg_decls  = {},
-                    body = {
-                        _tag = "ast.Stat.Block",
-                        stats = {} } } },
-        })
+        ]],
+        {{
```
hugomg, line 232: Could we keep theses tests indented the way they were before? I'm not sure I like this `{{` at the start -- it doesn't scale well to ASTs where there is more than one toplevel node.

Another problem is that making this kind of change makes the diff harder to read. Generally speaking it is better to put this kind of indentation change in a separate commit/PR.

PR #309, opened 2020-08-16
Run the benchmarks and examples in the test suite

### LINKED ISSUE #207: We should automatically test the example programs and the benchmarks.
I just received an email pointing out two things that recently broke:
1. example/rectangle.pln has a syntax error because it was not updated to use the `typeealias` keyword.
2. benchmarks/binarytrees/pallene.pln gives a compilation warning (started after merging the multiple-returns branch)

We should have some kind of automated test to detect these. My only question is if I should add these tests as part of the main `busted` test suite or if I should put them in a separate test script that is run in the .travis.yml. If possible  I would like to avoid having the busted test suit take much longer to run than it does now.

hugomg:
Fixes #207.


hugomg:
The tests are passing on my machine but they are failing on Travis because it doesn't have LuaJIT installed. We'll need to figure out what would be the best way to do that.

hugomg:
Thinking a bit more, I think  it would be better if there was no hard dependency on LuaJIT to get the tests to pass. This way, if someone just wants to try out Pallene they don't need to also install LuaJIT just to be able to run those tests.

That said, it would still be nice if the Travis test suite could test the LuaJIT stuff. I'm opening a separate issue for that.

PR #316, opened 2020-08-20
Updated the translator to prevent redundant spaces before assignment …

itssamuelrowe:
…and as operator.

Added a utility method to the translator: `add_partial`. It ensures that the partials inserted are
not empty. This constraint is required by the translator when removing redundant spaces.

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -14,6 +14,7 @@ local Parser = util.Class()
 
 function Parser:init(lexer)
     self.lexer = lexer
+    self.prev = false -- Token
     self.next = false -- Token
     self.look = false -- Token
     self.loop_depth = 0       -- Are we inside a loop?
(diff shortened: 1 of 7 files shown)


pallene/translator.lua
```diff
@@ -44,44 +44,42 @@ end
...
-
-        table.insert(self.partials, first_partial)
-        table.insert(self.partials, last_partial)
-    else
-        local partial = region:gsub("[^\n\r]", "")
-        table.insert(self.partials, partial)
+    if #self.partials > 0 then
+        local last_partial = self.partials[#self.partials]
+        local space_index = last_partial:find(" *$")
+        self.partials[#self.partials] = last_partial:sub(1, space_index - 1)
```
hugomg, line 65: Can we change the ` *$` to ` +$` and put this `:sub` inside an if statement?

pallene/parser.lua
```diff
@@ -14,6 +14,7 @@ local Parser = util.Class()
 
 function Parser:init(lexer)
     self.lexer = lexer
+    self.previous = false -- Token
```
hugomg, line 17: Can we call this `self.prev`, so everything has 4 characters? ^_^

PR #323, opened 2020-08-24
Parallel tests

hugomg:
This PR aims to speed up how long it takes to run the test suite. Overall I got about a 2x improvement on my machine. The main bottleneck now seems to be how long it takes to run the coder_spec.

1. If you ask to run the entire test suite it, it now runs the tests in parallel by default. 
2. The pretty printer tests are now on a separate spec file. Speeds up how long it takes to run the coder_spec by itself.
3. The coder_spec and translator_spec now use different file names instead of both using \_\_test\_\_. (This is required to be able to run in parallel)


hugomg:
If you don't have GNU parallel installed it prints a message to the terminal saying that. But I suppose that also mentioning it in the README wouldn't hurt...

Issue #225, opened 2020-06-03
Should we show a better stack trace instead of error("impossible")?


hugomg:
At the end of and if-else-if chain we use `error("impossible")` to mark a type error when someoone passes in the wrong kind of table.

```
local tag = exp._tag
if tag == "ast.Nil" then
    handle_nil()
elseif tag == "ast.Exp.Bool" then
    handle_bool()
-- ...
else
    error("impossible")
end
```

The problem is that whenever the `error("impossible")` happens, we don't get any information about what was the tag that caused the problem. To debug we typically have to add a temporary "print" statement.

I propose that we create a new helper function to show a better error message for these errors. One option would be something like `typedecl.tag_error(tag)`.

hugomg:
If we implement this, don't forget to grep for all uses of the words error and impossible. There are some variations in the codebase such as `error(tag .. " impossible")`.

inJuly0:
Hi there, after taking a thorough look at the Pallene's Lua front end, I'd like to work on this issue.
I'm assuming the issue refers to places like this one for example, in `types.lua`: 

```lua
function types.indices(t)
    local tag = t._tag
    if     tag == "types.T.Table" then
        return t.fields

    elseif tag == "types.T.Record" then
        return t.field_types

    else
        error("impossible")
    end
end
```

more specifically, the `error("impossible")` at the end ?

I'd like to propose an idea that a more descriptive error message would be easier to debug. Say, for example we could do :
`typedecl.tag_error(tag, "cannot index type.")`

which would  then show a proper error message like `Boolean: cannot index this type.`
the second argument to`typedecl.tag_error` could be optional, and an "impossble" is thrown in it's absence.

An even further improvement (in my opinion) could be an optionally interpolated 2nd argument, like so:
`typedecl.tag_error(tag, "cannot index type ${tag} .")`.

This is similar to what I've been using to debug internal compiler errors and type errors in my langauge, and I think it would be a good fit for Pallene too. :)

hugomg:
We appreciate any help here :)

> I'm assuming the issue refers to places like `error("impossible")`

Exactly!

> I'd like to propose an idea that a more descriptive error message would be easier to debug.

I think the thing that would help the most is if the `tag_error(tag)` function included the tag in the error message, instead of just saying "impossible".

>Say, for example we could do `typedecl.tag_error(tag, "cannot index type.")`

In this particular example, the `error("impossible")` is taking care of two kinds of errors

[...]

PR #329, opened 2020-09-16
Fix to #225, added better error reporting for internal compiler errors.

inJuly0:
Previously, when an incorrect table was passed into a function dealing with a certain type of table, only an "impossible" error was thrown followed by a Lua stack trace.

Now it's possible to see the tag which caused the error, and wherever appropriate, an error message following it. Calls to `error("impossible")` have been replaced with a call to `typedecl.tag_error(tag, "(optional error message.)")`.

Hopefully, this should fix [#225 ](https://github.com/pallene-lang/pallene/issues/225).

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -124,7 +124,7 @@ declare_type("Name", {
 })
 
 function Checker:add_type(name, typ)
-    assert(string.match(typ._tag, "^types%.T%."))
+    assert(typedecl.tag_matches(typ._tag, "types.T."))
     self.symbol_table:add_symbol(name, checker.Name.Type(typ))
 end
 
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -218,7 +218,7 @@ function Checker:from_ast_type(ast_typ)
         return types.T.Function(p_types, ret_types)
 
     else
-        error("impossible")
+        typedecl.tag_error(tag)
     end
 end
 
@@ -360,7 +360,7 @@ function Checker:check_program(prog_ast)
                     self:add_type(tl_node.name, typ)
 
                 else
-                    error("impossible")
+                    typedecl.tag_error(tag)
                 end
             end
 
@@ -455,7 +455,7 @@ function Checker:check_stat(stat)
             elseif loop_type._tag == "types.T.Float" then
                 stat.step = ast.Exp.Float(stat.limit.loc, 1.0)
             else
-                error("impossible")
+                typedecl.tag_error(loop_type._tag, "incorrect loop step type.")
             end
         end
 
@@ -515,7 +515,7 @@ function Checker:check_stat(stat)
         -- ok
 
     else
-        error("impossible")
+        typedecl.tag_error(tag)
     end
 
     return stat
@@ -547,7 +547,7 @@ function Checker:check_var(var)
                 "cannot reference module name '%s' without dot notation",
                 var.name)
         else
-            error("impossible")
+            typedecl.tag_error(cname._tag)
         end
 
     elseif tag == "ast.Var.Dot" then
@@ -603,7 +603,7 @@ function Checker:check_var(var)
         var._type = arr_type.elem
 
     else
-        error("impossible")
+        typedecl.tag_error(tag)
     end
     return var
 end
@@ -618,8 +618,10 @@ function Checker:coerce_numeric_exp_to_float(exp)
         return exp
     elseif tag == "types.T.Integer" then
         return self:check_exp_synthesize(ast.Exp.ToFloat(exp.loc, exp))
+    elseif typedecl.tag_is_type(tag) then
+        typedecl.tag_error(tag, "this type cannot be coerced to float.")
     else
-        error("impossible")
+        typedecl.tag_error(tag)
     end
 end
 
@@ -687,7 +689,7 @@ function Checker:check_exp_synthesize(exp)
             check_type_is_condition(exp.exp, "'not' operator")
             exp._type = types.T.Boolean()
         else
-            error("impossible")
+            typedecl.tag_error(tag, string.format("unknown unary operator '%s'.", op))
         end
 
     elseif tag == "ast.Exp.Binop" then
@@ -797,7 +799,7 @@ function Checker:check_exp_synthesize(exp)
             exp._type = types.T.Integer()
 
         else
-            error("impossible")
+            typedecl.tag_error(tag, string.format("unknown binary operator '%s'.", op))
         end
 
     elseif tag == "ast.Exp.CallFunc" then
@@ -865,7 +867,7 @@ function Checker:check_exp_synthesize(exp)
         exp._type = types.T.Float()
 
     else
-        error("impossible")
+        typedecl.tag_error(tag)
     end
 
     return exp
@@ -898,7 +900,7 @@ function Checker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
                         field.exp, expected_type.elem,
                         "array initializer")
                 else
-                    error("impossible")
+                    typedecl.tag_error(ftag)
                 end
             end
 
@@ -928,7 +930,7 @@ function Checker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
                         field.exp, field_type,
                         "table initializer")
                 else
-                    error("impossible")
+                    typedecl.tag_error(ftag)
                 end
             end
 
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -618,7 +618,7 @@ function Checker:coerce_numeric_exp_to_float(exp)
         return exp
     elseif tag == "types.T.Integer" then
         return self:check_exp_synthesize(ast.Exp.ToFloat(exp.loc, exp))
-    elseif typedecl.tag_is_type(tag) then
+    elseif typedecl.tag_matches(tag, "types.T") then
         typedecl.tag_error(tag, "this type cannot be coerced to float.")
     else
         typedecl.tag_error(tag)
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -455,7 +455,7 @@ function Checker:check_stat(stat)
             elseif loop_type._tag == "types.T.Float" then
                 stat.step = ast.Exp.Float(stat.limit.loc, 1.0)
             else
-                typedecl.tag_error(loop_type._tag, "incorrect loop step type.")
+                typedecl.tag_error(loop_type._tag, "loop type is not a number.")
             end
         end
 
@@ -799,7 +799,7 @@ function Checker:check_exp_synthesize(exp)
             exp._type = types.T.Integer()
 
         else
-            typedecl.tag_error(tag, string.format("unknown binary operator '%s'.", op))
+            typedecl.tag_error(op)
         end
 
     elseif tag == "ast.Exp.CallFunc" then
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -39,7 +39,7 @@ local function ctype(typ)
     elseif tag == "types.T.Table"    then return "Table *"
     elseif tag == "types.T.Record"   then return "Udata *"
     elseif tag == "types.T.Any"      then return "TValue"
-    elseif typedecl.tag_is_type(tag) then
+    elseif typedecl.tag_matches(tag, "types.T") then
         typedecl.tag_error(tag, "unable to get corresponding c type.")
     else typedecl.tag_error(tag)
     end
@@ -119,7 +119,7 @@ local function set_stack_slot(typ, dst_slot, value)
     elseif tag == "types.T.Table"    then tmpl = "sethvalue(L, $dst, $src);"
     elseif tag == "types.T.Record"   then tmpl = "setuvalue(L, $dst, $src);"
     elseif tag == "types.T.Any"      then tmpl = "setobj(L, $dst, &$src);"
-    elseif not typedecl.tag_is_type(tag) then
+    elseif not typedecl.tag_matches(tag, "types.T") then
         typedecl.tag_error(tag, "not a type.")
     else
         typedecl.tag_error(tag)
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -39,7 +39,9 @@ local function ctype(typ)
     elseif tag == "types.T.Table"    then return "Table *"
     elseif tag == "types.T.Record"   then return "Udata *"
     elseif tag == "types.T.Any"      then return "TValue"
-    else error("impossible")
+    elseif typedecl.tag_is_type(tag) then
+        typedecl.tag_error(tag, "unable to get corresponding c type.")
+    else typedecl.tag_error(tag)
     end
 end
 
@@ -90,7 +92,7 @@ local function lua_value(typ, src_slot)
     elseif tag == "types.T.Table"    then tmpl = "hvalue($src)"
     elseif tag == "types.T.Record"   then tmpl = "uvalue($src)"
     elseif tag == "types.T.Any"      then tmpl = "*($src)"
-    else error("impossible")
+    else typedecl.tag_error(tag)
     end
     return (util.render(tmpl, {src = src_slot}))
 end
@@ -117,8 +119,12 @@ local function set_stack_slot(typ, dst_slot, value)
     elseif tag == "types.T.Table"    then tmpl = "sethvalue(L, $dst, $src);"
     elseif tag == "types.T.Record"   then tmpl = "setuvalue(L, $dst, $src);"
     elseif tag == "types.T.Any"      then tmpl = "setobj(L, $dst, &$src);"
-    else error("impossible")
+    elseif not typedecl.tag_is_type(tag) then
+        typedecl.tag_error(tag, "not a type.")
+    else
+        typedecl.tag_error(tag)
     end
+
     return (util.render(tmpl, { dst = dst_slot, src = value }))
 end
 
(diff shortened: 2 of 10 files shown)


pallene/types.lua
```diff
@@ -246,8 +251,10 @@ function types.tostring(t)
 
     elseif tag == "types.T.Record" then
         return t.name
+    elseif not typedecl.tag_matches(tag, "types.T") then
+        typedecl.tag_error(tag, "not a type.")
```
hugomg, line 255: I also think we might not need this elseif

pallene/checker.lua
```diff
@@ -455,7 +455,7 @@ function Checker:check_stat(stat)
             elseif loop_type._tag == "types.T.Float" then
                 stat.step = ast.Exp.Float(stat.limit.loc, 1.0)
             else
-                error("impossible")
+                typedecl.tag_error(loop_type._tag, "incorrect loop step type.")
```
hugomg, line 458: Nice. This is exactly the kind of situation that benefits the most from a custom error message.

Small suggestion: what do you think about changing the error message to `loop type is not a number type`?

pallene/checker.lua
```diff
@@ -687,7 +689,7 @@ function Checker:check_exp_synthesize(exp)
             check_type_is_condition(exp.exp, "'not' operator")
             exp._type = types.T.Boolean()
         else
-            error("impossible")
+            typedecl.tag_error(tag, string.format("unknown unary operator '%s'.", op))
```
hugomg, line 692: With the new default error message I proposed, I think we could use `typedecl.tag_error(op)`.

pallene/checker.lua
```diff
@@ -797,7 +799,7 @@ function Checker:check_exp_synthesize(exp)
             exp._type = types.T.Integer()
 
         else
-            error("impossible")
+            typedecl.tag_error(tag, string.format("unknown binary operator '%s'.", op))
```
hugomg, line 802: Again, the error message should reference `op`, not the `tag`. With the error message I suggested I think we could use `typedecl.tag_error(op)`.

pallene/coder.lua
```diff
@@ -39,7 +39,9 @@ local function ctype(typ)
     elseif tag == "types.T.Table"    then return "Table *"
     elseif tag == "types.T.Record"   then return "Udata *"
     elseif tag == "types.T.Any"      then return "TValue"
-    else error("impossible")
+    elseif typedecl.tag_is_type(tag) then
+        typedecl.tag_error(tag, "unable to get corresponding c type.")
```
hugomg, line 43: I think the default error message is fine here.

pallene/coder.lua
```diff
@@ -117,8 +119,12 @@ local function set_stack_slot(typ, dst_slot, value)
     elseif tag == "types.T.Table"    then tmpl = "sethvalue(L, $dst, $src);"
     elseif tag == "types.T.Record"   then tmpl = "setuvalue(L, $dst, $src);"
     elseif tag == "types.T.Any"      then tmpl = "setobj(L, $dst, &$src);"
-    else error("impossible")
+    elseif not typedecl.tag_is_type(tag) then
+        typedecl.tag_error(tag, "not a type.")
```
hugomg, line 123: This is another case where I think the default error message is ok.

pallene/coder.lua
```diff
@@ -165,8 +171,8 @@ local function pallene_type_tag(typ)
     elseif tag == "types.T.Array"    then return "LUA_TTABLE"
     elseif tag == "types.T.Table"    then return "LUA_TTABLE"
     elseif tag == "types.T.Record"   then return "LUA_TUSERDATA"
-    elseif tag == "types.T.Any"    then error("value is not a tag")
-    else error("impossible")
+    elseif tag == "types.T.Any"      then typedecl.tag_error(tag, "value is not a tag")
```
hugomg, line 174: I just noticed that this error message is outdated. In the past, the `any` type used to be called `value` but we forgot to update the error message when that changes.

Could you please change the error message to `"Any" is not a Lua type tag`?

pallene/coder.lua
```diff
@@ -324,8 +330,10 @@ function Coder:c_value(value)
         local f_id = value.id
         local typ = self.module.functions[f_id].typ
         return lua_value(typ, self:function_upvalue_slot(f_id))
+    elseif typedecl.tag_matches(tag, "ir.Value") then
+        typedecl.tag_error(tag, "unable to get C expression for this value type.")
     else
-        error("impossible")
+        typedecl.tag_error(tag, "not a value.")
```
hugomg, line 336: I think it is ok if we use the default error message here.

pallene/to_ir.lua
```diff
@@ -495,10 +497,10 @@ function ToIR:exp_to_value(cmds, exp, _recursive)
                 return ir.Value.Function(id)
 
             elseif cname._tag == "checker.Name.Builtin" then
-                error("not implemented")
+                typedecl.tag_error(cname._tag, "not implemented")
```
hugomg, line 500: If the error happens, it should be clear what the type tag was because of the preceding if statement. I think we can continue using `error("not implemented")` here.

pallene/to_ir.lua
```diff
@@ -775,7 +780,9 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
             elseif src_typ._tag == "types.T.Any" then
                 table.insert(cmds, ir.Cmd.FromDyn(loc, dst_typ, dst, v))
             else
-                error("impossible")
+                typedecl.tag_error(tag,
+                    string.format("error casting from type '%s' to '%s'",
+                        types.tostring(src_typ), types.tostring(dst_typ)))
```
hugomg, line 785: I think the `tag_error(tag` is not appropriate here, because the tags that actually matter for the error are `dst_typ._tag` and `src_typ.tag` and not the `exp._tag`.

I like the new error message but I think we could just use a plain `error(string.format(` instead.

pallene/typedecl.lua
```diff
@@ -82,4 +82,35 @@ function typedecl.declare(module, mod_name, type_name, constructors)
...
+
+-- Throw an error at the given type tag.
+--
+-- @param tag The type tag at which the error is to be thown (string)
+-- @param message The optional error message. (?string)
+
+function typedecl.tag_error(tag, message)
+    local error_msg = message
+        and string.format("error at tag '%s': %s", tag, message)
+        or string.format("error at tag '%s'.", tag)
```
hugomg, line 111: What do you think about changing the error message to not mention the word "tag". This way it would also be usable in other kinds of if-elseif like the binary operators and the builtin function names. Suggestion:

```
function typedecl.tag_error(tag, message)
    message = message or "input has the wrong type or an elseif case is missing"
    error(string.format("the if-elseif did not handle case %s: %s", tag, message))
end
```

pallene/to_ir.lua
```diff
@@ -649,7 +651,7 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
                 assert(#xs == 1)
                 table.insert(cmds, ir.Cmd.BuiltinTostring(loc, dsts, xs))
             else
-                error("impossible")
+                typedecl.tag_error(cname._tag, "unknown builtin function.")
```
hugomg, line 654: This tag_error should be referencing the `bname`, not the `cname._tag`.

```
typedecl.tag_error(bname)
```

hugomg (changes requested):
Lovely! Thanks for going through this. I think you even found some cases that I would have missed if I tried doing this myself.

I do have some suggestion for changes before we can merge this though.

The first is that if we choose a good default error message for tag_error then I think we should be able to use the default error message in more places. I think it would be good if we had a clear rule for whether we should have a special error message or not. That way, we don't need to decide for every possible if-elseif whether it should have a special error message or not. One rule of thumb we could follow is that if the if-elseif covers all the possible cases then it can use the default error message and if it only covers a subset of the possible cases then it should have a special error message.

Another thing that we could change is that there are some places in the code that could be updated to use  the new `tag_matches` function. You can find them by running `grep -R 'string.match(.*tag'`.

pallene/typedecl.lua
```diff
@@ -82,4 +82,35 @@ function typedecl.declare(module, mod_name, type_name, constructors)
     end
 end
 
+-- Returns true if the 2nd argument is a prefix of the first.
+--
+-- @param tag: The type name (string)
+-- @param tag_prefix: The prefix to test (string)
+
```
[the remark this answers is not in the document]
inJuly0: Ah, I somehow missed that. will do.

pallene/typedecl.lua
```diff
@@ -82,4 +82,35 @@ function typedecl.declare(module, mod_name, type_name, constructors)
...
+function typedecl.tag_matches(tag, tag_prefix)
+    return type(tag) == "string"
+        and string.find(tag, tag_prefix, 1, true) ~= nil
```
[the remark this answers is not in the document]
inJuly0: Oh shoot, I forgot about that too.

 Yeah, I definitely have intentions of making contributions in the future since I love compilers and have been exploring type systems as of late.

I actually messed this up because I was coding on my Linux PC, If I manage to get Pallene built up and running on windows these silly errors most likely won't persist.

pallene/coder.lua
inJuly0: will do

pallene/types.lua
```diff
@@ -194,7 +198,8 @@ local function equivalent(t1, t2, is_gradual)
         return t1 == t2
 
     else
-        return error("impossible")
+        return typedecl.tag_error(tag1,
+            string.format("attempt to check equivalence of types %s and %s.", tag1, tag2))
```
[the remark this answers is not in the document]
inJuly0: Oh, I noticed that in places where `string.matches` is used, the prefix being tested is `"types.T."`.
I think it would make more sense to check every tag prefix with the the second "." included, no ?

pallene/checker.lua
inJuly0: Yeah, this error message is a bit unclear. fixed.

pallene/typedecl.lua
inJuly0: Yes, I think that would allow for more uses of `tag_error`.
My reasoning was that the function name implies the error is always associated with a tag.

However I do agree this allows for more liberal use of the function. I'll change it. :+1:

inJuly0:
I made the changes as asked, excuse the CI fails in older commits, I made some mistakes by merging into master too early. Everything should now work as expected.

One part I couldn't figure out is this line, in coder.lua:1508 :

```lua
local name = assert(string.match(cmd._tag, "^ir%.Cmd%.(.*)$"))
```

I had originally replaced that with this:

```lua
local name = cmd._tag
assert(typedecl.tag_matches(name, "ir.Cmd."))
```
however, doing so would cause the tests to fail, and I seem to be unable to diagnose why. 
So I left that one line unchanged. It would be awesome if you could explain that. cheers.

hugomg:
> string.match(cmd._tag, "^ir%.Cmd%.(.*)$")
> however, doing so would cause the tests to fail, and I seem to be unable to diagnose why.

The `string.match()` here is using a string capture. For example, if `cmd._tag` is `ir.Cmd.Binop` then `name` will be `Binop`.

Maybe we should change the tag_matches function to return this string instead of just true/false?

hugomg (changes requested):
Thanks. I marked all the things that you fixed as resolved. There are only a handful that were missed that are still there:

- The tag_error for unknown unary operators
- There is stll an `assert(string.match(typ._tag, "^types%.T%."))` in types.lua
- There is the `name = assert(string.match(cmd._tag, "^ir%.Cmd%.(.*)$"))` that broke the CI

hugomg:
About the broken CI, if we really wanted we could use `git rebase -i` to fix the old commits so they pass the CI. However, in this case it is only a very small problem so I don't think we have to bother with it.

inJuly0:
> The `string.match()` here is using a string capture. For example, if `cmd._tag` is `ir.Cmd.Binop` then `name` will be `Binop`.
> 
> Maybe we should change the tag_matches function to return this string instead of just true/false?

Yeah that would seem like a good idea. 
I missed the capture group, kinda rusty with regex. Thanks for explaining.

I feel like the name `tag_matches` implies a boolean. What would you say about letting that one line stay as is instead of changing intent of the function for just one usage ? Thoughts ?

inJuly0:
> Good idea. It would be safer to always add the second .. What do you think about adding this . inside the tag_matches function, so the caller doesn't have to remember to add it themselves?

Since the `tag_matches` function does prefix matching, I personally would find it easier if it only matched against the string it receives as an argument instead of adding a "`.`". It could, at times cause confusion. Thoughts ?

hugomg:
>  I personally would find it easier if it only matched against the string it receives as an argument instead of adding a ".".

Maybe it should have a different name instead of `tag_matches` then? What we we are trying to see is if a certain tag (such as `"types.T.Boolean"`) has a given type (such as `types.T`). We aren't trying to match against an arbitrary string...

> I feel like the name tag_matches implies a boolean. 

If you can think of a good name that doesn't imply a boolean we could use that. If not, I think it is OK if we continue using `string.match("ir%.Cmd%.(.*)")` in that one place.

inJuly0:
> If you can think of a good name that doesn't imply a boolean we could use that.

I went with `match_tag` and made the changes as asked. 
The function now adds a second "." and captures and returns everything following it.

Probably not the best name but this is loosely similar to what `string.match` does in Lua , or Javascript,
match a pattern in a string and return the matched sequence. Except this returns the part following the `.`.

Do let me know if I missed something yet again :D

pallene/typedecl.lua
```diff
@@ -82,4 +82,31 @@ function typedecl.declare(module, mod_name, type_name, constructors)
...
+-- typedecl.match_tag("ast.Exp.Bool", "ast.Exp")
+-- ```
+-- would return "Bool" since `"ast.Exp."` is indeed a prefix of
+-- "ast.Exp.bool". Check types.lua for more usage examples.
+--
+-- @param tag: The type name (string)
+-- @param tag_prefix: The prefix to test (string)
+function typedecl.match_tag(tag, tag_prefix)
+    return type(tag) == "string" and string.match(tag, tag_prefix .. ".(.*)")
+end
```
hugomg, line 101: One problem with using string.match here is that special characters in the tag_prefix will be treated as a regular expression. For example, a prefix of `types.T` would also match `typesaT` or `typesbT`, which is not what we want to happen.

One possible fix is to use string.sub to get the parts of the string that we are looking for:

```

function typedecl.match_tag(tag, tag_prefix)
    local n = #tag_prefix
    if type(tag) == "string" and
       string.sub(tag, 1, n) == tag_prefix and
       string.byte(tag, n+1) == 46 -- "."
    then
        return string.sub(tag, n+2)
    else
        return false
    end
end
```

hugomg (changes requested):
Nice! I like the new name.

I found a couple of small bugs but they should be easy to fix and after that I think this should be good for merging.

It would also be nice if we added some test cases to typedecl_spec.lua. In particular, we should test that something like `typedecl.match_tag("abc.d.e", "a.c")` returns `false` because that is a bug that already hit us before and it happened again today. Would you be able to help by writing those tests?

pallene/print_ir.lua
```diff
@@ -214,7 +215,8 @@ local function Cmd(cmd)
     elseif tag == "ir.Cmd.CallStatic" then rhs = Call(Fun(cmd.f_id),  Vals(cmd.srcs))
     elseif tag == "ir.Cmd.CallDyn"    then rhs = Call(Val(cmd.src_f), Vals(cmd.srcs))
     else
-        local tagname = assert(string.match(cmd._tag, "^ir.Cmd.(.*)"))
+        local tagname = cmd._tag
+        assert(typedecl.match_tag(tagname, "ir.Cmd"))
```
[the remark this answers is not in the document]
inJuly0: Oh yes, my bad.

inJuly0:
> It would also be nice if we added some test cases to typedecl_spec.lua. In particular, we should test that something like typedecl.match_tag("abc.d.e", "a.c") returns false
Yeah, I should be able to put together some tests for that.

inJuly0:
I updated that `assert` guard, changed `match_tag` to no longer treat it's arguments as regex strings.
Added 3 basic test cases to `typedecl_spec.lua`. 

Do let me know if I messed up somewhere. :D

spec/typedecl_spec.lua
```diff
@@ -8,4 +8,10 @@ describe("Typedecl", function()
             typedecl.declare(mod, "TESTTYPE", "Foo", { Bar = {"x"} })
         end, "tag name 'TESTTYPE.Foo.Bar' is already being used")
     end)
+
+    it("'match_tag' works as expected.", function ()
+        assert.falsy(typedecl.match_tag("foo.Bar.baz", "f.o.Bar"))
+        assert.truthy(typedecl.match_tag("foo.Bar.baz", "foo.Bar"))
+        assert.falsy(typedecl.match_tag("types.T.Float", "types.T."))
+    end, "")
```
hugomg, line 16: * We could put each test case in a separate `it` block. That way we can give a descriptive name for each of them. For example the first test can say that it is testing that the "." is not being interpreted as a regex character.
* The second test case could test that the return value is `"baz"` instead of only testing that it is truthy
* I think we don't need to put that `"")` at the end

hugomg (changes requested):
Just some small things in the typedecl_spec and then I think this is done.

spec/typedecl_spec.lua
inJuly0: True, should have gone with more descriptive test logs.
Done :+1:.

hugomg (approved):
Perfect! Thanks a lot for the contribution! It's much appreciated.

Issue #164, opened 2020-01-25
Implement loops with ipairs


hugomg:
It would be nice to be able to use ipairs in Pallene:

```
for i, x in ipairs(xs) do
    -- do something
end
```

Implementing general for-in loops as well would be a nice bonus. But the particular case of ipairs is what I miss the most.

Two things to watch out for when implementing this:

1. `ipairs` is not a keyword. It can be shadowed by local declarations.
2. The ipairs loop is not equivalent to `for i=1, #xs`, if there are holes in the array. The ipairs version always stops at the first hole, while the `#xs` version may or may not do that.

Akylzhan:
I can work on this issue but I do not know how to do this. Can you guide me?

ahmedkrmn:
Hi @Akylzhan.
What do you have in mind to implement this issue?
We can discuss it together to figure out a solution.

Akylzhan:
@ahmedkrmn I did not look the source code in details and I am quite clueless, but the only things that comes to my mind is to create new command for ```gen_cmd``` in ```coder.lua```. Though as @hugomg mentioned, ```ipairs``` should not be keyword; therefore this is not good solution.

Sorry, if I am mistaken

ahmedkrmn:
`ipairs` is currently not a reserved keyword in the language, you can
declare a variable that has the name of `ipairs`.
I believe the first step is to add `ipairs` to the list of reserved words
during the transpiler lexical analysis phase.

Akylzhan:
as I understood, I need to first declare 'ipairs' function in ```pallene/builtins.lua``` and then declare it in ```pallene/ir.lua``` in table ```ir_cmd_constructors``` as ```BuiltinIpairs``` (but did not get what should I include to this table element). Then, add new ```elseif``` after line 419 of ```pallene/to_ir.lua```. Right?

ahmedkrmn:
@Akylzhan before jumping into implementation details, you first have to add `ipairs` to the list of reserved keywords found in `/pallene/lexer.lua` then you should modify the language grammar in `/pallene/parser.lua` to allow for the new syntax introduced by having `ipairs` in for loops.

Akylzhan:
@ahmedkrmn I thought that I need to add ipairs as built_in function

hugomg:
Hi, @Akylzhan, I'm glad to see that you are interested in getting involved in Pallene.  To solve this particular issue, I imagine that it could be broken down in two main tasks:

[...]

PR #332, opened 2020-10-26
Implement looping with ipairs

srijan-paul:
This PR is in reference to issue #164 .
The following changes have been made: 

* The AST now has two different kinds of nodes representing for-loops, `ForNum` for numeric loops and `ForIn` for the other kind.
* The parser now emits `ast.Stat.ForIn` nodes for `for...in` loops , and `ast.Stat.ForNum` nodes for numeric for-loops.
* A test case has been added in `parser_spec.lua` to reflect the same.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -162,21 +162,6 @@ function Checker:from_ast_type(ast_typ)
     if     tag == "ast.Type.Nil" then
         return types.T.Nil()
 
-    elseif tag == "ast.Type.Boolean" then
-        return types.T.Boolean()
-
-    elseif tag == "ast.Type.Integer" then
-        return types.T.Integer()
-
-    elseif tag == "ast.Type.Float" then
-        return types.T.Float()
-
-    elseif tag == "ast.Type.String" then
-        return types.T.String()
-
-    elseif tag == "ast.Type.Any" then
-        return types.T.Any()
-
     elseif tag == "ast.Type.Name" then
         local name = ast_typ.name
         local cname = self.symbol_table:find_symbol(name)
--- a/pallene/typedecl.lua
+++ b/pallene/typedecl.lua
@@ -82,20 +82,12 @@ function typedecl.declare(module, mod_name, type_name, constructors)
     end
 end
 
--- Check if the 2nd tag is a prefix of the first.
--- If appending a "." to the second argument makes it a
--- prefix of the first, captures the rest of the string and
--- returns it
+-- Check if the given type tag belongs to the specified type.
+-- If it does, returns the last component of the tag name.
 --
--- for example:
--- ```
--- typedecl.match_tag("ast.Exp.Bool", "ast.Exp")
--- ```
--- would return "Bool" since `"ast.Exp."` is indeed a prefix of
--- "ast.Exp.bool". Check types.lua for more usage examples.
---
--- @param tag: The type name (string)
--- @param tag_prefix: The prefix to test (string)
+-- Examples:
+--    typedecl.match_tag("ast.Exp.Bool", "ast.Exp")   --> "Bool"
+--    typedecl.match_tag("ast.Stat.While", "ast.Exp") --> false
 function typedecl.match_tag(tag, tag_prefix)
     local n = #tag_prefix
 
@@ -104,9 +96,9 @@ function typedecl.match_tag(tag, tag_prefix)
        string.byte(tag, n + 1) == 46 -- "."
     then
         return string.sub(tag, n + 2)
+    else
+        return false
     end
-
-    return false
 end
 
 -- Throw an error at the given tag.
--- a/spec/typedecl_spec.lua
+++ b/spec/typedecl_spec.lua
@@ -9,18 +9,21 @@ describe("Typedecl", function()
         end, "tag name 'TESTTYPE.Foo.Bar' is already being used")
     end)
 
-    describe("'match_tag'", function ()
-        it("doesn't treat a '.' in the prefix string as regex", function ()
-            assert.falsy(typedecl.match_tag("foo.Bar.baz", "f.o.Bar"))
+    describe("match_tag", function ()
+        it("returns the tag name", function ()
+            assert.equals("baz", typedecl.match_tag("foo.Bar.baz", "foo.Bar"))
         end)
 
-        it("works as expected with input strings", function ()
-            assert.are.equals(typedecl.match_tag("foo.Bar.baz", "foo.Bar"), "baz")
+        it("doesn't crash with a non-string tag", function()
+            assert.equals(false, typedecl.match_tag(nil, "types.T"))
+        end)
+
+        it("doesn't treat a '.' in the prefix string as regex", function ()
+            assert.equals(false, typedecl.match_tag("foo.Bar.baz", "f.o.Bar"))
         end)
 
         it("doesn't require a '.' at the end of prefix.", function ()
-            assert.falsy(typedecl.match_tag("types.T.Float", "types.T."))
+            assert.equals(false, typedecl.match_tag("types.T.Float", "types.T."))
         end)
     end)
-
 end)


spec/parser_spec.lua
```diff
@@ -579,6 +579,31 @@ describe("Pallene parser", function()
...
+                { _tag = "ast.Decl.Decl", name = "v", type = false }, },
+              exps =  {
+                { _tag = "ast.Exp.CallFunc", exp  = { _tag = "ast.Exp.Var",
+                  var = { _tag = "ast.Var.Name", name = "ipairs" } },
+                  args = {
+                    { _tag = "ast.Exp.Var" ,
+                    var = { name = "t" } } } } }
+            }
+        })
+    end)
```
hugomg, line 605: I like that you added a test case for this. People often forget to add tests :)

I'd suggest adding another test for when there is only one variable. That is a code path that we are not testing.

    for x in foo()

It would also be nice if we had some test cases for errors, to ensure that it does something sensible there as well. Those can go with those `assert_statements_syntax_error` near the bottom, next to the syntax errors for the ForNum loops. 

    for in ipairs() do -- missing the declarations
    for i x in ipairs do -- missing a comma

By the way, it is better when the error message is similar to Lua but it is different then it is not a big deal.

hugomg (commented):
Looking great so far! What part are you thinking of doing next?

You can answer here on this PR. I think it might be a better place to continue the discussion.

srijan-paul:
> What part are you thinking of doing next?

What do you think about tackling this bottom-up, from the parser to the coder? That way the next natural place to would be `checker.lua`, right ?
I'll share what I think I know so far.
So to type check `ForIn` loop that looks like so: 
```lua
for <declList> in <expList> do
  --code
end
```

We need to ensure the following:

1. The first item in `expList` is a function, that takes `#expList - 1` arguments. 
2. The number of items in `declList` is equal to `#expList - 1`
3. Finally, we define the symbols in the loop's body  by assigning the corresponding types from the function's 
signature and then the rest can be handled by the existing type checking code present in `checker.lua`.

spec/parser_spec.lua
srijan-paul: Ah right, I'll add those.

hugomg:
Starting with the checker sounds like a good idea.

> The first item in expList is a function, that takes #expList - 1 arguments.

IIRC, the explist has to return 3 things: the iterator function, the initial state, and the initial value of the control variable. We can do this using the expand_function_returns method. It converts an explist with a multi-return function into an explist that makes those extra returns explicit:

    before: ipairs(xs)
    after:  ipairs(xs), ExtraRet(2), ExtraRet(3)

After the expansion to add the required ExtraRets, the explist should contain 3 things:

1. The iterator function. Receives 2 arguments of type `any` and returns any number of results of type `any`.
2. The state, of type `any`
3. The initial value, of type  `any`. 

One catch is that we may need to modify the expand_function_returns method so that instead of receiving the `lhs` list as an argument, it receives the length of the list. In the ForIn loop the "lhs" is the list of three things (iterator, state, value), which is not the same thing as the list of variables (k, v). Let me know if this part is still confusing.

> The number of items in declList is equal to the return value count of the function.

Yes.

> Finally, we define the symbols in the loop's body by assigning the corresponding types from the function's
signature and then the rest can be handled by the existing type checking code present in checker.lua.

Yes. This part should work similarly to the ForNum case.

srijan-paul:
Right, I'll read up on the details and update you. 
Should I request a review after I add the tests, or after I'm done with that *and* have made some progress with the checker ?

hugomg:
I think it would make sense to have a commit to add new tests that I suggested and then a later commit to add the checker stuff. But you can do whatever works for you.

srijan-paul:
> In the ForIn loop the "lhs" is the list of three things (iterator, state, value), which is not the same thing as the list of variables (k, v). Let me know if this part is still confusing.

Yeah, sorry I didn't quite get that part. 
isn't `lhs` an array of `ast.Decl.Decl` nodes ?

hugomg:
> isn't lhs an array of ast.Decl.Decl nodes ?

Yes. However, I think we might want to change that.

----------------

Currently, we use the `expand_function_returns` method to implement multiple assignments,. It looks like this:

```
x, y, z = f()                            -- before
x, y, z = f(), ExtraRet(2), ExtraRet(3)  -- after
```

To do the expansion we need to know the size of the LHS. Currently, the `expand_function_returns` receives the lhs array and then calls `#` to get it's length.

However, this isn't going to work for the for-in loops because we always want to expand to 3 values (the iterator, the state, and the initial value), no matter how many variables the programmer is declaring. For example, in ipairs we typically declare 2 variables but we still want to expand the explist to three values.

```
for i, x in ipairs() do                           -- before
for i, x in ipairs(), ExtraRet(2), ExtraRet(3) do -- after
```

One fix for this would be to change the `expand_function_returns` method so that instead of receiving the `lhs` array it receives an integer. All the current call sites would become `expand_function_returns(#lhs, rhs)` and the for-in case would be `expand_function_returns(3, explist)`.

hugomg:
> However, I understood that part, what I don't understand is, shouldn't the "lhs" just be a number then?

Exactly. I think it really should have been just a number all along.
This PR would be a good opportunity to fix that mistake :)

srijan-paul:
Ah, understood !

Oh, and one more question if you don't mind, about this: 
> The iterator function. Receives 2 arguments of type `any` and returns any number of results of type `any`

Should it not take two arguments having the same type as the `ExtraRet(2)` and `ExtraRet(3)`? I was thinking about resolving the types in the following order:

1. First, expand the `expList` using `expand_function_returns`. 
2. Infer the types of `ExtraRet(2)` and `ExtraRet(3)` using `check_exp_sysnthesize`. 
3. Check that the the inferred types are assignable as arguments to the iterator function. Maybe something like 
 
```lua
local iteratorfn  = self:check_exp_synthesize(stat.exp[1])
local f_type = iteratorfn._type

if f_type._tag != "types.T.Function" then
  type_error(iteratorfn.loc, "expected an iterator function")
end
-- perform other checks similar to ast.Exp.CallFunc 

for i = 2, #stat.exp do
  stat.exp[i] = 
  self:check_exp_verify(stat.exp[i], f_type.arg_types[1],
       "%d to function call", i) 
end
```

4. match the return count of the function with the declaration count of the loop.
5. Finally, add the declaration names to the symbol table.

Also, what do you think about a helper function that matches an `expList` against a `typeList`?
Similar to `exp_check_verify` but for multiple values.
Maybe this could also help with matching call arguments in the type inference in `exp_check_synthesize` ?

hugomg:
Just to be clear, maybe we should start using a consistent set of names for the stuff that the for loop is manipulating:

1. The `iterator` function
2. The `state` value
3. The `control` variable

> Should it not take two arguments having the same type as the ExtraRet(2) and ExtraRet(3)? I was thinking about resolving the types in the following order:

I think that the `control` argument needs to be `any` no matter what, because it has to be able to store `nil`. For example, in `ipairs` the control argument is either `nil` or `integer`. The only type that can do that is `any`.

However, for the iterator and state arguments that is an excellent question. In theory we could allow the state argument to have a more precise type, as you suggested. However, I would imagine that just having the state be `any` might be a simpler thing to do for the initial implementation of this feature. What is the main motivation you have for allowing more precise types for the state argument? Is this more about having "better types" or more about performance?

> Also, what do you think about a helper function that matches an expList against a typeList?

In the multiple-assignment case we use `check_initializer_exp` which cleverly deals with both of these cases:

```
local x   = e -- type of 'x' is inferred from 'e'
local x:T = e -- type of 'e' is verified to be 'T'. Possibly inserting a type cast.
```

However, we might not be able to use this function in the for-in loops, because there isn't an expression acting as the "initializer". I honestly don't know if a helper function would help or not. Would that proposed helper function be used just for the for-in loop or for other places as well?

srijan-paul:
> What is the main motivation you have for allowing more precise types for the state argument?

Just type soundness really, but yeah as you said I think we could just go with `any` for the initial implementation. 

>  Would that proposed helper function be used just for the for-in loop or for other places as well?

It could be used in `check_exp_synthesize` for example, to match the call arguments against the function's annotated types.
My idea was to separate this: 

```lua
if #f_type.arg_types ~= #exp.args then
        type_error(exp.loc,
            "function expects %d argument(s) but received %d",
            #f_type.arg_types, #exp.args)
end

for i = 1, #exp.args 
    exp.args[i] =
            self:check_exp_verify(
                exp.args[i], f_type.arg_types[i],
                "argument %d of call to function", i)
end
```

into to it's own little function, say `check_call_arguments` or something similar, so we could use this again in `ipairs`. What do you think ?

hugomg:
> My idea was to separate this into to it's own little function

The  `ipairs` function call should be an `ast.Exp.CallFunc` node. If that's the case, I think we should be able to just call `check_exp_synthesize` on it, without needing to create a separate helper function.

srijan-paul:
Right, understood.

srijan-paul:
I made minor changes to accommodate for the `for-in` loops in `checker.lua`, 
I'm not sure if I'm checking the iterator function's type correctly. Should I really be checking if the state variable is passed in the way it's being done now?

I've worked with type systems before but Pallene doesn't seem to use an explicit type graph, so let me know if I messed up somewhere.

pallene/checker.lua
```diff
@@ -451,9 +452,42 @@ function Checker:check_stat(stat)
...
+        else
+            rhs[2] = self:check_exp_verify(rhs[2], types.T.Any(), "state variable")
+        end
+
+        rhs[3] = rhs[3] and self:check_exp_verify(rhs[3], types.T.Any(), "control variable")
+
+        if #stat.decls ~= #iteratorfn._type.ret_types then
+            type_error(stat.decls[1].loc, "expected %d values, but function returns %d",
+                       #stat.decls, #iteratorfn._type.ret_types)
+        end
```
hugomg, line 479: One thing to watch out here if we set `expected_ret_types` to an array of `any` and then call `check_exp_verify`, then it will insert a type coercion and the return type of `iteratorfn` will literally be an array of `any` as well. Maybe that is a sign that we shouldn't be using `check_exp_verify` up there.

pallene/checker.lua
```diff
@@ -451,9 +452,42 @@ function Checker:check_stat(stat)
...
+
+        self.symbol_table:with_block(function()
+            for i = 1, #stat.decls do
+                stat.decls[i]._type = iteratorfn._type.ret_types[i]
+                self:add_local(stat.decls[i])
```
hugomg, line 484: We should also handle the case where the programmer inserted a type annotation. That is, when `stat.decls[i].type` exists. (See check_initializer_exp). Perhaps use `types.consistent` to see if the type returned to the iterator can be assigned to the declared type?

hugomg (commented):
I think one thing that could help clarify all of this is if we had some test cases for well-typed programs. These would normally go in the coder.lua. Of course, those test cases would fail because the code generation isn't implemented yet. However, if they fail because they are crashing in the code generator, then you know that the type checker is working ;)

Could you add a couple of those tests? For now you can leave them commented out or skip them using busted's `pending()` feature.

hugomg:
Also, does your question about how to type check the state argument have to do with that comment I made about check_exp_verify inserting type coercions?

pallene/checker.lua
```diff
@@ -451,9 +452,42 @@ function Checker:check_stat(stat)
...
+        end
+
+        local iteratorfn = self:check_exp_verify(rhs[1], types.T.Function({
+            types.T.Any(), types.T.Any()
+        }, expected_ret_types), "loop iterator")
+
+        if not rhs[2] then
+            type_error(stat.loc, "Missing state argument in for-in loop")
+        else
+            rhs[2] = self:check_exp_verify(rhs[2], types.T.Any(), "state variable")
```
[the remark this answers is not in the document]
srijan-paul: Ah yes, I totally didn't take that into account. Will fix that.

pallene/checker.lua
srijan-paul: Ah yes, so what do you suggest I do? Inline some code there that does the type checking or use another function?

srijan-paul:
> Could you add a couple of those tests?

Right, I'll add those.

> Also, does your question about how to type check the state argument have to do with that comment I made about check_exp_verify inserting type coercions?

For now since we're not type checking the state argument against the iterator's actual parameter, I think not so much. I'll get back to you after making the requested changes.

hugomg:
I just noticed that you are using merge commits when you're making changes. Do you know if it would be possible to rebase this branch to have a linear history, without any merges?

Captura de tela de 2020-11-02 10-42-10

pallene/checker.lua
```diff
@@ -451,9 +452,70 @@ function Checker:check_stat(stat)
...
+                types.tostring(itertype), types.tostring(iteratorfn._type))
+        end
+
+        rhs[2] = self:check_exp_synthesize(rhs[2])
+        rhs[3] = self:check_exp_synthesize(rhs[3])
+
+        if rhs[2]._type._tag ~= "types.T.Any" or
+           rhs[3]._type._tag  ~= "types.T.Any" then
+            type_error(rhs[2].loc, "expected state and control values to have type any")
+        end
```
hugomg, line 492: Could we have separate type checks for the state and for the control variable? That way the error message will inform which of them was the one that had a problem. It would also be nice if the error message also said the actual type as well.

srijan-paul:
So I made some fixes (excuse the repeated commits :p).
One question I had about this:
> But in this case, we actually do want to check if the type is exactly any

So let's consider this stupid but simple Lua code snippet:

```lua
function foo(a, b)
    return b, a
end

for x, y in foo, 2, 3 do
    print(x, y)
end
```

This would print `3, 2` in an infinite loop. Nevertheless, the program is syntactically, and semantically valid.
So let's assume the Pallene equivalent of the snippet above:

```lua
export function foo(a: any, b: any): (any, any)
    return b, a
end

export function main()
    for x, y in foo, 2, 3 do
        print(x, y)
    end
end
```

Now, since we require that the state and control arguments be exactly of type "any", this wouldn't compile, and the user would have to put some redundant casts like `2 as any` and `3 as any`. So I was wondering if that's an intended effect?

Oh and the ir/codegen tests, I'll be adding those in a minute.

srijan-paul:
> Do you know if it would be possible to rebase this branch to have a linear history, without any merges?

Oops, I'll rebase my branch such that all my commits are always on top of the latest commit on upstream/main.

pallene/checker.lua
srijan-paul: True, I'll change it to something like `expected <type> but found <type> in for-in state value`.

hugomg:
> Now, since we require that the state and control arguments be exactly of type "any", this wouldn't compile, and the user would have to put some redundant casts like 2 as any and 3 as any. So I was wondering if that's an intended effect?

Well spotted. If we use `_synthesize` instead of `_verify` there then it will have this effect of requiring the programmer to add an explicit type connversion. I agree that it would be unintuitive given how the rest of Pallene works. However, 99% of the time the programmer will actually write `for x in blah()` instead of explicitly passing the initial value and control, so I'm not sure if it is a real problem in practice.

Here is what I would suggest: let's leave this problem aside until after we finish implementing the code generator. 
I'm worried that the automatic type coercions might silence type errors (because anything can be converted to `any`) or that they might not interact well with the logic that we will be using to optimize the special case of the `ipairs` iterator. I think we will be better able to determine if these are real problems once we finish implementing the code generation for the simple version without coercions, with a working test-suite in place. If it turns out that there isn't actually a problem with the coercions then we can go back and add them.

srijan-paul:
To clean up the merge commits and lock my forked master to the upstream master, I'll be changing the branch of this pull request. This is a mistake that I shouldn't have made, It may cause some commits to disappear but the changes will stay untouched

srijan-paul:
I made a huge mistake by not locking origin/master to upstream, apologies on that. Is it okay if I create a new PR, that is pulling in from a branch other than `origin/master`?

hugomg:
Should the --force-with-lease be enough to clean the history without having to reopen the PR?

srijan-paul:
The intent was to lock origin/master with upstream/master, my local branch had a different name which caused a bit of confusion :/

Issue #124, opened 2019-11-21
Implement maps based on Lua tables


gligneul:
AKA associative array. The proposed syntax is `{T1 -> T2}`. Like arrays, indexing an entry that yields nil might result in a runtime type error.

hugomg:
The proposed syntax is ambiguous with arrays of functions.

We could use `{A=>B}` instead. Or `{A:B}`, like Typed Lua

gligneul:
Indeed. I would use `{A=>B}` to make clear it is not a table.

gligneul:
Anyone who implement this could use the PR #139 as inspiration.

PR #334, opened 2020-11-03
Implement for-in loops and ipairs

srijan-paul:
This is  PR is supposed to address issue #164.
The following changes have been made:

1. The ast now has separate `ast.Stat.ForNum` nodes for numeric for loops and `ast.Stat.ForIn` nodes for `for..in..` loops. Tests added to reflect the same.
2. The type checker is able to type check `ForIn` ast nodes.Test cases in `spec/checker_spec.lua` for the same as well.

(This PR is a re-open of #332)


srijan-paul:
I'll squash the smaller commits once/if the PR is ready to be merged if that's okay.

srijan-paul:
Do you think the checker is at a point where we could start working on the IR/codegen?

hugomg:
Did you manage to test if the type checker is doing something sensible for programs that don't have type errors? If yes, then I think we can proceed to the IR/codegen. Feel free to ask any questions about that part, by the way.

srijan-paul:
> Did you manage to test if the type checker is doing something sensible for programs that don't have type errors?

Would that be achieved by running a test case through the checker and then asserting for truthiness? 
Since right now there's only `assert_error` in `checker_spec.lua`.

hugomg:
We tend to prefer to test the non-error behavior in the coder_spec, because it tests both the type checker and the code generator at the same time. 

What I would do here is to go to the coder_spec and write test cases for the for-in loop. They'll obviously fail but the error message will tell you what is going on. If they fail at the type checking stage, then there is a problem with the type checker or with the typing rules that we created. Alternatively, if it crashes because the code generation is not yet implemented then it means that the type checking was sucessful. After you find out which one of these it was, you can comment out the test cases so the test suite passes again :)

srijan-paul:
Oh, makes sense.
Right, I'll do that then.

srijan-paul:
Quick question, should I be adding these positive tests to `execution_tests.lua`, or directly in the `coder_spec.lua`?

Also, will those tests need to exist as separate files or inlined into the file as strings?

hugomg:
You'll want to put the test cases in the `execution_tests.lua`, as an inlined string.

Sorry for the confusion. We recently refactored the coder_spec and the actual test cases now reside in the `execution_tests` file. But I forgot to tell you about it 😅.

If you're curious why it's this way, the reason is that we recently added a Lua backend to the compiler. We wanted to be able to test that backend using the same tests as the C backend. These shared tests now live in the `execution_tests` file.

srijan-paul:
Done.
So I added the test cases to `execution_tests.lua`, and as anticipated an error was thrown by `to_ir.lua`.
image

For now I only added two pending test cases, could add the rest after we have an expected code flow.
Once this is done, the next natural modules to fall in place are `ir.lua` and `to_ir.lua`, correct?

I had one idea in to translate to the IR but then I don't know if it would fit well with the current code generator.  Do you have any preferred idea in mind about what the IR/gen would look like?

hugomg:
> So I added the test cases to execution_tests.lua, and as anticipated an error was thrown by to_ir.lua.

Nice! Looks good so far. 

Do you know if the type checker is also working for a straightforward `for i, x in ipairs(xs)`? I suspect that we might still need to add the `ipairs` function to the builtin function list on  `builtins.lua`.

> Once this is done, the next natural modules to fall in place are ir.lua and to_ir.lua, correct?

Yup. 

> I had one idea in mind to translate to the IR but then I don't know if it would fit well with the current code generator. 

What's the idea that you had?

>  Do you have any preferred idea in mind about what the IR/gen would look like?

The main thing I would say is that we probably want to treat ipairs specially. Have one instruction in the IR for generic for-in loops, and another instruction for `ipairs` loops.  That way, when we generate code for the ipairs loop we can turn it into a plain old for loop, without any iterator functions.

srijan-paul:
> I suspect that we might still need to add the ipairs function to the builtin function list on builtins.lua.

You're right, ipairs is not listed as a builtin yet, so we'll need to do that first.

> What's the idea that you had?
I think to use Lua's actual `ipairs` function, one would need to emit `GETTABUP` bytecode with `__ENV  "ipairs"` as it's operand, to avoid that we could use a custom desugared version of `ipairs` like I explain below.

Now If the user provides his own  iterate, state and control values (uncommon, but still valid Pallene code building upon what we did in the frontend)  we could "desugar" it a bit.

We can follow from [the official reference manual](http://www.lua.org/manual/5.2/manual.html#3.3.5).
So assuming the following user written code:
```lua
function iter(state, current)
	if current == #state then
		return nil
	end

	return current + 1, state[current + 1]

end

function xpairs(t)
	return iter, t, 0
end
```
A for in loop that looks like the following:

```lua
for i, x in xpairs(xs) do
	print(i, x)
end
```

would result in the same c code as if the user wrote this instead:

```lua
do
       local iter, st, ctrl = xpairs(xs)
       while true do
         local i, x = iter(st, ctlr)
         if i == nil then break end
         ctrl = i
         print(i, x)
       end
end
```
According to the reference manual, both have the exact same effect.
We could either warrant a `ForIn` IR type and handle the whole while loop in `coder.lua` or use the same IR that Pallene has currently and compile a `ast.Stat.ForIn` to an IR that would resemble a while loop like the one above, that way `coder.lua` requires fewer changes.

All that said, the fact still remains that users will 99% of the time use `for i, x in foo(xs)` instead of an explicit expression list. I personally use a custom generator for my game's ECS (currently written in teal, a similar typed dialect of lua) , a part of which I was looking to port over to Pallene once it has Union types and `for-in`. :sweat_smile: 

But this is quite roundabout so If you think there is an easier way then we should definitely do that instead. 

>  That way, when we generate code for the ipairs loop we can turn it into a plain old for loop, without any iterator functions.

What I pitched is kind of similar but I also want to know what kind of for loop we'd be emitting here.

hugomg:
> I think to use Lua's actual ipairs function, 

For now I think we can leave the actual ipairs function unimplemented. Most of the time the programmer will use `for i, x in ipairs(xs)` , which we will compile into a loop that doesn't actually call ipairs.

That said, this reminds me that we probably should implement the actual functions for the builtins at some point. I opened a separate issue for that. #337

> one would need to emit GETTABUP bytecode with __ENV "ipairs" as it's operand

Pallene doesn't work with Lua bytecodes. It's better if you think from the point of view of a C extension module, which interacts with Lua stuff via the C API. But the basic idea is what you said. To call the real ipairs function from Pallene we would need to grab it from the ENV and save it to an upvalue. See #337 for details.

> We could either warrant a ForIn IR type and handle the whole conversion to while loop in coder.lua or use the same IR that Pallene already has and compile a ast.Stat.ForIn to a representation that would resemble a while loop like the one above, that way coder.lua requires fewer changes.

I love this idea of using desugaring to avoid making changes to the IR. 

In the non-ipairs case we can use the iterator loop similar to the one in the Lua manual, as you just described. However, in the special case of `for i, x in ipairs(xs)` I think we should desugar it into something that doesn't call any iterator functions. Avoiding calling the iterator function is already faster in plain Lua code and I suspect that it will make even more of a difference in Pallene.

```
for i, x in ipairs(xs) do
    body()
end
```

becomes

```
local _xs = xs
local _i = 1
while true do
    local x = _xs[_i]
    if x == nil then break
    local i = _i
    -- ...
    _i = _i + 1
```

srijan-paul:
> I think we should desugar it into something that doesn't call any iterator functions. 

Ah yes, I didn't think of that one. `ipairs` can simply be a plain loop.
So I guess I'll be adding `for..in` loop support to `to_ir.lua`, leaving `coder.lua` and `ir.lua` untouched like we discussed?

hugomg:
Exactly

Em 9 de novembro de 2020 02:37:48 BRT, srijan_ <notifications@github.com> escreveu:

srijan-paul:
Hi,  I've run into a couple of problems  trying to solve this.
To "hand-emit" the IR, I first added this small helper:

```lua
-- Adds a new local variable and assings @exp to it.
-- returns the numeric key of the local.
function ToIR:init_local(cmds, decl, exp)
    local v = ir.add_local(self.func, decl.name, decl._type)
    self.loc_id_of_decl[decl] = v
    self:exp_to_assignment(cmds, v, exp)
    return v
end
```

then in the `ToIR:convert_stat` method, 

```lua
 elseif tag == "ast.Stat.ForIn" then
        local decls = stat.decls
        local exps = stat.exps
        
        local v_iter = self:init_local(cmds, decls[1], exps[1])
        local v_stat = self:init_local(cmds, decls[2], exps[2])
        local v_ctrl = self:init_local(cmds, decls[3], exps[3])
        
        local body = {}
        local cmpOp = type_specific_binop("==", v_iter._type, "types.T.Nil")
        local cond = ir.Cmd.Binop(<loc>, <dst>, cmpOp, ir.Value.LocalVar(v_iter),
            ir.Value.LocalVar(v_iter))
...
```

the `type_specific_binop` local is declared below this method, in order for it to accessible here we'd have to move type_specific_unop and type_specific_binop above the method. Should I be doing this some other way?

hugomg:
> To "hand-emit" the IR, I first added this small helper:

I like the idea of making a helper function to ensure that we remember to add the decl to the `loc_id_of_decl` when we create a new local variable based on a `ast.Decl`. But maybe it would be better to have the method name refer to the decl?

```
function ToIR:new_local_from_decl(decl)
    local v = ir.add_local(self.func, decl.name, decl._type)
    self.loc_id_of_decl[decl] = v
    return v
end
```

This doesn't initialize the variable at the same time but I think it's OK to just call exp_to_assignment afterwards.

```
local v_iter = self:new_local_from_decl(decls[1])
local v_stat = self:new_local_from_decl(decls[2])
local v_ctrl = self:new_local_from_decl(decls[3])

self:exp_to_assignment(cmds, v_iter, exps[1])
self:exp_to_assignment(cmds, v_stat, exps[2])
self:exp_to_assignment(cmds, v_ctrl, exps[3])
```

> the type_specific_binop local is declared below this method

I think that for this part the type of the `iter` variable has to be `any` and the comparison with nil always needs to be an `AnyEq`. .The `nil` constant in the `== nil` would also needs to be casted to `any` and you may need to create a temporary variable of type `any` in order to do so. We wouldn't need to ask type_specific_binop about it

Alternatively, we could make a small change to the IR, adding a new `ir.Cmd`  called `IsNil`. It would be similar to the IsTruthy instruction except that it would call the the Lua macro `ttisnil` instead of `pallene_is_truthy`. That might be more efficient than using `==`, which calls the Lua function `luaV_equalobj`. 

I understand that we were trying to avoid making changes to the IR but I have a feeling that this `IsNil` instruction will be genuinely useful and that it is likely that we would eventually create something like it at some point.

srijan-paul:
> But maybe it would be better to have the method name refer to the decl?
Good point, I'll do that.

> I think that for this part the type of the iter variable has to be any and the comparison with nil always needs to be an AnyEq. 

Yeah I missed that, thanks.

> lternatively, we could make a small change to the IR, adding a new ir.Cmd called IsNil
This seems like a good change, will do.

srijan-paul:
Hi, I added support for `for-in` loops without ipairs in the IR and codegen stage, along with the `IsNil` cmd. 
Now the `new_local_from_decl` helper is unused but I think one could remove the many `add_local` followed by `exp_to_assignment` with a call to that method.

The pending test cases that were added earlier have been passing so I think that it's working fine so far. 
Once all issues with this commit are resolved I think we'd be good to go for `ipairs`, what do you think?

hugomg:
Nice! I'll review this carefully later. Until then, one thing you could do is clean up the branch history with `rebase -i` to get rid of the merge commits and other redundant commits.

hugomg:
> In the codegen stage I named the iter, st and ctrl local variables with a prefixed $ to avoid shadowing user defined variables

The variable names that we use in `add_local` are only to help debugging. In the generated C code, they appear inside a comment next to the variable declaration. Those names don't have anything to do with shadowing. At this point, the name that actually identifies the variable is its integer ID. This is also why in the generated code the variable names are x1, x2, x3, and so on.

Without thinking too much into it, my guess is that the $iter, $st, and $ctrl for the comments should be fine. But I think it would be nice if you could have a look at it yourself. Write a small test program, emit the C code using `./pallenec --emit-c`, and try to see if the result is somewhat readable.

srijan-paul:
>  and try to see if the result is somewhat readable.

I tried a few test programs, and comparing them to the other Pallene generated code, I personally think that it looks totally fine, but I'm not sure if it's possible to tell by looking at the generated C code that it was originally a `for-in` loop, but other than that it looks okay, from my judgement. I think you would be a better judge for that, should I post some snippets of the test programs?

hugomg:
I can take a look at it myself while I am reviewing the code. That said, having some more interesting examples might help. Could you please send some examples of small programs using for-in loops that don't just go into an infinite loop?

The obvious ones that I would write myself all use `ipairs` but that one isn't implemented yet. :)

pallene/to_ir.lua
```diff
@@ -205,6 +205,46 @@ function ToIR:convert_stat(cmds, stat)
...
+
+        local xs = { ir.Value.LocalVar(v_state), ir.Value.LocalVar(v_ctrl) }
+        local iter_f_id = self.fun_id_of_decl[exps[1].var._name.decl]
+        table.insert(body, ir.Cmd.CallStatic(exps[1].loc, itertype, v_decls, iter_f_id, xs))
+
+        -- if i == nil then break
+        local v_cond = ir.add_local(self.func, "$cond", types.T.Boolean())
+        local then_ = {}; table.insert(then_, ir.Cmd.Break())
+        table.insert(body, ir.Cmd.IsNil(stat.loc, v_cond, ir.Value.LocalVar(v_decls[1])))
+        table.insert(body, ir.Cmd.If(stat.loc, ir.Value.LocalVar(v_cond), ir.Cmd.Seq(then_), ir.Cmd.Seq({})))
```
hugomg, line 241: I think we can use `ir.Cmd.If(stat.loc, ir.Value.LocalVar(v_cond), ir.Cmd.Break(), ir.Cmd.Nop())`, without those Seqs.

pallene/to_ir.lua
```diff
@@ -205,6 +205,46 @@ function ToIR:convert_stat(cmds, stat)
...
+        local v_decls = {}
+        for _, decl in ipairs(decls) do
+            local v = ir.add_local(self.func, decl.name, decl._type)
+            self.loc_id_of_decl[decl] = v
+            table.insert(v_decls, v)
+        end
+
+        local xs = { ir.Value.LocalVar(v_state), ir.Value.LocalVar(v_ctrl) }
+        local iter_f_id = self.fun_id_of_decl[exps[1].var._name.decl]
+        table.insert(body, ir.Cmd.CallStatic(exps[1].loc, itertype, v_decls, iter_f_id, xs))
```
hugomg, line 235: This code is assuming that `exps[1]` is an `ast.Exp.Var`, which isn't always true! That will work for something like `for x in a, b, c` but won't work for something like `for i, x in ipairs()`.

Furthermore, we should be using CallDyn here instead of CallStatic, because the iterator function isn't necessarily known at compile time.

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,42 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
         end)
     end)
 
+    describe("For-in loops", function()
```
hugomg, line 1752: Currently these tests cases are testing if the code is compiling successfully but are not testing that they produce the expected result when you run the code. We need to add an `it` block to do that.

Furthermore, I think it would be nice if we had a more intuitive example. One possibility would be to create a custom iterator function called `rev_ipairs` that works similarly to ipairs, except that it iterates in decreasing order instead of increasing order.

BTW, while I like the idea of testing loops of the form `for i, x in a, b, c`, I think we should also test the more common pattern where the RHS is a function call like `for i, x in rev_ipairs(xs)`.

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,42 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+                for i, j, k in iter, a, b do
+                    i = j
+                    k = i
+                end
+            end
+        ]])
+    end)
+
+    describe("Nested for-in loops", function()
+        compile([[
```
hugomg, line 1770: To speed up the test suite, we put related code in the same `compile` block. I'd put the test case for nested for-in loops in the same compile block as the other non-nested loop.

hugomg (changes requested):
This looks like a great step in the direction of implementing generic for-in loops. The main problem that we still have to fix is to make it work for loops that use function calls, such as `for i, x in rev_ipairs(xs)`. This will require both adding new test cases as well as making corrections to the code generation.

Once we get the generic for-in loop working the only thing left to do will be to add support for `ipairs`.

srijan-paul:
I made the changes, now it's possible to evaluate loops like `k, v in my_xpairs(xs)`. There are still only two tests, the first one has some assertions to test the output. I feel that adding more tests would make sense after adding in support for the `ipairs` builtin, what would you say?

Speaking of the C code, would a couple of pastebin links suffice for the Pallene <-> C comparisons or should I post screenshots instead?

hugomg:
Thanks. I'll look at it closely later.

Don't worry about the C code. Now that we have some examples I can look at it myself.

Em 19 de novembro de 2020 10:12:50 BRT, srijan_ <notifications@github.com> escreveu:

srijan-paul:
I just noticed I squashed the commits a bit farther back than I should have into an older commit that you had made, I'll separate those.

hugomg:
I'm a bit confused. Am I supposed to look at the `srijan/for-in-loops` or the `srijan/ipairs` branch?

The for-in-loops  branch isn't rebased on top of master, and averything seems to have been squashed into a single commit that jut says "Shorten the match_tag comment".

srijan-paul:
Oops. Yeah, it's the `/for-in-loops` branch.
I squashed all the older commits into one since they won't be necessary. Should I not have done that?  😓

hugomg:
I think it's easier to figure out what is going on when a set of changes is split into clearly defined commits with good commit titles and commit descriptions. I like having more than one commit for each PR (as you can see in the git history for Pallene) but I know some other projects like squashing everything in a single big commit. But either way, the current state of a single large commit with a commit message that is just a list of what the old commit titles were isn't much better than the old status quo.

One thing that can help when you have to do a big history rewrite like this is to make a backup copy of the repository so you can do a https://xkcd.com/1597/ and start over if necessary. Anyway, if you don't have any way to undo it I can still try to figure it out so don't loose sleep over it :)

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,71 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+                end
+
+                return current as integer + 1, (state as {any})[current as integer + 1]
+            end
+
+            typealias iterfn = (any, any) -> (any, any)
+            local function xpairs(xs: {any}): (iterfn, any, any)
+                return iter, xs, 0
+            end
+
```
hugomg, line 1766: I think we can improve this a bit by taking more advantage of implicit casting and by renaming the function to make it clear that it is reimplementing ipairs. But still, thanks a lot for the example you gave. It's easier to suggest an improvement than to come up with it from scratch ;)

```
local function iter(arr: {any}, prev: integer): (any, any)
    local i = prev + 1
    local x = arr[i]
    if x == (nil as any) then
        return nil, nil
    else
        return i, x
    end
end

typealias iterfn = (any, any) -> (any, any)
export function my_ipairs(xs: {any}): (iterfn, any, any)
    return iter, xs, 0
end
```

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,71 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+        it("", function()
+            run_test([[
+                local xs = test.double({1, 2})
+                assert(xs[1] == 2 and xs[2] == 4)
+
+                xs = {0.0, 2.1, -1.0, -4.5}
+                test.abs_list(xs)
+                assert(xs[1] == 0.0 and xs[2] == 2.1 and xs[3] == 1.0 and xs[4] == 4.5)
+            ]])
+        end)
```
hugomg, line 1795: Although we only have one `compile` block for everything, we prefer having a separate  `it` block for each test case. That way, if a test case fails it is easier to know which one it was.

What is the difference between the test for `double` and the test for `abs_list`? If there is an important difference, that should be reflected in the name of the test that we pass to the `it` function. If there isn't an important difference, then I think we can get rid of the `abs_list` test and just keep the `double` one.

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,71 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+            typealias iterfn = (any, any) -> (any, any)
+            local function xpairs(xs: {any}): (iterfn, any, any)
+                return iter, xs, 0
+            end
+
+
+            export function double(xs: {integer}): {integer}
+                local out: {integer} = {}
+                for i, x in xpairs(xs) do
+                    out[#out + 1] = x as integer * 2
```
hugomg, line 1771: Can we use `out[i] = ` here?

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,71 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+                local a: any = 2
+                local b: any = 3
+                for i, j in iter, a, b do
+                    for k, v in iter, b, a do
+                        i = v
+                        k = j
+                    end
+                end
+            end
+        ]])
```
hugomg, line 1814: This test case is missing an `it` block.  My suggestion would be to rewrite this test case to use that xpairs/my_ipairs function from before, because it's easier to think of an example if we use that. We can put it next to the other `my_ipairs` test cases.

spec/execution_tests.lua
```diff
@@ -1749,6 +1749,71 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+
+
+            export function double(xs: {integer}): {integer}
+                local out: {integer} = {}
+                for i, x in xpairs(xs) do
+                    out[#out + 1] = x as integer * 2
+                end
+
+                return out
+            end
```
hugomg, line 1775: Currently we are testing the non-function-call RHS at the same time as we test nested for-in loops. IMO, it would be better to have that as a separate test case that tests just the non-function-call case.

```
export function double_with_weird_rhs(xs: {any}): {integer}
    local out: {integer} = {}
    for i, x in iter as iterfn, xs as any, 0 as any do
        out[i] = x as integer * 2
    end
    return out
end
```

hugomg (changes requested):
Excellent work so far. We are getting close to completing the implementation of generic for-in loops. In addition to the comments I made for the code, these are the final steps that I think we are missing:

1. Please rebase this branch on top of `master`. Currently it this branch is branching from an older commit.
2. It's OK to keep all the commits squashed as it is now. But if possible, could you rewrite the  commit message and the commit description to give a nice explanation that this commit is implementing generic for-in loops? 
3. We will need to add some documentation in the `manual.md`. The only tricky thing is that we ask to put each sentence on a separate line. That helps keep the diffs easy to read.

After that, it would be nice if we could optimize the `ipairs`builtin. But we should probably finish the generic for-in loops first and then decide if we continue the ipairs work in this PR or if we open a separate PR for the ipairs stuff.

ps.: The generated C code looks great.

srijan-paul:
>  But if possible, could you rewrite the commit message and the commit description to give a nice explanation that this commit is implementing generic for-in loops?

Does that mean I should edit the latest commit message and make it more descriptive?

hugomg:
Yes

Em 23 de novembro de 2020 10:32:04 BRT, srijan_ <notifications@github.com> escreveu:

pallene/to_ir.lua
```diff
@@ -205,6 +205,58 @@ function ToIR:convert_stat(cmds, stat)
...
+        local v_state = ir.add_local(self.func, "$st", exps[2]._type)
+        self:exp_to_assignment(cmds, v_state, exps[2])
+        local v_ctrl = ir.add_local(self.func, "$ctrl", exps[3]._type)
+        self:exp_to_assignment(cmds, v_ctrl, exps[3])
+
+        --Body of the `while` loop.
+        local body = {}
+
+        local itertype = exps[1]._type
+        local v_decls = {}
```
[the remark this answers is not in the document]
srijan-paul: I'll name it `decl_ids` then.

hugomg:
> I'll name it decl_ids then.

I think it would be better to not use `decl` in the name. What about
`v_lhs`?

srijan-paul:
Hi hugo, apart from the changes requested I was running some more tests and I found a bug that I've been trying to fix.  It happens when the left hand side of a loop has type annotations. 

```lua
local sum = 0
for i, x: integer in my_ipairs(xs) do
    sum = sum + x
end
```
Running this through the code generator, I see that `x` is declared as a `lua_Integer` , and as such when we apply the return value of the iterator function to x, we are assigning a boxed Lua value to a `lua_Integer`. 

```c
lua_Integer x7; /* x */
// ...
lua_call(L, 2, 2);
        {
            L->top--;
            TValue *slot = s2v(L->top);
            x7 = *(slot);
        }
```
I'm not sure what the best way to fix this would be. 
From what I can tell, this is happening because I used `types.consistent` in `checker.lua` to type check the lhs decls against the rhs.

```lua
if decl.type then
    decl._type = self:from_ast_type(decl.type)
    if not types.consistent(decl._type, ret_types[i]) then
        type_error(decl.loc, "expected value of type %s, but iterator returns %s",
                    types.tostring(decl._type), types.tostring(ret_types[i]))
    end
else
```

`types.consistent` returns true when any of the two types are `any` .
Since we require that the iterator function must return values that are *exactly* any, I'm not sure if it annotating the lhs is desirable, since we would be assigning a boxed lua value to an integer.

hugomg:
Allowing the programmer to use type annotations to automatically downcast things from any sounds like it would be a convenient feature. I think we can modify the desugaring to add a downcast from any to the declared type. For example, if we start with the following code:

```
function f(xs: {float})
    for i:integer, x:float in my_ipairs(xs) do
        <loop body>
    end
end
```

it could be compiled into something like:

```
function f(xs: {float})
    local iter, st, ctrl = my_ipairs(xs)
    while true do
      local i_any, x_any = iter(st, ctrl)
      if i_any == nil then break end
      i = i_any as integer
      x = x_any as float
      <loop body>
    end
end
```

There are two ways that we could do this: 
1. Keep the desugaring in the to_ir.lua as it is right now. If a loop variable is declared with a type other than `any`, create a fresh variable for it and downcast using `FromDyn`.
2. Do the for-in desugaring in the checker.lua step instead of the to_ir.lua step. This would allow us to use `check_exp_verify` to automatically insert those downcasts, just as we are doing in the rest of the checker. However, if we use this strategy  we would have to teach the to_ir step how to recognize the `== nil` and compile it to the special `IsNil` instruction.

srijan-paul:
Hi, I'm kinda busy with an upcoming atm, so that's what's been causing the delay. I'll try to sort the changes out by end of this week at the latest.

hugomg:
You chose a good time to be busy; right now I have also been super busy myself XD

But don't worry. Any kind of help is always greatly appreciated and we aren't in a hurry :)

srijan-paul:
Hi, loop declarator annotations are now supported.

In the generated C code, should the loop `decl`s that are down casted later (`i_any` , `x_any` etc) have a name or nah? Currently they're named and have the same name as the original declarator followed by a `_dyn` (eg - `x_dyn`). My reasoning was that when a loop has many `decls` in the LHS, the generated C code will be easier to read since there will be a visible one to one correspondence between the `x` and `x_any` in the comments.  

Lastly, In `execution_tests.lua` the definitions  of `my_ipairs` and `iter` is now being repeated across every `describe` block. One *could* store the definition of these functions along with `typealias iterfn = ...` in a string and then do something akin to:
```lua
describe("...", function()
   compile(loop_function_defs .. [[ -- test for in loops here ]])
end)
```

The only thing this would do is reduce the size of `execution_tests.lua` so I'm not really sure if that's worth it. Which way would you rather have it be?

hugomg:
> should the loop decls that are down casted later (i_any , x_any etc) have a name or nah

That's a great idea. The only suggestion I'd make is to add a `$` at the start of the name, to make it clear that this is an internal variable created by the compiler. For example, `$x_dyn`.

> Lastly, In execution_tests.lua the definitions of my_ipairs and iter is now being repeated across every describe block.

I think the simplest solution would be to have a single `describe` block, containing a single `compile` block and several `it` blocks for the test cases. This is what we tend to do in the rest of the test suite. A good example would be the "Function calls" block.

srijan-paul:
About the changes to `manual.md`. I assume all that needs to be added is a simple example procedure using `for-in` loops and changes to the EBNF grammar at the end?

hugomg:
That's the basic idea. It's OK to refer to the Lua reference manual if you want. By the way, make sure that you explain how the types work, because that's the part that is new in Pallene and doesn't exist in Lua.

pallene/to_ir.lua
```diff
@@ -205,6 +205,71 @@ function ToIR:convert_stat(cmds, stat)
...
         table.insert(cmds, ir.Cmd.For(stat.loc, v, start, limit, step, ir.Cmd.Seq(body)))
 
+    elseif tag == "ast.Stat.ForIn" then
+        local decls = stat.decls
+        local exps = stat.exps
+
+        -- for-in loops are desugared into regurlar loops before compiling.
+        -- For example, a loop like this:
+        --- ```
+        -- for a: T, b: T2 in <RHS> do
```
hugomg, line 215: We could say `a: T1` instead of `a: T`.

pallene/to_ir.lua
```diff
@@ -205,6 +205,71 @@ function ToIR:convert_stat(cmds, stat)
...
+        -- local iter, st, ctrl = RHS[1], RHS[2], RHS[3]
+        -- while true do
+        --   local a_any, b_any = iter(st, ctrl)
+        --   if a_any == nil then break end
+        --   local a = a_any as T
+        --   local b = b_any as T2
+        --   ctrl = i_any
+        --   <loop body>
+        -- end
+        -- ```
```
hugomg, line 230: A question: Would it be a good idea to unify the `a_any` and `ctrl` variables, to create one less variable?

```
local iter, st, a_any, b_any
iter, st, a_any = RHS[1], RHS[2], RHS[3]
while true do
    a_any, b_any = iter(st, a_any)
    if a_any = nil then break end
    local a = a_any as T1
    local b = b_any as T2
    <loop body>
end
```

hugomg (commented):
Looking great! The tests are very clean now. I think the onlys thing we need to do before we can merge this PR is to update the manual and to address some very minor points in the desugaring.

We can leave the ipairs builtin for another PR.

srijan-paul:
Hi there, I'll try and get these changes made and pushed by the end of tomorrow. I'll make the `ipairs` PR within this week. I'm kinda in the middle of my finals so that's what's been up with me lately I guess lol.

pallene/to_ir.lua
srijan-paul: Not sure, My personal POV is that these are locals anyway and not using the same symbol would make it much easier to spot the 1:1 correspondence between variables inserted by the compiler and those written by the user. That said, I think it's best we go with your gut. What do you say?

hugomg:
> My personal POV is that these are locals

I suspect that reducing the number of variables might be more efficient in this case though. In Pallene, the garbage collector has to do a bit of work when we use variables of type any. Whenever we assign to a variable with a gargabe-collectable type, the compiler also generates code to save that value to the Lua stack. This happens in the assignment `ctrl = i`, because ctrl has type any and could potentially contain a garbage-collectable value.

The reason we have to do this is that the Lua garbage collector doesn't know how to look inside the C stack, to find out what values are being kept alive by Pallene local variables.

srijan-paul:
Hi, I unified the two variables and added some documentation to `manual.md`. Not sure if I wrote too much/ too little. Also, there is a failing test in line 389 of execution tests, however it seems to be caused by changes to this repo's master branch? The test suite still passes however.

hugomg (approved):
Excelent!! Let's get this merged :)

hugomg:
> Not sure if I wrote too much/ too little.

It looks fine to me. The important thing is to get things documented. I can always edit the text later if necessary.

> Also, there is a failing test in line 389 of execution tests if I use the ./test-project script instead of busted -o gtest -v ./spec. Is that intended?

Uh, that's unusual. But it doesn't happen on my machine. Does that failing test in line 389 always happen or only sometimes? Also, does recompiling the project with `make clean; make linux-readline` help?

srijan-paul:
That seems to have worked, thanks.
Should I open a PR for `ipairs`?

hugomg:
Sure, feel free to continue what we've been doing and implement the ipairs. As before, it would be a welcome addition.

First step is to type check programs with ipairs. Add the ipairs function to the list of builtin function, and create some test cases with ipairs. This point would be a good spot for a round of code review.

Next step after that would be to optimize the loops with ipairs. In the to_ir, recognize when the programmer is using ipairs in a for loop, and generate more efficient code that directly accesses the table without calling the iterator function at each step. It's OK if we copy paste some of the logic from the generic for-in loops. The implementation of the optimized ipairs loop is likely to change in the near future as we continue working on optimizations so I don't want to make it too coupled to the non-optimized implementation. Better give it some time to settle before worrying too much about code duplication.

Finally, it's also OK if we don't actually implement calling ipairs as a function outside of a for-in loop, at least for now. You can have Pallene raise a compile time error if the programmer tries to do that.

srijan-paul:
Hi, I realize that I am going on a bit of tangent and this isn't strictly on-topic, but I was wondering how/if 
`pairs` is possible in Pallene record instances? Since one can't really iterate over structs in C. Are there any plans to support that sometime in the future?

hugomg:
Pallene has two kinds of record types, the Pallene records (declared with the "record" keyword implemented as userdata) and the Lua records (implemented as tables). We really ought to rename them to something less confusing but anyway, anwering with the names we have today...

For the userdata records, Lua cannot iterate over them using pairs. Would make more sense if Pallene also would not.

When we figure out a way to add a map datatype to Pallene (#124), I would expect that it would be desirable to also allow pairs to work on maps. Just like ipairs is the natural way to iterate over an  array, pairs is the natural way to iterate over a map.

For other table-backed types, including arrays and Lua records, I'm not sure if we need or want to make them iterable with pairs. In theory it would be possible, but the tradeoff in additional type system complexity might not be worth it. Perhaps  one escape valve would be to allow arrays and records to be casted to an any=>any map using an explicit type conversion, as an "escape valve"? Something to think about when we design the map data type...

srijan-paul:
Yeah can see a map being ideal for pairs too. So the usage of pairs would be restricted to maps, and the maps compile down to C hashtables? Maybe one could make them nearly as fast as structs if the index hashes are computed at compile time, since we would have all the necessary information (table, name of the index as a string).

By the way, I have been trying to embed Pallene in my game engine on Manjaro i3, (I tried building on windows but it seems that it would require replacing any `so` libs with `dll`s and then replacing any possible calls to LoadLibrary with `dlopen` in the C headers ?). 

I was wondering if there is any reference/guide for embedding (similar to how one would embed Lua) or if I'm gonna have to wing it.

Thanks.

hugomg:
> So the usage of pairs would be restricted to maps, and the maps compile down to C hashtables?

The kind of map type that we talk about in issue 131 would be implemented as a plain Lua table, similarly to how Pallene arrays are implemented as Lua tables. 

> Maybe one could make them nearly as fast as structs if the index hashes are computed at compile time, since we would have all the necessary information (table, name of the index as a string).

Since the main application of a map datatype is when the keys are not known at compile-time, I don't think we can do that kind of optimization. In fact, I'd expect that there won't be big optimizations of the sort for the map datatype. The complicated part about them is actually the type system. We haven't decided yet what should happen if we try to read a key that is not in the map. Should it raise an error (as is the case for arrays) or should it return nil? ...

The Lua records (the ones declared with syntax `typealias Foo = {x: integer}`) are also implemented as tables but in this case we do know  the key names at compile-time. We implement an "inline cache" optimization where we save the index where we last found the key in the hash table. Not every Lua table will store the same key at the same place. However, tables that are used close to each other often were created together and store their keys in the same order, meaning that if you cache the index for the last table you can speed up the hash table lookup for the next one.

That said, if the objective is to get something that is as good as a struct, the way to go is to use a Pallene record (the ones declared with syntax `record x: integer end`). Those are literally implemented as a C struct and are exposed to Lua as an opaque userdata.

> By the way, I have been trying to embed Pallene in my game engine on Manjaro i3

Very cool! We never tried doing anything of the sort, so let us know if you can get it working :)

At the moment, I'd expect that you'll need to embed our modified version of Lua instead of vanilla version of Lua. As for the Pallene code itself, if the `so` files generated by `pallenec` don't work, one thing you can try is to use `pallenec --emit-c` to generate C code and then you can compile that C code by hand, using the compiler and compiler flags of your choice.

> I tried building on windows but it seems that it would require replacing any so libs with dlls

Yeah, we still need to implement Windows support. My guess is that there is not much that is *fundamentally* Linux only but until someone puts the hard work of porting Pallene to windows it's hard to say for sure. I'd expect that the main things that we will need to change are related to compiling the C code (either Lua itself, or the C code that Pallene generates). For example, generating dlls instead of so... 

Another problem is that, while Lua itself is written in C89, the pallene-specific code in pallene_core.h is using some C99 features. It may or may noto work on MSVC compiler (although I expect that it should work with GCC or Clang).

> and then replacing any possible calls to LoadLibrary with dlopen in the C headers

We probably won't need to something like that. We have not written any Pallene-specific code for loading dynamic libraries. That part is handled by Lua, inside the loadlib.c.

srijan-paul:
> We probably won't need to something like that. We have not written any Pallene-specific code for loading dynamic libraries. That part is handled by Lua, inside the loadlib.c.

Right, I confused the the `.so` files that Pallene results in.

> As for the Pallene code itself, if the so files generated by pallenec don't work, one thing you can try is to use pallenec --emit-c to generate C code and then you can compile that C code by hand, using the compiler and compiler flags of your choice.

Good idea! I could get a Python script to compile that to dlls for the engine. Thanks.

> In fact, I'd expect that there won't be big optimizations of the sort for the map datatype. 
Looks like I misunderstood the map datatype as an iterable record, whoops.

About the porting to windows part, I could give it a try after getting this `ipairs` thing done.

PR #343, opened 2020-12-09
Update Lua to version 5.4.2

### LINKED ISSUE #338: Update Lua to Lua 5.4.2
They changed the stack implementation back to "stackless", which uses less C stack space.

We should check if the rc1 is breaking anything in Pallene and then when the final 5.4.2 comes out we can update our bundled version of Lua.

We should also check if the stream sieve benchmark is still using a workaround for the C stack space.  If yes, it's possible that we might be able to get rid of that.

hugomg:
Closes #338.


hugomg:
By the way, remember to rerun `make` in your local copy of the repository. Every time we update Lua it's required that you rebuild your version of the Pallene runtime libraries.

mambrozio (approved):

PR #346, opened 2021-01-10
Add support for looping with ipairs.

srijan-paul:
This PR introduces an `ipairs` builtin to mirror that of Lua. (to address #164)
Only difference being, the `ipairs` in Pallene is desugared to generic for in loops.

* `ipairs` builtin added to `builtins.lua`
* `checker.lua` now marks loop statement nodes with unshadowed ipairs as (is_ipairs)
* `to_ir.lua` now handles ipairs separately
* `spec/execution_tests.lua` has one test cases added to test looping with ipairs.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -482,11 +482,6 @@ function Checker:check_stat(stat)
                 types.tostring(itertype), types.tostring(iteratorfn._type))
         end
 
-        if iteratorfn._tag == "ast.Exp.CallFunc" then
-            local iterfn_var = iteratorfn.exp.var._name
-            stat.is_ipairs = iterfn_var._tag == "checker.Name.Builtin" and iterfn_var.name == "ipairs"
-        end
-
         rhs[2] = self:check_exp_synthesize(rhs[2])
         rhs[3] = self:check_exp_synthesize(rhs[3])
 
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -482,6 +482,11 @@ function Checker:check_stat(stat)
                 types.tostring(itertype), types.tostring(iteratorfn._type))
         end
 
+        if iteratorfn._tag == "ast.Exp.CallFunc" then
+            local iterfn_var = iteratorfn.exp.var._name
+            stat.is_ipairs = iterfn_var._tag == "checker.Name.Builtin" and iterfn_var.name == "ipairs"
+        end
+
         rhs[2] = self:check_exp_synthesize(rhs[2])
         rhs[3] = self:check_exp_synthesize(rhs[3])
 
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -209,8 +209,15 @@ function ToIR:convert_stat(cmds, stat)
         local decls = stat.decls
         local exps = stat.exps
 
+        local is_ipairs = false
+        local iterfn = exps[1]
+        if iterfn._tag == "ast.Exp.CallFunc" and iterfn.exp._tag == "ast.Exp.Var" then
+            local iterfn_var = iterfn.exp.var._name
+            is_ipairs = iterfn_var._tag == "checker.Name.Builtin" and iterfn_var.name == "ipairs"
+        end
 
-        if stat.is_ipairs then
+
+        if is_ipairs then
             -- `ipairs` are desugared down to regular for-loops
             -- ```
             -- for i, x in ipairs(xs) do
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -220,56 +220,87 @@ function ToIR:convert_stat(cmds, stat)
         if is_ipairs then
             -- `ipairs` are desugared down to regular for-loops
             -- ```
-            -- for i, x in ipairs(xs) do
+            -- for i: T1, x: T2 in ipairs(xs) do
             --   <loop body>
             -- end
             -- ```
             -- would get compiled down to:
             -- ```
-            -- local x
-            -- local i_num
-            -- for i_num = 1, #xs do
-            --   i = i_num
-            --   x = xs[i]
+            -- local x_dyn: any
+            -- local i_num: integer = 1
+            -- local limit: integer = #xs
+            -- while true do
+            --   if i_num > limit then
+            --     break
+            --   end
+            --   local i = i_num as T1
+            --   x_dyn = xs[i_num]
+            --   if x_dyn == nil then
+            --     break
+            --   end
+            --   local x = x_dyn as T2 
             --   <loop body>
+            --   i_num = i_num + 1
             -- end
             -- ```
-
             -- the table passed as argument to `ipairs`
             local arr =  exps[2].call_exp.args[1]
             local v_arr = ir.add_local(self.func, "$xs", arr._type)
             self:exp_to_assignment(cmds, v_arr, arr)
 
-            local v_inum = ir.add_local(self.func, "$i_num", types.T.Integer())
+            -- local i_num: integer = 1
+            local v_inum = ir.add_local(self.func, "$"..decls[1].name.."_num", types.T.Integer())
             local start = ir.Value.Integer(1)
             table.insert(cmds, ir.Cmd.Move(stat.loc, v_inum, start))
-
-            local step = ir.Value.Integer(1)
-            local v_limit = ir.add_local(self.func, "$xs_len", types.T.Integer())
+        
+            -- local limit: integer = 1
+            local v_limit = ir.add_local(self.func, "$"..decls[2].name.."_len", types.T.Integer())
             table.insert(cmds, ir.Cmd.Unop(stat.loc, v_limit, "ArrLen", ir.Value.LocalVar(v_arr)))
-
             local limit = ir.Value.LocalVar(v_limit)
 
+            -- body of the while loop.
+            local body = {}
+
+            -- if i_num > limit then break end
+            local v_cond = ir.add_local(self.func, "$loop_limit_reached", types.T.Boolean())
+            table.insert(body, ir.Cmd.Binop(stat.loc, v_cond, "IntGt", ir.Value.LocalVar(v_inum), limit))
+            table.insert(body, ir.Cmd.If(stat.loc, ir.Value.LocalVar(v_cond), ir.Cmd.Break(), ir.Cmd.Nop()))
+
+            -- local i: T1 = i_num as T1
             local v_i = ir.add_local(self.func, decls[1].name, decls[1]._type)
             self.loc_id_of_decl[decls[1]] = v_i
-            local v_x = ir.add_local(self.func, decls[2].name, decls[2]._type)
-            self.loc_id_of_decl[decls[2]] = v_x
-
-            local body = {}
-            -- i = start
             if decls[1]._type._tag == "types.T.Integer" then
-                table.insert(cmds, ir.Cmd.Move(stat.loc, v_i, ir.Value.LocalVar(v_inum)))
+                table.insert(body, ir.Cmd.Move(stat.loc, v_i, ir.Value.LocalVar(v_inum)))
             else
                 table.insert(body, ir.Cmd.ToDyn(stat.loc, types.T.Integer(), v_i, ir.Value.LocalVar(v_inum)))
             end
-            -- x = xs[i]
+
+            -- x_dyn = xs[i]
+            local v_x_dyn = ir.add_local(self.func, "$"..decls[2].name.."_dyn", types.T.Any())
+            self.loc_id_of_decl[decls[2]] = v_x_dyn
             local src_arr =  ir.Value.LocalVar(v_arr)
             local src_i =  ir.Value.LocalVar(v_inum)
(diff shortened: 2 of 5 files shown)


srijan-paul:
So I'm not sure if having the checker flag a node as `is_ipairs` is the way to go when differentiating builtin loops from *other* for-in loops but it was really simple so I went with it for now.

pallene/checker.lua
```diff
@@ -482,6 +482,11 @@ function Checker:check_stat(stat)
                 types.tostring(itertype), types.tostring(iteratorfn._type))
         end
 
+        if iteratorfn._tag == "ast.Exp.CallFunc" then
+            local iterfn_var = iteratorfn.exp.var._name
```
hugomg, line 486: Before we access the `.var` field we have to check that `iteratorfn.exp._tag` is "ast.Exp.ExpVar".

pallene/checker.lua
```diff
@@ -482,6 +482,11 @@ function Checker:check_stat(stat)
...
+            stat.is_ipairs = iterfn_var._tag == "checker.Name.Builtin" and iterfn_var.name == "ipairs"
```
hugomg, line 487: I think we can look at the `_name` annotation in the to_ir.lua. The checker doesn't need to know about this optimization that we are doing with ipairs.

pallene/to_ir.lua
```diff
@@ -209,66 +209,123 @@ function ToIR:convert_stat(cmds, stat)
...
-        local v_lhs_dyn = {}
-        for _, decl in ipairs(decls) do
-            local v = ir.add_local(self.func, "$" .. decl.name .. "_dyn", types.T.Any())
-            table.insert(v_lhs_dyn, v)
-        end
 
-        local v_ctrl = v_lhs_dyn[1]
-        self:exp_to_assignment(cmds, v_ctrl, exps[3])
+        if stat.is_ipairs then
+            -- `ipairs` are desugared down to regular for-loops
```
hugomg, line 214: I suspect that this desugaring might be wrong in situations where the table has "holes" with nil. When we use ipairs it always stops at the first nil, which might not be the same thing as the `#xs` in a table with holes.

What do you think about using the same desugaring that we used for the regular for-in loop, except that instead of calling an iterator function, we use `xs[i]` and `i = i+1` instead? Basically just "inlining" the definition of the iter function for ipairs.

spec/execution_tests.lua
```diff
@@ -1814,6 +1814,17 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
             end
+
+            -----------------------
+
+            export function double_list_ipairs(xs: {integer}): {integer}
+                local out: {integer} = {}
+                for i, x in ipairs(xs) do
+                    out[i] = x as integer * 2
+                end
+
```
hugomg, line 1825: We can avoid this extra space.

hugomg (commented):
Thanks! This is looking good.

Since you mentioned that you are thinking about adding extra tests, I think I'll wait until after that to give a closer look at the actual implementation of the ipairs desugaring. Specially because I suspect that we may also need to rewrite that part to use a while loop instead of a for loop with `#xs`.

hugomg:
> So I'm not sure if having the checker flag a node as is_ipairs is the way to go

Inspecting the `_name` tag as you are doing now is certainly the way to. The only suggestion I'd make is that we can do that inspection in the to_ir instead of in the checker. This way we don't need to add that weird "is_ipairs" field.

srijan-paul:
Hi! I changed the IR gen to now desugar loops with `ipairs` to `while` loops.
Also added 2 more tests, one for typed LHS and for breaking at `nil` values within the array.

pallene/to_ir.lua
```diff
@@ -209,60 +209,99 @@ function ToIR:convert_stat(cmds, stat)
         local decls = stat.decls
         local exps = stat.exps
 
+        local is_ipairs = false
+        local iterfn = exps[1]
+        if iterfn._tag == "ast.Exp.CallFunc" and iterfn.exp._tag == "ast.Exp.Var" then
+            local iterfn_var = iterfn.exp.var._name
+            is_ipairs = iterfn_var._tag == "checker.Name.Builtin" and iterfn_var.name == "ipairs"
+        end
```
hugomg, line 217: The variable names here might be a bit misleading. The iterfn refers to a function *call*, not to a function. And the iterfn_var refers to a Checker.Name, not to a Var. What do you think about just getting rid of those variables?

```
local e1 = exps[1]
local is_ipairs = (
    e1._tag == "ast.Exp.CallFunc" and
    e1.exp._tag == "ast.Exp.Var" and
    e1.exp.var._name._tag == "checker.Name.Builtin" and
    e1.exp.var._name.name == "ipairs")
```

hugomg (changes requested):
Thanks, I think the new tests help a lot.

The new desugaring is better, but I think it is not quite there yet. We should not use `#xs`; see the other comment for the details.

I also noticed something when I reread the code for the generic for-in loops. In some places we are using `_any` but in others we are using `_dyn`. I think either is fine but it would be better to be consistent. Could you please choose one and update the code and the comments to always use that one?

pallene/to_ir.lua
```diff
@@ -209,60 +209,99 @@ function ToIR:convert_stat(cmds, stat)
...
+            --     break
+            --   end
+            --   local i = i_num as T1
+            --   x_dyn = xs[i_num]
+            --   if x_dyn == nil then
+            --     break
+            --   end
+            --   local x = x_dyn as T2 
             --   <loop body>
+            --   i_num = i_num + 1
```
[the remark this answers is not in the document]
srijan-paul: > In order for the `xs[i_num]` to produce an any as we want, the xs must have type {any}. Are we already doing this?

Would you say that's the checker's job to make sure the ipair's argument is of type `{any}`?
Or do we do it in `to_ipairs`?

> ps.: By the way, I think I found a small bug with the comment for the generic for-in loop. We should be using a_any as the control variable instead of ctrl.

Whoops, yeah I missed that.

pallene/to_ir.lua
srijan-paul: Yeah, This looks good. I initially was under the impression that using those names would make it easier to read but I think this looks better and reads more naturally.

pallene/to_ir.lua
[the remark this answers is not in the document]
hugomg: I think the checker probably already does it, but I don't remember 100%. Might be worth adding an `assert` call, just in case.

srijan-paul:
So I added an assert guard for `{any}` as ipairs argument, and changed the desugaring to only break on holes.
I'm wondering if we need any more tests?

pallene/to_ir.lua
```diff
@@ -209,66 +209,150 @@ function ToIR:convert_stat(cmds, stat)
...
+            -- x_dyn = xs[i_num]
+            local v_x_dyn = ir.add_local(self.func, "$"..decls[2].name.."_dyn", types.T.Any())
+            local src_arr =  ir.Value.LocalVar(v_arr)
+            local src_i =  ir.Value.LocalVar(v_inum)
+            table.insert(body, ir.Cmd.GetArr(stat.loc, types.T.Any(), v_x_dyn, src_arr, src_i))
+
+            -- if x_dyn == nil then break end
+            local v_cond_checknil = ir.add_local(self.func, "$is_"..decls[2].name.."_nil", types.T.Boolean())
+            table.insert(body, ir.Cmd.IsNil(stat.loc, v_cond_checknil, ir.Value.LocalVar(v_x_dyn)))
+            table.insert(body, ir.Cmd.If(stat.loc, ir.Value.LocalVar(v_cond_checknil), ir.Cmd.Break(), ir.Cmd.Nop()))
```
hugomg, line 265: Since the `_is_nil` variable is only used in the next command after this one, I think it's OK if we just don't give a name for it. That is, pass `false` as the second argument to `add_local`. This should also avoid the issue with `decls[2]` that I previously mentioned.

pallene/to_ir.lua
```diff
@@ -209,66 +209,150 @@ function ToIR:convert_stat(cmds, stat)
...
+            end
+
+            -- local x = x_dyn as T2
+            local v_x = ir.add_local(self.func, decls[2].name, decls[2]._type)
+            self.loc_id_of_decl[decls[2]] = v_x
+            if decls[2]._type._tag == "types.T.Any" then
+                table.insert(body, ir.Cmd.Move(stat.loc, v_x, ir.Value.LocalVar(v_x_dyn)))
+            else
+                table.insert(body, ir.Cmd.FromDyn(stat.loc, decls[2]._type, v_x, ir.Value.LocalVar(v_x_dyn)))
+            end
```
hugomg, line 283: In this block of code we are also using `decls[2]`. If we choose to allow `for i in ipairs(xs)` then we need to update this part of the code. If we chose to forbid that kind of loop then this part of the code can stay unchanged.

hugomg (commented):
I looked at the code more carefully now that the desugaring seems to be alright.

The main issue I noticed is that we are currently assuming that `#args == 1` and that `#decls == 2`. I believe that the type checker currently forbids both of these, but it would be nice to have test cases checking that. We should also add appropriate `assert` calls to the `to_ir` code, to document those assumptions.

pallene/to_ir.lua
srijan-paul: Ah yes. I don't think ipairs is used just for iterating from `1` to `#xs`, but at the same time It would be good to have Pallene resemble Lua as closely as possible and be a rough superset. So yea I'll change the `decls[2]` part.

pallene/to_ir.lua
```diff
@@ -209,66 +209,150 @@ function ToIR:convert_stat(cmds, stat)
...
+            --   end
+            --   local i = i_num as T1
+            --   local x = x_dyn as T2 
+            --   <loop body>
+            --   i_num = i_num + 1
+            -- end
+            -- ```
+
+            -- the table passed as argument to `ipairs`
+            local arr =  exps[2].call_exp.args[1]
```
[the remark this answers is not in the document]
srijan-paul: True, the error for ipairs won't make a lot of sense, I'll open an issue.

pallene/to_ir.lua
hugomg: The golden rule that we follow is that if something is allowed in Pallene, then that program should behave the same as a the Lua program that you get by erasing all the type annotations. Pallene is allowed to give an error for things that would not be an error in Lua.

> Pallene mandates the LHS decl count and RHS return count be the same for loops so I'm kinda torn on this one.

If we choose to allow ipairs with a different number of variables, I think we would also have to change this rule accordingly (and the generic case of the for-in loops as well).

That said, since 99% of the time the programmers are going to want to use two variables in the ipairs, I think it's OK to keep the current behavior, at least for now. If it turns out that for-in loops with a mismatched number of arguments are something that people actually need we can always change it later.

pallene/to_ir.lua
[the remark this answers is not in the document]
srijan-paul: > ipairs is called with one argument, which is a function call returning one value

Should this really be an error if the value is a table value?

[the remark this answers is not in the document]
hugomg: Now that you mention it, I agree that it would be redundant to test what happens the argument to ipairs is a function call.
The important thing to test is what happens if we call ipairs with multiple arguments, where the additional arguments might have side effects.

    ipairs(f(), g())
    
In the current implementation, I believe the call to `g()` would not be generated at all, which can be a problem if the function call does something (like printing some output, or modifying a mutable table).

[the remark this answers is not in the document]
hugomg: Just to be clear, I think that what should be done in that example with the `g()` is raise a compile-time type error saying that it expected 1 argument but found 2. Then, in the `to_ir` we can put an assert saying that `#args == 1`.

The example was just to motivate why I think it is important to assert that `#args == 1` in the `to_ir`. Because if it there were more arguments, the current implementation would be buggy. Since this is an important assumption, might as well assert it to be sure that it holds.

srijan-paul:
Added some more assert guards and tests.
Though for now `ipairs` can only be called with exactly two variables on the loop LHS to keep things plain and simple.

srijan-paul:
Since we'll be throwing specific error message for `ipairs` when it has 3 or more decls in the LHS, is that going to be done in the checker?

srijan-paul:
Alright, In that case I think I might have addressed the issue of missing `assert` guards and tests to my knowledge, Just to be clear. If you think there are some more missing tests, I'll have to add those too. Mentioned because I'm not sure what code paths remain to be tested.

pallene/to_ir.lua
```diff
@@ -209,66 +209,150 @@ function ToIR:convert_stat(cmds, stat)
...
+            assert(arr._type.elem._tag == "types.T.Any")
```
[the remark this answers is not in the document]
hugomg: @srijan-paul I think the only thing left to do is delete the line with the `assert(arr._type.elem._tag`. Everything else looks perfect.

hugomg (changes requested):

srijan-paul:
> I think the only thing left to do is delete the line with the assert(arr._type.elem._tag. Everything else looks perfect.

Done. :+1:

hugomg (approved):
Excellent! Thank you very much.

PR #349, opened 2021-01-21
ir change proposal: NormArr

leokaplan:
This branch envision the optimization to renormalize_array calls.
A first step is to decouple the calls from SetArr and GetArr commands. 
Here I am proposing a new command, NormArr, that only performs renormalize_array calls.
Later we will be identifying the minimal amount of renormalize calls to be added to the emitted code.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1141,18 +1141,15 @@ gen_cmd["GetArr"] = function(self, cmd, _func)
     local arr = self:c_value(cmd.src_arr)
     local i   = self:c_value(cmd.src_i)
     local dst_typ = cmd.dst_typ
-    local line = C.integer(cmd.loc.line)
 
     return (util.render([[
         {
-            pallene_renormalize_array(L, $arr, $i, PALLENE_SOURCE_FILE, $line);
             TValue *slot = &$arr->array[$i - 1];
             $get_slot
         }
     ]], {
         arr = arr,
         i = i,
-        line = line,
         get_slot = self:get_luatable_slot(dst_typ, dst, "slot", arr,
             cmd.loc, "array element"),
     }))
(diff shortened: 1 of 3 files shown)


pallene/coder.lua
```diff
@@ -1163,22 +1160,34 @@ gen_cmd["SetArr"] = function(self, cmd, _func)
...
         set_heap_slot = set_heap_slot(src_typ, "slot", v, arr),
     }))
 end
 
+gen_cmd["NormArr"] = function(self, cmd, _func)
+    local arr = self:c_value(cmd.src_arr)
+    local i   = self:c_value(cmd.src_i)
+    local line = C.integer(cmd.loc.line)
+    return (util.render([[
+        {
```
hugomg, line 1181: Acho que não precisa desse `{}`.

pallene/coder.lua
```diff
@@ -1163,22 +1160,34 @@ gen_cmd["SetArr"] = function(self, cmd, _func)
...
     ]], {
         arr = arr,
         i = i,
-        v = v,
-        line = line,
         set_heap_slot = set_heap_slot(src_typ, "slot", v, arr),
     }))
 end
 
+gen_cmd["NormArr"] = function(self, cmd, _func)
```
hugomg, line 1176: Que tal chamar de `NormalizeArr`?

pallene/ir.lua
```diff
@@ -136,6 +136,7 @@ local ir_cmd_constructors = {
 
     GetArr     = {"loc", "dst_typ", "dst", "src_arr", "src_i"},
     SetArr     = {"loc", "src_typ",        "src_arr", "src_i", "src_v"},
+    NormArr    = {"loc", "src_typ",        "src_arr", "src_i"},
```
hugomg, line 139: O que você acha de botar um comentário explicando as invariantes desses comandos?

```
    -- Arrays
    -- For performance reasons, GetArr and SetArr require that the index be in the array part of
    -- the Lua table. That is always the case for arrays returned by NewArr and by RenormalizeArr.
```

pallene/to_ir.lua
```diff
@@ -613,6 +614,8 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
                 local iv = ir.Value.Integer(i)
                 local vv = self:exp_to_value(cmds, field.exp)
                 local src_typ = field.exp._type
+                -- one could add a NormArr here, but we can assume
+                -- that a new array is already normalized up to sizehint
```
hugomg, line 618: Eu escreveria esse comentário como

```
-- We don't need to renormalize here because NewArr returns a normalized array
```

hugomg (changes requested):
Acho que também faltou botar os renormalizes do lado do `GetArr`?

Além disso, você poderia fazer o rebase dessa PR na versão mais recente do master, que contém a nova implementação do `ipairs`?

hugomg:
> Later we will be identifying the minimal amount of renormalize calls to be added to the emitted code.

Just to be clear, I'd prefer if that were on a separate optimization pass  instead of on the to_ir.

Sure, we can implement that optimization pass by deleting all the renormalize operations and adding them back in only where necessary. But either way, I'd be happier if we could put the complicated logic on a separate file.

leokaplan:
> > Later we will be identifying the minimal amount of renormalize calls to be added to the emitted code.
> 
> Just to be clear, I'd prefer if that were on a separate optimization pass instead of on the to_ir.
> 
> Sure, we can implement that optimization pass by deleting all the renormalize operations and adding them back in only where necessary. But either way, I'd be happier if we could put the complicated logic on a separate file.

I agree, that was already my plan. Sorry that it wasn't clear enough.

hugomg:
There is a test failing. Maybe it is because the code for `ipairs` also has a `GetArr` needs a NormalizeArr.

PR #361, opened 2021-04-13
  Remove export keyword and expect statements at the toplevel

VictorNogueiraRio:
Instead of exporting variables and functions by means of the `export` keyword, this PR implements a different approach, similar to the one Lua uses, that is, if you want to export variables or functions in your program, you should create a variable of type `module` and return it at the end of the program.
Example:

```
local m: module = {}
m.var = 22.0
return m
```

Other examples are available at the `examples` folder in this PR.

### DIFF
--- a/pallene/ast.lua
+++ b/pallene/ast.lua
@@ -12,7 +12,7 @@ local function declare_type(type_name, cons)
 end
 
 declare_type("Program", {
-    Program = {"tls", "type_regions", "comment_regions"}
+    Program = {"loc", "tls", "type_regions", "comment_regions"}
 })
 
 declare_type("Type", {
@@ -21,11 +21,11 @@ declare_type("Type", {
     Array    = {"loc", "subtype"},
     Table    = {"loc", "fields"},
     Function = {"loc", "arg_types", "ret_types"},
+    Module   = {"loc"},
 })
 
 declare_type("Toplevel", {
-    Func      = {"loc", "visibility", "decl", "value"},
-    Var       = {"loc", "visibility", "decls", "values"},
+    Stat      = {"loc", "stat"},
     Typealias = {"loc", "name", "type",},
     Record    = {"loc", "name", "field_decls"},
 })
@@ -46,6 +46,7 @@ declare_type("Stat", {
     Call   = {"loc", "call_exp"},
     Return = {"loc", "exps"},
     Break  = {"loc"},
+    Func   = {"loc", "name", "decl", "value"},
 })
 
 -- Things that can appear in the LHS of an assignment. For example: x, x[i], x.name
@@ -82,28 +83,5 @@ declare_type("Field", {
 --
 -- note: the following functions are why we need `if type(conss) == "table"` in parser.lua
 --
-
--- Returns a sequence containing the variable names declared by the specified
--- toplevel node.
-function ast.toplevel_names(tl_node)
-    local names = {}
-    local tag = tl_node._tag
-    if     tag == "ast.Toplevel.Func" then
-        table.insert(names, tl_node.decl.name)
-    elseif tag == "ast.Toplevel.Var" then
-        for _, decl in ipairs(tl_node.decls) do
-            table.insert(names, decl.name)
-        end
-    elseif tag == "ast.Toplevel.Typealias" then
-        table.insert(names, tl_node.name)
-    elseif tag == "ast.Toplevel.Record" then
-        table.insert(names, tl_node.name)
-    elseif tag == "ast.Toplevel.Builtin" then
-        table.insert(names, tl_node.name)
-    else
-        typedecl.tag_error(tag)
-    end
-    return names
-end
 
 return ast
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -21,6 +21,8 @@ local util = require "pallene.util"
 --      - ast.Toplevel.Record
 --
 --   * _name: In ast.Var.Name nodes, a checker.Name that points to the matching declaration.
+--   * _modname: In ast.Stat.Decl nodes, which represent a module variable or function declaration.
+--   It holds the module's name. For instance, m.var = 2, would have _modname = "m"
 --
 -- We also make some adjustments to the AST:
 --
@@ -120,7 +122,7 @@ declare_type("Name", {
     Global   = { "decl" },
     Function = { "decl" },
     Builtin  = { "name" },
-    Module   = { "name" }
+    Module   = { "name", "is_main_mod" }
 })
 
 function Checker:add_type(name, typ)
@@ -133,14 +135,26 @@ function Checker:add_local(decl)
     self.symbol_table:add_symbol(decl.name, checker.Name.Local(decl))
 end
 
+local function verify_mod_decl(name, decl, symbol_table)
+    local cname = symbol_table:find_symbol(name)
+    if decl._modname and cname then
+        scope_error(decl.loc,
+          "duplicate module field '%s', previous one at line %d", decl.name, cname.decl.loc.line)
+    end
+end
+
 function Checker:add_global(decl)
     assert(decl._tag == "ast.Decl.Decl")
-    self.symbol_table:add_symbol(decl.name, checker.Name.Global(decl))
+    local name = decl._modname and (decl._modname .. '.' .. decl.name) or decl.name
+    verify_mod_decl(name, decl, self.symbol_table)
+    self.symbol_table:add_symbol(name, checker.Name.Global(decl))
 end
 
 function Checker:add_function(decl)
     assert(decl._tag == "ast.Decl.Decl")
-    self.symbol_table:add_symbol(decl.name, checker.Name.Function(decl))
+    local name = decl._modname and (decl._modname .. '.' .. decl.name) or decl.name
+    verify_mod_decl(name, decl, self.symbol_table)
+    self.symbol_table:add_symbol(name, checker.Name.Function(decl))
 end
 
 function Checker:add_builtin(name, id)
@@ -148,9 +162,9 @@ function Checker:add_builtin(name, id)
     self.symbol_table:add_symbol(name, checker.Name.Builtin(id))
 end
 
-function Checker:add_module(name)
+function Checker:add_module(name, is_main_mod)
     assert(type(name) == "string")
-    self.symbol_table:add_symbol(name, checker.Name.Module(name))
+    self.symbol_table:add_symbol(name, checker.Name.Module(name, is_main_mod))
 end
 
 --
@@ -212,32 +226,12 @@ local letrec_groups = {
     ["ast.Toplevel.Func"]      = "Func",
     ["ast.Toplevel.Typealias"] = "Type",
     ["ast.Toplevel.Record"]    = "Type",
+    ["ast.Toplevel.Stat"]      = "Stat",
 }
 
 function Checker:check_program(prog_ast)
     assert(prog_ast._tag == "ast.Program.Program")
     local tls = prog_ast.tls
-    do
-        -- Forbid top-level duplicates
-        --
-        -- To avoid ambiguities that could happen if the programmer tried to declare
-        -- multiple toplevel entities with the same name, we give a compilation
-        -- error.
-        local names = {}
-        for _, top_level_node in ipairs(tls) do
-            local top_level_names = ast.toplevel_names(top_level_node)
-            local node_location = top_level_node.loc
-            for _, name in ipairs(top_level_names) do
-                local old_location = names[name]
-                if old_location then
-                    scope_error(node_location,
-                        "duplicate toplevel '%s', previous one at line %d",
-                        name, old_location.line)
-                end
-                names[name] = node_location
-            end
-        end
-    end
 
     -- Add most primitive types to the symbol table
     self:add_type("any",     types.T.Any())
@@ -264,6 +258,7 @@ function Checker:check_program(prog_ast)
             local tag1  = node1._tag
             assert(letrec_groups[tag1])
 
+            node1.i = i
             local group = { node1 }
             local j = i + 1
             while j <= N do
@@ -275,6 +270,7 @@ function Checker:check_program(prog_ast)
                     break
                 end
 
+                node2.i = j
                 table.insert(group, node2)
                 j = j + 1
             end
@@ -290,38 +286,6 @@ function Checker:check_program(prog_ast)
         if     group_kind == "Import" then
             local loc = tl_group[1].loc
             type_error(loc, "modules are not implemented yet")
-
-        elseif group_kind == "Var" then
-
-            for _, tl_var in ipairs(tl_group) do
-
-                self:expand_function_returns(tl_var.values)
-
-                for i, decl in ipairs(tl_var.decls) do
-                    tl_var.values[i] =
-                        self:check_initializer_exp(
-                            decl, tl_var.values[i],
-                            "declaration of module variable %s", decl.name)
-                end
-
-                for _, decl in ipairs(tl_var.decls) do
-                    self:add_global(decl)
-                end
-            end
-
-
-        elseif group_kind == "Func" then
-
-            for _, tl_func in ipairs(tl_group) do
-                local decl = tl_func.decl
-                decl._type = self:from_ast_type(decl.type)
-                self:add_function(decl)
-            end
-
-            for _, tl_func in ipairs(tl_group) do
-                tl_func.value =
-                    self:check_exp_verify(tl_func.value, tl_func.decl._type, "toplevel function")
-            end
 
         elseif group_kind == "Type" then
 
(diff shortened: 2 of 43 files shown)


pallene/checker.lua
```diff
@@ -349,14 +348,60 @@ function Checker:check_program(prog_ast)
...
     end
 
+    local total_nodes = #prog_ast.tls
+    local last_toplevel_node = prog_ast.tls[total_nodes]
+    if last_toplevel_node._tag ~= "ast.Toplevel.Stat" then
+        type_error(last_toplevel_node.loc, "Last Toplevel element must be a statement")
+    end
+    local last_stat = last_toplevel_node.stat
+    if last_stat._tag ~= "ast.Stat.Return" then
+        type_error(last_stat.loc, "Last Toplevel statement must be a return statement")
```
hugomg, line 378: Maybe we could merge these two error messages? In both cases we could say that the error is that the last thing in the toplevel must be a return statement.

pallene/checker.lua
```diff
@@ -370,7 +415,7 @@ function Checker:expand_function_returns(rhs)
     end
 end
 
-function Checker:check_stat(stat)
+function Checker:check_stat(stat, toplevel)
```
hugomg, line 418: 1. I suggest using the name "isXXX" for boolean variables
2. Consider using `check_stat(stat, false)` instead of just `check_stat(stat)`

pallene/checker.lua
```diff
@@ -512,6 +575,29 @@ function Checker:check_stat(stat)
...
+                    local mod_cname = self.symbol_table:find_symbol(top_var.name)
+                    if mod_cname and mod_cname._tag == "checker.Name.Module" then
+                        local field_name = stat.exps[i].var.name
+                        if not mod_cname.fields[field_name] then
+                            scope_error(stat.exps[i].loc,
+                              "variable '%s.%s' is not declared", top_var.name, field_name)
+                        end
+                    end
+                end
+            end
```
hugomg, line 590: Wen we type check an expression we convert qualified identifiers such as `io.write` from an `ast.Var.Dot` to an `ast.Var.Name`. We might want to do the same with module globals.

pallene/to_ir.lua
```diff
@@ -494,13 +445,30 @@ function ToIR:convert_stat(cmds, stat)
 
     elseif tag == "ast.Stat.Decl" then
         for _, decl in ipairs(stat.decls) do
-            self.loc_id_of_decl[decl] = ir.add_local(self.func, decl.name, decl._type)
+            if decl.visibility == "export" then
```
hugomg, line 448: In the previous version of the program, the `visibility` referred to the `export` or `local` keyword, which is something that we are trying to get rid of. Keeping the visibility concept in the new code seems suspicious.

pallene/types.lua
```diff
@@ -27,6 +27,7 @@ declare_type("T", {
         "field_names", -- same order as the source type declaration
         "field_types", -- map { string => types.T }
     },
+    Module  = {},
```
hugomg, line 30: Adding this case to the `Types.T` union makes me nervous. The fake "Void" type has already caused trouble for us and this would have the same pattern.

hugomg (changes requested):
Hi,

Thanks for tackling this problem. I haven't been able to look at everything in full detail but I think I can already leave some comments. It's a good place to start the discussion on what remains so that we can merge this.

pallene/types.lua
VictorNogueiraRio: I think I would need to bypass some checks made by the compiler, if I were to remove this, but I can try.

pallene/to_ir.lua
VictorNogueiraRio: I basically kept the same code generation logic, that is, `m.var` generates the exact same code as `export var` did. With that in mind, I thought I could keep the `visibility` variable also. However, if you think it wouldn't be wise to do so, I could check the `cname` tag and see if it's of type `checker.Mod_field.Var`. I believe both approaches would work.

pallene/checker.lua
VictorNogueiraRio: I may be spewing nonsense, because I don't know the code as much as you do, but after the `checker` phase, I believe the module name becomes irrelevant. I could transform it to an `ast.Var.Name`, excluding the module name. We could differentiate it from other `ast.Var.Name` variables by checking that its associated `cname` has the `checker.Mod_field.Var` tag.

pallene/checker.lua
VictorNogueiraRio: Ok, will do.

pallene/checker.lua
VictorNogueiraRio: Agreed.

pallene/checker.lua
```diff
@@ -349,14 +348,60 @@ function Checker:check_program(prog_ast)
...
+
         else
             error("impossible")
         end
     end
 
+    local total_nodes = #prog_ast.tls
+    local last_toplevel_node = prog_ast.tls[total_nodes]
+    if last_toplevel_node._tag ~= "ast.Toplevel.Stat" then
```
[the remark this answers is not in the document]
VictorNogueiraRio: 🤦 thanks.

pallene/checker.lua
```diff
@@ -349,14 +348,60 @@ function Checker:check_program(prog_ast)
...
 
+        elseif group_kind == "Stat" then
+            local tot_rets = 0
+            for _, tl_node in ipairs(tl_group) do
+                local stat = tl_node.stat
+                -- The toplevel return statement will be checked further on
+                if stat._tag == "ast.Stat.Return" then
+                    tot_rets = tot_rets + 1
+                    if tot_rets > 1 then
+                        type_error(stat.loc, "There can only be one toplevel return statement")
```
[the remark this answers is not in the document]
VictorNogueiraRio: Ok, will do.

pallene/checker.lua
```diff
@@ -114,13 +114,19 @@ local function declare_type(type_name, cons)
     typedecl.declare(checker, "checker", type_name, cons)
 end
 
+declare_type("Mod_field", {
+    Function = {"loc", "decl"},
+    Var      = {"loc", "name", "typ"}
```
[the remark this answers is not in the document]
VictorNogueiraRio: A could certainly create a "fake `Decl`", i.e., a table containing `type` and `name`, if you think it would be better.

pallene/ast.lua
```diff
@@ -88,18 +89,27 @@ declare_type("Field", {
...
+            end
+        elseif stat._tag == "ast.Stat.Func" then
+            local f_name = stat.name
+            local var = f_name.var
+            if var._tag ~= "ast.Var.Name" and var._tag ~= "ast.Var.Dot" then
+                error("impossible")
+            elseif var._tag == "ast.Var.Name" then
+                table.insert(names, var.name)
+            end
+        end
```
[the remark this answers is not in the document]
VictorNogueiraRio: I was using it to detect, for example, `local` functions and variables with the same name, or `typealiases` with the same name as `local`variables. If these duplicate names are being checked later in the code, then we really don't gain anything by keeping this function.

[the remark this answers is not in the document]
hugomg: It's OK if those things shadow, the same way as Lua. In fact, being more similar to Lua is one of the motivations for this rewrite...

The reason we currently have a no-shadowing rule is that Pallene to Lua translator works by adding a return statement to the end of the file:

    export function foo()
    end
    
    -- ...
    
becomes

   local function foo()
   end
   
   -- ...

    return { foo = foo }

However, if the `...` code has another definition of `foo`, it might shadow the definition we want to export, causing trouble. Anyway, the new syntax we're implementing should avoid this by making return statement explicit and getting rid of the magic.

pallene/checker.lua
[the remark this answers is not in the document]
hugomg: Using a real `ast.Decl` might also be an option. Keep in mind that in the parser.lua, the Decl we create for the toplevel functions also is an "imaginary" one, that doesn't exist in the original program text.

pallene/checker.lua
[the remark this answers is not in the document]
hugomg: To reduce the number of emails that github sends out, it's OK if you only add a thumbs up "reaction" instead of posting a full reply.

pallene/checker.lua
hugomg: > We could differentiate it from other ast.Var.Name variables by checking that its associated cname  tag.

Yes, this is the idea. Except that the tag we will want to check will probably be one of the cname tags. That is, `checker.Name.Local`, `checker.Name.Global`, etc.

The `_name` annotation is a `checker.Name` describing what the name means in the current scope. It should either directly contain the relevant information or point to somewhere that does (like the `decl`).

pallene/to_ir.lua
hugomg: Whatever we choose, the most important thing is to give it a clear name, and document it correctly.

Annotations that we add to nodes after the fact always start with an underscore (e.g. `_type` and `_name`) and the comment in the top of the checker.lua should explain what nodes receive the annotation and what it should contain.

Fields that are always part of the AST node should be declared in the  `ast.lua` and be initialized by the constructor of the node.

Currently, the `"export"` and `"local"` strings refer to the keywords that we use. If we keep some kind of "visibility" annotation, I think it might make more appropriate to use booleans instead.

pallene/types.lua
hugomg: Could you clarify what you mean by "need to bypass some  checks"?

VictorNogueiraRio: I believe "bypass" was a poor choice of verb. I meant that I'll need to add exceptions wherever a type is expected and we have a module variable. Putting it in the `Types.T` union simplified things a little, but I do see your concern and I also agree that it shouldn't be treated as a regular type. I'll remove it from the union and handle the exceptions, np.

pallene/checker.lua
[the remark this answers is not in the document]
VictorNogueiraRio: Sorry, I should've used clearer adjectives here, instead of "fake" and "real". I thought that using `ast.Decl` for something that doesn't correspond to a "declaration", in this case, an assignment, would be problematic (I actually thought about doing it). Since you think it's ok, I'm heavily considering doing so.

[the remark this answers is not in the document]
hugomg: You don't need to be sorry ;)

In the grammar, we use the Decl non-terminal for everything that is a name followed by an optional type annotation. That includes variable declarations and function arguments.

In the type checker we use references to these decl nodes to refer to the place where a given local variable is declared.

For toplevel declarations we also use Decls in some places but admittedly it's a bit more of a hack in this part.

pallene/types.lua
hugomg: The main thing that would worry me is code that might silently accept the fake module type. For example, `type.equals` might say that `module == module`. Another potential problem would be type inference; `local x = m` might infer `x` to have type `module` if `m` is a module reference.

But I agree 100% that keeping things simple is an important goal. One thing we could try is to raise an error as soon as we convert an `ast.Type` to a `types.T`. Only allow the "module" type annotation to be used for declaring a module reference in the toplevel.

pallene/to_ir.lua
```diff
@@ -520,6 +486,27 @@ function ToIR:convert_stat(cmds, stat)
     elseif tag == "ast.Stat.Break" then
         table.insert(cmds, ir.Cmd.Break())
 
+    elseif tag == "ast.Stat.Func" then
+        local prev_f_id = 1
```
hugomg, line 490: Maybe we should refactor the `enter_func` to allow entering another function while inside another function. When `enter_func` is called, take the current values of `self.func`, etc, that refer to the outer function and save them  to a stack. Then restore those variables when the inner function exits.

I really don't like that currently we have a `self:enter_function(prev_f_id)` without a matching `self:exit_function`.

hugomg (changes requested):
The `checker.Name.Module` is exactly the thing I was suggesting before. You might not have understood what I meant originally but you did end up doing what I asked :) That is, when the `_name` is a `checker.Name.Module` it means that this name refers to a module.

I don't understand, however, what is the purpose of the separate `checker.Module` type. What do the `checker.Module.Function` and the `checker.Module.Var` mean?

In particular, I think it is very strange that there are some if-elseif chains that mix the `checker.Name` and the `checker.Module` types:

    elseif cname._tag == "checker.Name.Global" or
             cname._tag == "checker.Module.Var" then

One important rule is that when we're pattern matching on a type tag all the cases should belong to the same type. The `cname` should always be a `checker.Name`.

Because I didn't fully understand what this checker.Module is for, I did not review the parts of the PR that are related to it, which includes a good part of the type checker. I'll review those again in the next round of review.

pallene/to_ir.lua
VictorNogueiraRio: Yes, I thought about that, but didn't want to perform yet another "big change", this PR seems big enough as is. However, I totally agree with you, this is a ghastly workaround.
Maybe I should refactor it on a separate commit.

pallene/parser.lua
```diff
@@ -277,6 +222,9 @@ function Parser:SimpleType()
 
     elseif self:peek("NAME") then
         local tok = self:e()
+        if tok.value == 'module' then
+            return ast.Type.Module(tok.loc)
+        end
```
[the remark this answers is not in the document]
VictorNogueiraRio: I'll try option 2

pallene/to_ir.lua
hugomg: If we aren't refactoring that on this PR, could you please create a separate issue for it, so that we don't forget?

VictorNogueiraRio:
> The `checker.Name.Module` is exactly the thing I was suggesting before. You might not have understood what I meant originally but you did end up doing what I asked :) That is, when the `_name` is a `checker.Name.Module` it means that this name refers to a module.
> 
> I don't understand, however, what is the purpose of the separate `checker.Module` type. What do the `checker.Module.Function` and the `checker.Module.Var` mean?

They tell me if the variable is a function or a variable. With this value, I know if, for example, I should extract the id from `fun_id_of_decl` or from `glb_id_of_decl`. 
There are also other places where I need to know if the variable is a function or just a variable.
What I did was try to mimic what you do with `checker.Name.Function` and `checker.Name.Global`.

> In particular, I think it is very strange that there are some if-elseif chains that mix the `checker.Name` and the `checker.Module` types:
> 
> ```
> elseif cname._tag == "checker.Name.Global" or
>          cname._tag == "checker.Module.Var" then
> ```
> 
> One important rule is that when we're pattern matching on a type tag all the cases should belong to the same type. The `cname` should always be a `checker.Name`.

I see.
I can also store a `checker.Name.Module` in the `name` field and see if it's a function or a variable using another approach, it's just slightly messier.

pallene/to_ir.lua
VictorNogueiraRio: @hugomg https://github.com/pallene-lang/pallene/issues/363

pallene/parser.lua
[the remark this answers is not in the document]
VictorNogueiraRio: I'm answering here because github wouldn't let me answer as a reply to the original comment (https://github.com/pallene-lang/pallene/pull/361#discussion_r617976829).
Yes, we can delete `ast.toplevel_names`.
Sorry, I thought I had deleted it already.

hugomg:
> They tell me if the variable is a function or a variable.

That is indeed the purpose of the `_name` annotation. However, it should always be a `checker.Name`, not some other kind of object. Currently we use:

- Checker.Name.Local: local variables (declared inside a function)
- Checker.Name.Global: toplevel variables (declared outside a function)
- Checker.Name.Function: toplevel functions (declared in this module)
- Checker.Name.Builtin: functions from the standard library

It might require some minor tweaking once we add functions from modules that we required.

> I can also store a checker.Name.Module in the _name field and see if it's a function or a variable using another approach, it's just slightly messier.

Maybe we need a concrete example to help explain whats going on... I don't understand what you mean here. If the `_name` field is a checker.Name.Module it should mean that this identifier refers to a module. It can't also be a function or variable at the same time.

pallene/checker.lua
```diff
@@ -349,14 +306,60 @@ function Checker:check_program(prog_ast)
...
     end
 
+    local total_nodes = #tls
+    if total_nodes == 0 then
+        return prog_ast
+    end
+
+    for i = 1, total_nodes - 1 do
+        local stat = tls[i].stat
+        if stat then
```
hugomg, line 324: This `if` statement should check the `_tag` of the `tls[i]` instead of testing if the `stat` field is present.

pallene/checker.lua
```diff
@@ -349,14 +306,60 @@ function Checker:check_program(prog_ast)
...
+            -- skip
+
         else
             error("impossible")
         end
     end
 
+    local total_nodes = #tls
+    if total_nodes == 0 then
+        return prog_ast
```
hugomg, line 319: Should empty modules be allowed or should they be an error? If it were an error it would be one less corner case to worry about. What do you think?

pallene/checker.lua
```diff
@@ -135,22 +137,29 @@ end
 
 function Checker:add_global(decl)
     assert(decl._tag == "ast.Decl.Decl")
-    self.symbol_table:add_symbol(decl.name, checker.Name.Global(decl))
+    local name = decl._modname and (decl._modname .. '.' .. decl.name) or decl.name
+    local cname = self.symbol_table:find_symbol(name)
```
hugomg, line 141: This part where we  concatenate the modname caught my attention. My gut feeling is that instead of adding a `m.foo` item to the symbol table, the more natural thing to do would be to add the `m` module to the symbol table and then any use of `m.foo` would look for `foo` inside the module. The idea is that the name that's actually in scope is the `m`, not the `m.foo`. That said, if the current solution with the modname is working it's OK to keep it like this for now. I can always try changing it later if necessary. I'm just saying to keep an eye on this part of the design.

pallene/checker.lua
```diff
@@ -549,19 +598,43 @@ function Checker:check_stat(stat)
...
-        self:check_stat(stat.then_)
-        self:check_stat(stat.else_)
+        self:check_stat(stat.then_, false)
+        self:check_stat(stat.else_, false)
 
     elseif tag == "ast.Stat.Break" then
         -- ok
 
+    elseif tag == "ast.Stat.Func" then
+
```
hugomg, line 608: (extra whitespace)

pallene/checker.lua
```diff
@@ -549,19 +598,43 @@ function Checker:check_stat(stat)
...
+        local decl = stat.decl
+        local fname_exp = stat.name
+        decl._type = self:from_ast_type(decl.type)
+        self:check_funcname(fname_exp, decl)
+        stat.value =
+            self:check_exp_verify(stat.value, decl._type, "toplevel function")
```
hugomg, line 614: Does this fit in 100 columns without the line break?

pallene/checker.lua
```diff
@@ -549,19 +598,43 @@ function Checker:check_stat(stat)
...
 
     return stat
 end
 
+function Checker:check_funcname(name, decl)
+    local var = name.var
+    local tag = var._tag
+    decl.name = var.name
+    if tag == "ast.Var.Dot" then
+        local mod_cname = self.symbol_table:find_symbol(var.exp.var.name)
```
hugomg, line 628: Be careful when you access a deep sequence of fields like `var.exp.var.name`. We need to check that the `var.exp` is an `ast.Exp.Var` before we can access the `var.exp.var`. We need to check that this last var is an `ast.Var.Name` before we read the `.name` field.

As a rule of thumb, before we use any datatype that is a tagged union, we need to check the `_tag` with either an if statement or an `assert`. Whichever is more appropriate.

hugomg (changes requested):
Looks good so far. We can start working on fixing the broken tests.

pallene/checker.lua
VictorNogueiraRio: Your idea is what I originally implemented, with the `fields` table being an element of the module symbol. The difference was that I inserted the `field` type in the `_name` value. I can try and go back to the previous version

pallene/checker.lua
VictorNogueiraRio: I don't think it does.

hugomg:
Great! Now that all the tests are passing, I'm open to merging this. It'll be easier for others (including myself) to contribute once it's merged to the main branch.

What do you think about rebasing this on top of the latest master and then marking the PR as ready for review (no longer a draft)?

VictorNogueiraRio:
> Great! Now that all the tests are passing, I'm open to merging this. It'll be easier for others (including myself) to contribute once it's merged to the main branch.
> 
> What do you think about rebasing this on top of the latest master and then marking the PR as ready for review (no longer a draft)?

If it's ok to you, I'd rather create more tests (regarding module variables) before making it ready for review.
I'd like to certify that everything is working as should be before making it ready for review.

hugomg:
Sure, feel free to add more tests. But don't feel obliged to have a super perfect PR before we merge it. It's OK to split it over more than one PR.

hugomg (approved):

hugomg:
Good tests!

Issue #348, opened 2021-01-16
Better error messages for loops with ipairs.


srijan-paul:
Currently, an `ipairs` loop with incorrect number of decls (`for i in ipairs(xs)`) throws an error like so:
```
type error: expected function type (any, any) -> (any) but found function type (any, any) -> (any, any) in loop iterator
```
This is rather confusing as it attributes the error to the iterator function which isn't immediately clear to the user.
It would be better to have an error message that clear says something along the lines of `Expected 2 declarations in loop but found only 1`.

hugomg:
Fixed in #364

PR #364, opened 2021-04-27
Added appropriate error messages for for-in loops.

srijan-paul:
This PR addresses #348, and issue which was left over after the introduction of `ipairs` and for-in loops.
Instead of throwing a confusing error message like so: 
```
type error: expected function type (any, any) -> (any) but found function type (any, any) -> (any, any) in loop iterator
```
The checker now throws a more comprehensible error.
```
expected 2 declarations in for loop but found 1. 
```

Currently this error is thrown for both `ipairs` and non-ipairs for-in loops.
If the return count of the iterator function doesn't match the number of declarations in the loop LHS, a shorter error is thrown as
demonstrated above.

For all other kinds of mismatches (incorrect arg count, types), the usual type mismatch error is thrown.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -471,8 +471,8 @@ function Checker:check_stat(stat)
         if not types.equals(iteratorfn._type, itertype) then
             if iteratorfn._type._tag == "types.T.Function" and
                #decl_types ~= #iteratorfn._type.ret_types then
-                local decl_count            = #decl_types
-                local expected_decl_count   = #iteratorfn._type.ret_types
+                local decl_count          = #decl_types
+                local expected_decl_count = #iteratorfn._type.ret_types
 
                 if expected_decl_count == 1 then
                     type_error(iteratorfn.loc, "expected 1 declaration in for loop but found %d.",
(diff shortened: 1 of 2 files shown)


pallene/checker.lua
```diff
@@ -459,18 +459,32 @@ function Checker:check_stat(stat)
...
+               #decl_types ~= #iteratorfn._type.ret_types then
+                local decl_count          = #decl_types
+                local expected_decl_count = #iteratorfn._type.ret_types
+
+                if expected_decl_count == 1 then
+                    type_error(iteratorfn.loc, "expected 1 declaration in for loop but found %d.",
+                        decl_count)
+                else
+                    type_error(iteratorfn.loc, "expected %d declarations in for loop but found %d.",
+                        expected_decl_count, decl_count)
```
hugomg, line 482: I'd prefer if the program flow did not depend on the pluralization rules for English. What if we used a message that works both for the singular and plural case?

    expected %d variable(s)

hugomg (changes requested):
These new error messages look much nicer! Thanks for improving them.

By the way, you don't have to open a PR and issue at the same time. In cases like this it's OK to just have the PR by itself.

pallene/checker.lua
srijan-paul: Ah. I noticed it would also be more consistent that way since the function call arity errors are similarly worded.

srijan-paul:
>  In cases like this it's OK to just have the PR by itself.

Understood ^^

hugomg (approved):
Perfect. Thanks again!

PR #366, opened 2021-05-20
Allow module names to be used as type names

hugomg:
This removes a special case for the "string" name, which was present in the scope rules.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -122,7 +122,7 @@ declare_type("Name", {
     Global   = { "decl" },
     Function = { "decl" },
     Builtin  = { "name" },
-    Module   = { "name", "is_main_mod", "shadowed_type"}
+    Module   = { "name", "is_main_mod", "shadowed_symbol"}
 })
 -- TODO: add a comment explaining the Module case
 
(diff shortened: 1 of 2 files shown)


pallene/checker.lua
```diff
@@ -183,13 +196,16 @@ function Checker:from_ast_type(ast_typ)
...
+            if cname.shadowed_type then
+                return cname.shadowed_type.typ
+            else
+                -- fallthrough
+            end
+        else
+            -- fallthrough
         end
-        type_error(ast_typ.loc, "'%s' isn't a type", name)
+        type_error(ast_typ.loc, "'%s' is not a type", name)
```
VictorNogueiraRio, line 208: I'm not against this change, but I'm curious as to why it's necessary.

pallene/checker.lua
```diff
@@ -164,7 +165,19 @@ end
...
+    -- To allow the name "string" to be used both as a module name and as a type name, module
+    -- symbols are aware of the type that they are shadowing.
+    local curr = self.symbol_table:find_symbol(name)
+
+    local shadowed_type
+    if curr and curr._tag == "checker.Name.Type" then
+        shadowed_type = curr
+    else
+        shadowed_type = false
+    end
```
VictorNogueiraRio, line 178: Would this also work?
```suggestion
    local shadowed_type = curr and curr._tag and curr.typ
```

If so, I think it's more direct, because you are only storing the necessary information.

pallene/checker.lua
hugomg: The contraction was more informal.

VictorNogueiraRio (approved):

PR #368, opened 2021-05-20
Remove local m boilerplate from some tests

hugomg:
In some test cases, move the `local m={}` to the default test template, so that we don't have to repeat it every time.

There are some test files that still keep the `local m={}`. I'll admit that the choice was somewhat arbitrary.

In the parser, I did not add the local m on purpose, because it would complicate the generated AST. The other ones that I didn't change, one of the main reasons are that they are small files with less test cases in the file.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -151,7 +151,7 @@ describe("Scope analysis: ", function()
             local x: integer = 0
             return m
         ]],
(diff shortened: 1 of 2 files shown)


VictorNogueiraRio (approved):
Should've thought of that whilst adapting the tests.
Would've saved me a lot of time.
🤦

PR #373, opened 2021-05-24
Parsing closures

srijan-paul:
With this PR, the parser can process functions in both statement and expression contexts.

- The checker just croaks with a `"Not implemented"` error when type-checking function expressions.
- The type annotation nodes for the functions are now stored in `ast.Exp.Lambda`, and the `decl` nodes in `ast.Stat.Func` simply reference this node.
-  Function expressions are now considered under the `SimpleExp` derivation.
- The new `FuncExp` function starts at the opening `(` and parses the entire function body, this is used by both `Func` and `SimpleExp` to parse function statements and function expressions respectively. (should there be a comment documenting this?)

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -1064,8 +1064,8 @@ function Checker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
         -- These assertions are always true in the current version of Pallene, which does not allow
         -- nested function expressions. Once we add function expressions to the parser then we
         -- should convert these assertions into proper calls to type_error.
-        assert(expected_type._tag == "types.T.Function")
-        assert(#expected_type.arg_types == #exp.arg_decls)
+        assert(expected_type._tag == "types.T.Function", "function expressions not implemented yet")
+        assert(#expected_type.arg_types == #exp.arg_decls, "function expressions not implemented yet")
 
         table.insert(self.ret_types_stack, expected_type.ret_types)
         self.symbol_table:with_block(function()
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -494,10 +494,9 @@ function Parser:Block()
     return ast.Stat.Block(false, self:StatList())
 end
 
-function Parser:Func(is_local)
-    local start    = self:e("function")
+function Parser:FuncStat(is_local)
+    local start = self:e("function")
 
-    -- Function name
     local root = self:e("NAME")
 
     local fields = {}
@@ -516,9 +515,7 @@ function Parser:Func(is_local)
         self:syntax_error(root.loc, "Local function name has a '.' or ':'")
     end
 
-    local oparen   = self:e("(")
-    local params   = self:DeclList()
-    local _        = self:e(")", oparen)
+    local params = self:FuncParams()
 
     local return_types = {}
     if self:peek(":") then
@@ -629,7 +626,7 @@ function Parser:Stat(is_toplevel)
     elseif self:peek("local") then
         local start = self:e()
         if self:peek("function") then
-            return self:Func(true)
+            return self:FuncStat(true)
         else
             local decls = self:DeclList(); if #decls == 0 then self:forced_syntax_error("NAME") end
             local exps  = self:try("=") and self:ExpList1() or {}
@@ -653,7 +650,7 @@ function Parser:Stat(is_toplevel)
         end
 
     elseif self:peek("function") then
-        return self:Func(false)
+        return self:FuncStat(false)
 
     else
         -- Assignment or function call
@@ -760,6 +757,33 @@ function Parser:FuncArgs()
     end
 end
 
+function Parser:FuncParams()
+    local oparen = self:e("(")
+    local params = self:DeclList()
+    local _ = self:e(")", oparen)
+    return params
+end
+
+function Parser:FuncExp(func_token)
+    local params = self:FuncParams(true)
+
+    for _, decl in ipairs(params) do
+        if decl.type then
+            self:syntax_error(decl.loc, "Function expressions cannot be type annotated")
+        end
+    end
+
+    if self:peek(":") then
+        local colon = self:e()
+        self:syntax_error(colon.loc, "Function expressions cannot be type annotated")
+    end
+
+    local block = self:Block()
+    local _     = self:e("end", func_token)
+
+    return ast.Exp.Lambda(func_token.loc, params, block)
+end
+
 function Parser:SimpleExp()
     if     self:peek("NUMBER") then
         local id = self:e()
@@ -796,6 +820,9 @@ function Parser:SimpleExp()
         self:e("}", open)
         return ast.Exp.Initlist(open.loc, fields)
 
+    elseif self:peek("function") then
+        local start = self:e()
+        return self:FuncExp(start)
     else
         return self:SuffixedExp(false)
     end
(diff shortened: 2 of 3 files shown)


pallene/checker.lua
```diff
@@ -838,7 +838,7 @@ function Checker:check_exp_synthesize(exp)
         type_error(exp.loc, "missing type hint for initializer")
 
     elseif tag == "ast.Exp.Lambda" then
-        type_error(exp.loc, "missing type hint for lambda")
+        type_error(exp.loc, "higher order functions not implemented.")
```
hugomg, line 841: This "noy yet implemented" error should go into the `check_exp_verify`. 

For `check_exp_synthesize` the current error is correct. It is not possible to typecheck a lambda in isolation; we must always know the expected type first. (This is similar issue to what happens in table literals).

pallene/ast.lua
```diff
@@ -63,7 +63,7 @@ declare_type("Exp", {
     Float      = {"loc", "value"},
     String     = {"loc", "value"},
     Initlist   = {"loc", "fields"},
-    Lambda     = {"loc", "arg_decls", "body"},
+    Lambda     = {"loc", "arg_decls", "type", "body"},
```
hugomg, line 66: Lambdas should not carry a type. Don't be destracted by that evil decl that we used to add to give to the functions. It's no longer present after the refactoring, for good reason :)

spec/parser_spec.lua
```diff
@@ -432,6 +432,24 @@ describe("Pallene parser", function()
...
+            arg_decls = {},
+            body = {
+              _tag  = "ast.Stat.Block",
+              stats = { {
+                _tag = "ast.Stat.Return",
+                exps = { { _tag = "ast.Exp.Integer", value = 1 } } }
+              }
+            }
+        })
+    end)
```
hugomg, line 447: Can we add some function arguments to this test case?

pallene/parser.lua
```diff
@@ -589,6 +562,39 @@ function Parser:FuncArgs()
...
+    end
+
+    local block    = self:Block()
+    local _        = self:e("end", func_token)
+
+    local arg_types = {}
+    for i, decl in ipairs(params) do
+        if not decl.type then
+            self:syntax_error(decl.loc,
+                    "Parameter '%s' is missing a type annotation", decl.name)
```
hugomg, line 589: We should consider removing this restriction from the parser. It might be useful to allow lambdas without a type annotation, if they are used in a context where the type is known:

    local f : integer->integer = function(x) return x+ 1 end
   
    local function map(xs: {any}, f: any->any): {any}
        -- ...
    end
    map(my_list, function(x) return x end)

That said, this change doesn't need to be right now.

hugomg (changes requested):

pallene/parser.lua
srijan-paul: just to make sure, this is only for lambdas and not for function statements right?

pallene/ast.lua
srijan-paul: I ended up removing that, but now a call to `FuncExp` returns a lambda + type.
```lua
 return ast.Exp.Lambda(name_exp.loc, params, block), func_typ
```
Since I'm not supposed to add new ast nodes in this PR, is it okay to not carry types in a lambda?
I wonder how the type checker will access the type tags for function expressions that are just an `ast.Exp.Lambda` nodes by themselves. (Since the only other node representing a function is in `ast.Stat`).

pallene/parser.lua
hugomg: Yes. Perhaps that error checking shoulg be just on the fn statements.

pallene/ast.lua
hugomg: A lambda by itself cannot be type checked, similarly to how a table literal by itself cannot be type checked. This is why check exp synthesise raises an error if it hits a lambda.

A lambda always has to appear in a context where the expected type type is known.  (Check exp verify)

srijan-paul: Aah, so the annotation is used. Gotcha!

pallene/checker.lua
```diff
@@ -1137,8 +1137,8 @@ function Checker:check_exp_verify(exp, expected_type, errmsg_fmt, ...)
         -- These assertions are always true in the current version of Pallene, which does not allow
         -- nested function expressions. Once we add function expressions to the parser then we
         -- should convert these assertions into proper calls to type_error.
-        assert(expected_type._tag == "types.T.Function")
-        assert(#expected_type.arg_types == #exp.arg_decls)
+        assert(expected_type._tag == "types.T.Function" and #expected_type.arg_types == #exp.arg_decls,
+            "function expressions not implemented yet")
```
hugomg, line 1141: We prefer putting each check on a separate assertion, because if the assertion fails then the error message tells which line it was.

pallene/parser.lua
hugomg: Instead of having an is_stat argument, we could put this error checking in the `Func` (which we should probably also rename to `FuncStat`, by the way). After the `FuncExp` returns, iterate over the returned expression to see if if all the decls have the required type annotation.

pallene/parser.lua
```diff
@@ -589,6 +562,39 @@ function Parser:FuncArgs()
...
+    for i, decl in ipairs(params) do
+        if not decl.type and is_stat then
+            self:syntax_error(decl.loc,
+                    "Parameter '%s' is missing a type annotation", decl.name)
+        end
+      arg_types[i] = decl.type
+    end
+
+    local func_typ = ast.Type.Function(name_exp.loc, arg_types, rt_types)
+    return ast.Exp.Lambda(name_exp.loc, params, block), func_typ
```
hugomg, line 595: Do we need to receive name_exp as a parameter? We could use the func_token to get the location

hugomg (changes requested):

pallene/parser.lua
srijan-paul: I kept it so that the function's name is used for error reporting when parsing function statements. Should I change it?

hugomg: I think this is another reason why we should put the error logic in the FuncStat.

srijan-paul: Ah, okay.

pallene/parser.lua
```diff
@@ -577,25 +592,30 @@ function Parser:FuncParams()
...
+    local params = {}
 
-    local block    = self:Block()
-    local _        = self:e("end", func_token)
+    if self:peek("NAME") then
+        repeat
+            local id = self:e("NAME")
+            if self:peek(":") then
+                local colon = self:e()
+                self:syntax_error(colon.loc, "Function expressions cannot be type annotated")
```
hugomg, line 603: Would it be possible to implement this using `FuncParams()`? Parse the function parameters and then loop over the result to check if the type is present or not (similarly to what we did in the FuncStat).

hugomg (changes requested):

pallene/parser.lua
srijan-paul: Oh, yeah that's a great idea. done :+1:

hugomg (approved):
Great!

I'll approve this but not merge it yet. I would like to merge #374 first.

hugomg:
As expeced, there are some merge conflicts with #374. In theory they shouldn't be too hard to fix.

srijan-paul:
> As expeced, there are some merge conflicts with #374. In theory they shouldn't be too hard to fix.

True, I'll try to figure them out.

hugomg (approved):

PR #374, opened 2021-05-24
Refactor the type checker & Lua translator

hugomg:
1. The function name is now stored as part of the Stat.Func
2. Modules contain child symbol table. We no longer put all the builtin functions in the parent namespace.
3. Split the checker.Name into two separate types: checker.Symbol and checker.Def. This better separates the concerns of the type checker and the IR generator.
4. Changed the implementation of recursive functions. This part might deserve some extra love & code cleanup.
5. Local mutually recursive functions now use a forward declaration syntax.
6. Removed all traces of types.T.Module
7. Updated the translator to the new syntax
8. Moved the `local m: module` handling to the parser;
9. The `: module` pseudo type annotation is now optional.
10. Move the recursive function treatment into the parser.


hugomg:
One thing that I still need to double check is if the test suite is covering all the new error messages that were introduced in this PR.

hugomg:
@srijan-paul @VictorNogueiraRio if one of you could take a quick look at this PR before we merge it, I would be very grateful. Sometimes the code has something silly like a leftover printf statement, which a second pair of eyes helps to catch.

pallene/parser.lua
```diff
@@ -119,12 +119,83 @@ end
...
+    -- module contents
     local tls = {}
-    while not self:peek("EOF") do
+    while not self:peek("EOF") and not self:peek("return") do
         table.insert(tls, self:Toplevel())
     end
-    return ast.Program.Program(self.lexer:loc(), tls, self.type_regions,
-                                self.comment_regions)
+
+    -- returm <modname>
```
srijan-paul, line 166: I think this might be a small typo? `return <modname>`

pallene/checker.lua
```diff
@@ -11,30 +11,36 @@ local typedecl = require "pallene.typedecl"
...
+-- It produces a modified AST that is annotated with the following information:
 --
 --   * _type: A types.T in the following kinds of nodes
 --      - ast.Exp
 --      - ast.Var
 --      - ast.Decl
 --      - ast.Toplevel.Record
+--      - ast.Stat.Func
+--
+--   * _def: A checker.Def that describes the meaninge
```
srijan-paul, line 27: *meaning

srijan-paul:
> 
> 
> @srijan-paul @VictorNogueiraRio if one of you could take a quick look at this PR before we merge it, I would be very grateful. Sometimes the code has something silly like a leftover printf statement, which a second pair of eyes helps to catch.

Looks good overall to me. Thought we could wait to see what @VictorNogueiraRio has to say.

hugomg:
If it doesn't look immediately wrong, I think I'm going to merge it. I don't want to continue blocking the other PRs.

VictorNogueiraRio:
> If it doesn't look immediately wrong, I think I'm going to merge it. I don't want to continue blocking the other PRs.

I know it won't mean much now, but I've taken a look at it, and it seemed ok, at least at a first glance.

hugomg:
That's still good to know! Thanks.

PR #383, opened 2021-05-30
Also run the linter in the ./test-project

hugomg:
There have many times where I only found out about a lint error in the CI step, because I had forgotten to run the ./lint-project in addition to the ./test-project


srijan-paul:
Looks good.
Does this mean prior to this PR, no linting was done in `spec/`?

srijan-paul (approved):

hugomg:
The linting script was already linting the spec directory.

The problem is that if you wanted to fully test things locally, you had to run both the test script (./test-project) and the lint script (./lint-project). What ended up happening is that I'd often forget to call the linting script 😅

PR #388, opened 2021-06-06
Cleanup the multiple_definitions_error

hugomg:
If we use the symbol table as the single source of truth, we don't need to special-case the multiple assignments.

### DIFF
--- a/pallene/checker.lua
+++ b/pallene/checker.lua
@@ -27,7 +27,7 @@ local util = require "pallene.util"
 --   * _def: A checker.Def that describes the meaning of that name
 --      - ast.Var.Name
 --
---   * _is_exported: A boolean (true) marking assignments that are declaring an exported variable
+--   * _exported_as: A string telling that this var is exported, and which module field name.
 --      - ast.Var.Name
 --
 -- We also make some adjustments to the AST:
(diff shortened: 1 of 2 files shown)


pallene/checker.lua
```diff
@@ -507,18 +522,17 @@ function Checker:check_stat(stat, is_toplevel)
...
             if var._tag == "ast.Var.Dot" and self:is_the_module_variable(var.exp) then
                 -- Declaring a module field
                 if not is_toplevel then
                     type_error(var.loc, "module fields can only be set at the toplevel")
                 end
-                if self.module_symbol.symbols[var.name] or declarations[var.name] then
-                    multiple_definitions_error(var.loc, var.name)
-                end
-                declarations[var.name] = i
+                local qvar = ast.Var.Name(var.loc, var.name)
```
srijan-paul, line 531: What does `qvar` mean?

hugomg: That `qvar` name is referencing something that no longer exists. I'll change it.

When I was working on that checker refactoring that introduced the Defs, I experimented with having a new kind of `Var` called `ast.Var.QualifiedName`. However, I ended up not doing that, because the `Def` already served that purpose.

srijan-paul (approved):

PR #389, opened 2021-06-08
Implemented non-mutating closures

srijan-paul:
This PR adds support for closures that capture, but do not re-assign to their up-values.

- `ir.Function` now has a `captured_vars` (`{ integer -> ir.VarDecl }`)  and a `id_of_upvalue` (`{ integer -> integer }`) field.
    - `captured_vars` maps a local var id (which is really an upvalue) to it's `decl`
    - `id_of_upvalue` maps the local id of the upvalue to it's local id in the surrounding function.
    
- Here is why I thought the above was necessary:
    1. In `to_ir`, If we tried to assign consecutive local var ids to the arguments and the upvalues, we would have to preprocess the number of upvalues and reserve that many id slots in `loc_id_of_decl` and `ir.Function.vars` before we start registering the local variables.
    2. But since we're registering local variables and upvalues on the go, we may encounter an upvalue for the first time while having assigned some local var IDs already. So naturally we assign the next free local var id (`#local_vars + 1`) to the current upvalue.
    3. This means upvalues may not have consecutive IDs after parameters in the generated C code.
    4. To prevent multiple *declarations* of upvalues in the C code (once as a parameter, and once as a local variable), we make sure `captured_vars[id]` is `nil` before emitting the declaration code for `vars[id]`. (Because upvalues are not present one aftr the other following the parameter list).
    5. When capturing an upvalue (by adding it to `ccl->upvalue`), we need to know the current location of an upvalue in the C stack (it's var id), the `id_of_upvalue` helps with that.

- In `to_ir`, now whenever an identifier is encountered, the order of lookup is:
    1. Local var
    2. Captured var
    3. Global var
    
- In `coder`, some adjustments had to be made in the lua entry point code gen (to add code for capturing the vars), gen_cmd for `NewClosure` (to capture the upvalues) and the pallene entry point (to add the upvalues to the param list.)

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -300,6 +300,11 @@ function Coder:c_var(v_id)
     return "x" .. v_id
 end
 
+-- @returns the C parameter name for the upvalue u_id
+function Coder:c_upval(u_id)
+    return "uv" .. u_id
+end
+
 -- @returns the C return variable name for the variable ret_i
 function Coder:c_ret_var(ret_i)
     return "ret" .. ret_i
@@ -321,6 +326,8 @@ function Coder:c_value(value)
         return lua_value(types.T.String(), self:string_upvalue_slot(str))
     elseif tag == "ir.Value.LocalVar" then
         return self:c_var(value.id)
+    elseif tag == "ir.Value.Upvalue" then
+        return self:c_upval(value.id)
     elseif tag == "ir.Value.Function" then
         local f_id = value.id
         local typ = self.module.functions[f_id].typ
@@ -337,6 +344,14 @@ function Coder:prepare_local_var(func, v_id)
     local c_name = self:c_var(v_id)
     local typ    = func.vars[v_id].typ
     local p_name = func.vars[v_id].name
+    return typ, c_name, (p_name and " "..C.comment(p_name) or "")
+end
+
+-- The information of a C local var for a given Pallene captured var.
+function Coder:prepare_captured_var(u_id, decl)
+    local c_name = self:c_upval(u_id)
+    local typ    = decl.typ
+    local p_name = decl.name
     return typ, c_name, (p_name and " "..C.comment(p_name) or "")
 end
 
@@ -369,6 +384,11 @@ function Coder:pallene_entry_point_declaration(f_id)
     table.insert(args, {"lua_State *" , "L",    ""})
     table.insert(args, {"Udata *"     , "G",    ""})
     table.insert(args, {"StackValue *", "base", ""})
+
+    for i, upval in ipairs(func.captured_vars) do
+        local typ, c_name, comment = self:prepare_captured_var(i, upval.decl)
+        table.insert(args, {ctype(typ), c_name, comment})
+    end
     for i = 1, #arg_types do
         local v_id = ir.arg_var(func, i)
         local typ, c_name, comment = self:prepare_local_var(func, v_id)
@@ -497,6 +517,7 @@ function Coder:lua_entry_point_definition(f_id)
     local fname = func.name
     local arg_types = func.typ.arg_types
     local ret_types = func.typ.ret_types
+    local captured_vars = func.captured_vars
 
     -- We unconditionally initialize the G userdata here, in case one of the tag checking tests
     -- needs to use it. We don't bother to make this initialization conditional because in the case
@@ -519,16 +540,32 @@ function Coder:lua_entry_point_definition(f_id)
 
     local arg_vars  = {}
     local arg_decls = {}
+    for i, captured_var in ipairs(captured_vars) do
+        local name = self:c_upval(i)
+        table.insert(arg_vars, name)
+        table.insert(arg_decls, C.declaration(ctype(captured_var.decl.typ), name)..";")
+    end
     for i, typ in ipairs(arg_types) do
-        local name = "x"..i
-        arg_vars[i] = name
-        arg_decls[i] = C.declaration(ctype(typ), name)..";"
+        local name = self:c_var(i)
+        table.insert(arg_vars, name)
+        table.insert(arg_decls, C.declaration(ctype(typ), name)..";")
     end
 
+
     local init_args = {}
+    for u_id, upval in ipairs(captured_vars) do
+        local name = upval.decl.name
+        local typ  = upval.decl.typ
+        local dst  = arg_vars[u_id]
+        local src  = string.format("&func->upvalue[%s]", C.integer(u_id - 1))
+        table.insert(init_args,
+            self:get_stack_slot(typ, dst, src,
+                func.loc, "upvalue '%s'", C.string(name)))
+    end
+
     for i, typ in ipairs(arg_types) do
         local name = func.vars[i].name
-        local dst = arg_vars[i]
+        local dst = arg_vars[#captured_vars + i]
         local src = string.format("s2v(base + %s)", C.integer(i))
         table.insert(init_args,
             self:get_stack_slot(typ, dst, src,
@@ -1291,15 +1328,29 @@ gen_cmd["SetField"] = function(self, cmd, _func)
 end
 
 gen_cmd["NewClosure"] = function (self, cmd, _func)
+    local func = self.module.functions[cmd.f_id]
+
+    local capture_upvalues = {}
+    for i, upval_info in ipairs(func.captured_vars) do
+        local typ   = upval_info.decl.typ
+        local c_val = self:c_value(upval_info.value)
+        table.insert(capture_upvalues, set_stack_slot(typ,
+            string.format("&ccl->upvalue[%s]", C.integer(i - 1)), c_val))
+    end
+
     return util.render([[
         {
-            CClosure *ccl = luaF_newCclosure(L, 0);
+            CClosure *ccl = luaF_newCclosure(L, $num_upvalues);
             ccl->f = $lua_entry_point;
+            $capture_upvalues
+
             setclCvalue(L, &$dst, ccl);
         }
     ]], {
+        num_upvalues = C.integer(#func.captured_vars),
         dst = self:c_var(cmd.dst),
-        lua_entry_point = self:lua_entry_point_name(cmd.f_id)
+        lua_entry_point = self:lua_entry_point_name(cmd.f_id),
+        capture_upvalues = table.concat(capture_upvalues, "\n")
     })
 end
 
--- a/pallene/gc.lua
+++ b/pallene/gc.lua
@@ -47,6 +47,7 @@ function gc.compute_stack_slots(func)
     local last_use          = {} -- { var_id => integer }
     local first_definition  = {} -- { var_id => integer }
 
+    -- Upvalues can always be reached by the GC via their closures.
     for i, cmd in ipairs(flat_cmds) do
         for _, val in ipairs(ir.get_srcs(cmd)) do
             if val._tag == "ir.Value.LocalVar" then
--- a/pallene/ir.lua
+++ b/pallene/ir.lua
@@ -43,13 +43,21 @@ function ir.VarDecl(name, typ)
     }
 end
 
+function ir.UpvalInfo(decl, val)
+    return {
+        decl = decl, -- ir.VarDecl
+        value = val, -- ir.Value
+    }
+end
+
 function ir.Function(loc, name, typ)
     return {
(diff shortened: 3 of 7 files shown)


pallene/ir.lua
```diff
@@ -45,11 +45,13 @@ end
...
-        name = name,         -- string
-        typ = typ,           -- Type
-        vars = {},           -- list of ir.VarDecl
-        body = false,        -- ir.Cmd
+        loc = loc,          -- Location
+        name = name,        -- string
+        typ = typ,          -- Type
+        vars = {},          -- list of ir.VarDecl
+        captured_vars = {}, -- { integer => ast.Decl }
+        id_of_upval = {},   -- { integer => integer }
```
hugomg, line 53: Let's use a separatge list of upvalues, instead of reserving space in the local variables

pallene/to_ir.lua
```diff
@@ -98,8 +127,13 @@ function ToIR:is_local(decl)
             assert(not glb_id)
             return true, loc_id
         else
-            assert(glb_id)
-            return false, glb_id
+            local upval_id = self:resolve_upvalue(decl)
+            if upval_id then
+                return true, upval_id
```
hugomg, line 132: Is this still "true", now that upvalues are no lonver local variables?

pallene/to_ir.lua
```diff
@@ -90,6 +91,34 @@ end
 --
 --
 --
+function ToIR:resolve_upvalue(decl, func_index)
```
hugomg, line 94: Please add a comment to this function

hugomg (changes requested):

srijan-paul:
Upvalues are now stored separately from local variables.
The only thing that's different from what we discussed is still `ir.captured_vars`. Instead of being a list of `ir.VarDecl`, it is a list of `ir.VarDecl` + the id. This makes it convenient to generate the parameter list when emitting code for the pallene entry point's declaration.

hugomg (changes requested):
I'm very happy that we only required minimal changes to the coder. That is a good sign that we chose the right design!

pallene/gc.lua
```diff
@@ -49,7 +49,7 @@ function gc.compute_stack_slots(func)
 
     for i, cmd in ipairs(flat_cmds) do
         for _, val in ipairs(ir.get_srcs(cmd)) do
-            if val._tag == "ir.Value.LocalVar" then
+            if val._tag == "ir.Value.LocalVar" or val._tag == "ir.Value.Upvalue" then
```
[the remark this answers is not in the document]
hugomg: That said, I suppose it wouldn't hurt to mention this in that comment at top of the `gc.lua`.

pallene/to_ir.lua
```diff
@@ -87,23 +96,64 @@ function ToIR:exit_function(cmds)
...
+    else
+        -- upvalue is from the surrounding function's upvalue list.
+        assert(resolved.kind == "upvalue")
+        id = ir.add_upvalue(func_info.func, decl.name, decl._type, resolved.id, false)
+    end
+
+    return to_ir.Variable(id, "upvalue")
+end
+
 function ToIR:is_local(decl)
```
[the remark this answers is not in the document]
srijan-paul: I was trying to come up with one but couldn't think of any good ones 😅 
BTW, I think the `resolve_upvalue` can just be `resolve_variable` at this point (once it is capable of returning global vars).
What do you think?

pallene/to_ir.lua
```diff
@@ -820,9 +876,20 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
 
     elseif tag == "ast.Exp.Lambda" then
         local f_id = self:register_lambda(exp, "$lambda")
+        local func = self.module.functions[f_id]
         self:convert_func(exp)
         ir.add_exported_function(self.module, f_id)
-        table.insert(cmds, ir.Cmd.NewClosure(exp.loc, dst, f_id))
+
+        local upvalues = {}
+        for _, upval in ipairs(func.captured_vars) do
```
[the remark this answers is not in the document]
srijan-paul: If we did that, I'm curious as to how we'd get the type information when emitting C.

pallene/gc.lua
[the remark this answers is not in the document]
srijan-paul: >  Because the upvalues are stored in the function closure then if the function closure is reachable by the garbage collector 
Ah yeah, I somehow missed that 👍

pallene/to_ir.lua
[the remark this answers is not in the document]
hugomg: I agree. Maybe instead of `is_local` it could be `resolve_variable`.

pallene/to_ir.lua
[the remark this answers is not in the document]
hugomg: If the we have an `ir.Value.LocalVar` or an `ir.Value.Upvalue`we can look up the type in the `func.vars` and the `func.captured_vars`, respectively. Other kinds of  `ir.Value` would not give use the type information, but they should never happen so it's ok to `error("impossible")`.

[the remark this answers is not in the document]
srijan-paul: > If the we have an ir.Value.LocalVar or an ir.Value.Upvaluewe can look up the type in the func.vars and the func.captured_vars,

If it's a `LocalVar`, we can perform a lookup on `func.vars`, but if it's an upvalue, then how does a lookup in `captured_vars` help? Since `captured_vars` itself will `ir.Value`.  Do we perform some sort of recursive lookup until we get to the local var containing the `decl`?

pallene/coder.lua
```diff
@@ -519,16 +540,35 @@ function Coder:lua_entry_point_definition(f_id)
...
-        arg_vars[i] = name
-        arg_decls[i] = C.declaration(ctype(typ), name)..";"
+        local name = self:c_var(i)
+        table.insert(arg_vars, name)
+        table.insert(arg_decls, C.declaration(ctype(typ), name)..";")
     end
 
+
     local init_args = {}
+    local n_initialized_args = 0
```
hugomg, line 556: Could we get rid of the `n_initialized_args` if we used  `table.insert(arg_vars`?

pallene/coder.lua
```diff
@@ -1291,15 +1331,35 @@ gen_cmd["SetField"] = function(self, cmd, _func)
...
+    local func = self.module.functions[cmd.f_id]
+
+    local upvals = {}
+    for i, upval_info in ipairs(func.captured_vars) do
+        local c_val = self:c_value(upval_info.value)
+        upvals[i] = { c_val = c_val, typ = upval_info.decl.typ }
+    end
+
+    local capture_upvalues = {}
+    for i, c_upval in ipairs(upvals) do
```
hugomg, line 1343: Do we need the `upvals` local variable? Can we merge these two for loops into a single loop?

pallene/to_ir.lua
```diff
@@ -29,9 +29,18 @@ end
...
-        func           = func, -- ast.Exp.Lambda
-        f_id           = f_id  -- integer
+        loc_id_of_decl   = {},   -- { ast.Decl => integer }
+        upval_id_of_decl = {},   -- { ast.Decl => integer }
+        func             = func, -- ast.Exp.Lambda
+        f_id             = f_id  -- integer
+    }
+end
+
+function to_ir.Variable(id, kind)
```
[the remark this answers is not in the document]
hugomg: We might have forgotten to delete the `to_ir.Variable` function.

pallene/to_ir.lua
```diff
@@ -87,26 +106,44 @@ function ToIR:exit_function(cmds)
...
 end
 
---
---
---
-function ToIR:is_local(decl)
-    if decl._tag == "ast.Decl.Decl" then
-        local loc_id = self.loc_id_of_decl[decl]
-        local glb_id = self.glb_id_of_decl[decl]
+-- Returns a to_ir.Var object containing ID of `decl`.
```
hugomg, line 109: In addition to returning a Var, this function also has the side effect of registering the upvalues in this function and the surrounding functions. We should mention this in the comment.

pallene/to_ir.lua
```diff
@@ -87,26 +106,44 @@ function ToIR:exit_function(cmds)
...
+    while i <= #self.func_stack do
+        local loc_id = self.func_stack[i].loc_id_of_decl[decl]
         if loc_id then
-            assert(not glb_id)
-            return true, loc_id
+            var = to_ir.Var.LocalVar(loc_id)
+            break
+        end
+        i = i + 1
+    end
```
hugomg, line 131: Style suggestion:

```
local stack_id, var
for i, func in ipairs(self.func_stack)
    loc_id = ...
    if loc_id then
        stack_id = i
        var = ...
        break
    end
end
```

It's a matter of taste, but this lets us get rid of the mutable `i` variable and would allow us to rename the `j` so we do not have an `i` and `j` at the same time.

pallene/to_ir.lua
```diff
@@ -87,26 +106,44 @@ function ToIR:exit_function(cmds)
...
+
+    assert(var)
```
hugomg, line 133: Thanks for putting these asserts. They help understand the algorithm.

pallene/to_ir.lua
```diff
@@ -87,26 +106,44 @@ function ToIR:exit_function(cmds)
...
+    for j = i + 1, #self.func_stack do
+        local func = self.func_stack[j].func
+        local u_id
+        if var._tag == "to_ir.Var.LocalVar" then
+            u_id = ir.add_upvalue(func, decl.name, decl._type, ir.Value.LocalVar(var.id))
         else
-            assert(glb_id)
-            return false, glb_id
+            assert(var._tag == "to_ir.Var.Upvalue")
```
hugomg, line 140: I think it's clearer to put each case under an `elseif`.

```
elseif var._tag == "to_ir.Var.Upvalue" then
    ...
else
    error("impossible")
end
```

pallene/to_ir.lua
```diff
@@ -87,26 +106,44 @@ function ToIR:exit_function(cmds)
...
+        end
+        i = i + 1
+    end
+
+    assert(var)
+    for j = i + 1, #self.func_stack do
+        local func = self.func_stack[j].func
+        local u_id
+        if var._tag == "to_ir.Var.LocalVar" then
+            u_id = ir.add_upvalue(func, decl.name, decl._type, ir.Value.LocalVar(var.id))
```
hugomg, line 138: Is it possible that we call `add_upvalue` more than once for the same upvalue, if it is used multiple times inside the same function? I couldn't find an if statement guarding against that, but maybe I missed it.

hugomg (changes requested):

pallene/to_ir.lua
srijan-paul: You're correct, I had missed it after the refactor. I added a check for that back in, but I'm not sure if this makes the `resolve_variable` method a bit complicated to read. Should we refactor it out to smaller subroutines?

pallene/coder.lua
srijan-paul: I'm not sure about that. Currently `n_initialized_args` is being used in two places:
In the first loop:
```lua
local src  = string.format("&func->upvalue[%s]", C.integer(n_initialized_args - 1))
```
and in the second loop:
```lua
local dst = arg_vars[n_initialized_args]
```

I can't think of a way to get rid of it :sweat_smile: .

hugomg: Good question.

For the `func->upvalue[%s]`, I think it would be clearer if we used the upvalue id instead of `n_initialized - 1`.

For the `arg_vars`, what do you think about splitting it into two arrays? One list for the captured variables and one list for the function arguments?

pallene/to_ir.lua
hugomg: Instead of that `is_captured` for loop, can we use `self.upval_id_of_decl` to find out if an upvalue has already been registered?

pallene/coder.lua
srijan-paul: Ah, good idea. I forgot `captured_Vars` is a list now

srijan-paul:
(I had to rebase twice because of a weird issue where git would not ignore lines starting with `#`, everything should be ok now)

hugomg (approved):
Great! I edited the commit message to include the comments you gave in the PR.

We should be good to go. Next step is assignment conversion! :)

PR #390, opened 2021-06-10
Disallow func statements with multiple '.'

hugomg:
Because function statements in Pallene can only be used to initialize module functions, there is no use for "deep statements" like `function m.a.b.c.d()`. To simplify this, I'm changing the AST to only allow a single '.'.  I'm also replacing the "root" and "fields" by  "module" and "name". This way, in both cases you can get the funciton name using `.name`.

There is also a bug fix. I corrected how we check if a function is exported. Previously we used to check if the function statement had a "local" modifier, but that is no longer the right way to do it now that forward declarations exist.

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -433,12 +433,12 @@ function Parser:find_letrecs(stats)
             local func_names = {}
             for _, stat in ipairs(funcs) do
                 assert(stat._tag == "ast.Stat.Func")
-                if #stat.fields == 0 and not stat.method then
-                    if not forw_names[stat.root] then
+                if not stat.module then
+                    if not forw_names[stat.name] then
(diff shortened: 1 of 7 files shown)


pallene/parser.lua
```diff
@@ -502,16 +502,26 @@ function Parser:FuncStat(is_local)
 
     local fields = {}
     while self:try(".") do
```
srijan-paul, line 504: Since the grammar for toplevel functions is to only support names of kind `NAME("."NAME)?`, could this just be an `if self:try(".")` instead? Then we will no longer need the `if #fields == 1` check, since `field` could just be an NAME token or nil, instead of a table of NAME tokens.

srijan-paul (approved):

PR #392, opened 2021-06-11
Standardize the naming convention in the ir.Cmd

hugomg:
I renamed some fields to enforce the convention that all the fields with `ir.Value` are named `src` and all the fields with a variable ID are named `dst`.

Then I took advantage of this convention in the `ir.get_srcs` and `it.get_dsts`, so that we no longer need to write out all the field names. We also got rid of the "global" order of precedence for the field names. Now, the `ir.get_srcs` knows which field names to access for each kind of Cmd, and it accesses them in the same order that they appear in the constructor.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1447,7 +1447,7 @@ gen_cmd["Break"] = function(self, _cmd, _func)
 end
 
 gen_cmd["If"] = function(self, cmd, func)
-    local condition = self:c_value(cmd.src)
+    local condition = self:c_value(cmd.src_condition)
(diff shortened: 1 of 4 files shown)


srijan-paul (approved):
Personally , I think that `ir.Cmd.Loop` etc were more readable with field names like `cond`. Were there any problems that we were running into with the old field names and special cases for some cmds with unique fields?

hugomg:
We can change it to `src_cond` and so on.

srijan-paul (approved):

PR #395, opened 2021-06-20
Move the Stat.Func to a separate type

hugomg:
Currently, function statements can appear as either a standalone Stat.Func or inside a Stat.Letrec. This is awkward; let's clean it up ensuring that they are always part of the Letrec (which we also rename to Stat.Functions).

### DIFF
--- a/pallene/ast.lua
+++ b/pallene/ast.lua
@@ -34,19 +34,22 @@ declare_type("Decl", {
 })
 
 declare_type("Stat", {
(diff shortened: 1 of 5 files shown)


srijan-paul (approved):

srijan-paul:
I wonder if there exists a convention we're following that I was not aware of.
I found it reasonable that `Func` was under `ast,Stat`. Though the current representation isn't that odd either.

PR #397, opened 2021-06-21
Implement modules and require

VictorNogueiraRio:
Missing parts:

1. Tests focused on modules
2. Allow `Records` as a parameter to module functions


pallene/types.lua
```diff
@@ -26,14 +26,16 @@ declare_type("T", {
         "field_names", -- same order as the source type declaration
         "field_types", -- map { string => types.T }
     },
+    Module = {},
```
hugomg, line 29: Can we find a way to not have a module type? I'm happy that we recently got rid of the "fake" types `Void` and `Module` and I'm not looking forward to putting them back...

Perhaps by treating the `require` as special syntax instead of as a regular function call. Or maybe by saying that it returns `any`.

VictorNogueiraRio: @hugomg using `any` instead of modules works fine, as well.
Just committed this change now.

srijan-paul: Does that mean the user won't be able to shadow `require` anymore?
perhaps it could be a contextual keyword like `async` in Javascript.

VictorNogueiraRio: As of now, the user is still able to shadow `require`.
I think that's actually a good test case idea.

srijan-paul (approved):

pallene/coder.lua
```diff
@@ -1683,16 +1716,56 @@ function Coder:generate_luaopen_function()
                     entry_point = self:lua_entry_point_name(upv.f_id),
                     ix = C.integer(self.upvalue_of_function[upv.f_id]),
                 }))
+            elseif tag == "coder.Upvalue.ImportedFunction" then
+                local imported_func = self.module.imported_functions[upv.f_id]
+                modules[imported_func.mod] = modules[imported_func.mod] or {}
+                imported_func.id = upv.f_id
+                imported_func._tag = "func"
```
hugomg, line 1723: Assigning a `_tag` is strange. Should we create a proper tagged datatype, using `declare_type`?

pallene/ir.lua
```diff
@@ -114,14 +130,16 @@ end
...
-    Function   = {"id"},
+    Nil                = {},
+    Bool               = {"value"},
+    Integer            = {"value"},
+    Float              = {"value"},
+    String             = {"value"},
+    LocalVar           = {"id"},
+    Upvalue            = {"id"},
+    Function           = {"id"},
+    ImportedFunction   = {"id"},
```
hugomg, line 141: Silly question: why do we need separate cases for the Function and ImportedFunction?

pallene/coder.lua
VictorNogueiraRio: Fair enough, I think it would be better.

pallene/ir.lua
VictorNogueiraRio: I tried using `Function` at first and adding the imported functions to the `module.functions` table, but there are several places in the code that loop through `module.functions` and expect there to be a body field. Since we are not running the compiler past the `Checker` phase ( for obvious reasons),  the imported functions won't have a body and, thus, the checks will fail. In hindsight, I think it would be possible to stick with `ir.Value.Function` and just insert the imported functions into an `imported_functions` array. I'll see if it's actually possible.

hugomg: If the imported functions work differently, I think it is ok to have a separate case for them.

PR #399, opened 2021-06-22
Implement mutating closures

srijan-paul:
This PR introduces support for closures that can mutate their captured upvalues. (As log as they are 1. Not a function parameter and 2. initialized immediately upon declaration in their original scope).

The following notable changes have been made:
- Fixed a bug that crept in with the PR where we introduced non-mutating closures. Closures did not receive the global `G` userdata. This was fixed with a minor change in the IR instruction `NewClosure` in `coder.lua`.
- Added a new compiler pass, `assignment_conversion.lua`. A short brief overview of how this pass works has been added as a comment to the top of the file. Basically, it walks over the AST and for every function node:
    - Walk over it's body, and for each declared variable, maintain a list of all the places where it has been referenced in some way.
    - If it is later found that the decl is being mutated, then add it to a set to indicate that fact.
    - Before exiting the function, for each decl that has been mutated *and* captured by a nested closure, transform it's declaration node from `ast.Exp.*` to `ast.Exp.Initlist` and assign it a newly generated record type. Then, go to every place where this decl has been reference and change all `ast.Var.Name` references to `ast.Var.Dot`.

Function parameters and variables that are not initialized at their declaration site cannot be captured as of this PR.

### DIFF
--- a/pallene/assignment_conversion.lua
+++ b/pallene/assignment_conversion.lua
@@ -266,9 +266,6 @@ function Converter:visit_stat(stat)
     elseif tag == "ast.Stat.Call" then
         stat.call_exp = self:convert_exp(stat.call_exp)
 
-    elseif  tag == "ast.Stat.Func" then
-        self:visit_func(stat)
-
     elseif  tag == "ast.Stat.Break" then
         -- empty
     else
--- a/pallene/assignment_conversion.lua
+++ b/pallene/assignment_conversion.lua
@@ -55,13 +55,30 @@ local function FuncInfo()
     }
 end
 
+-- Encapsulates an update on an AST Node.
+-- A Node update has two parts: 
+-- `ref`: An ast.Var that denotes the node that has to be updated.
+-- `update_fn`: A `(ast.Var) -> ()` function that is used to update the node.
+-- ref's location is used to create a new AST Node. The new node is then used to replace the original
+-- occurance of `ref` in the AST by calling `update_fn` on the newly created node.
+local function NodeUpdate(ref, update_fn)
+    return {
+        ref       = ref,       -- ast.Var.Name
+        update_fn = update_fn, -- (ast.Var) -> ()
+    }
+end
+
 function Converter:init()
     self.func_stack = {} -- list of FuncInfo
     table.insert(self.func_stack, FuncInfo())
 
-    self.update_ref_of_decl      = {} -- { ast.Decl => list of (types.T) -> () }
+    self.update_ref_of_decl      = {} -- { ast.Decl => list of NodeUpdate }
+
+    -- The `func_depth_of_decl` is maps a decl to the depth of the function where it appears
+    -- This helps distinguish between mutated variables that are locals and globals from those
+    -- that are captured upvalues and need to be AST transformed.
     self.func_depth_of_decl      = {} -- { ast.Decl => integer }
-    self.update_init_exp_of_decl = {} -- { ast.Decl => (types.T) -> () }
+    self.update_init_exp_of_decl = {} -- { ast.Decl => NodeUpdate }
     self.box_records             = {} -- list of ast.Toplevel.Record
 
     -- used to assign unique names to subsequently generated record types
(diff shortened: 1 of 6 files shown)


pallene/assignment_conversion.lua
```diff
@@ -0,0 +1,347 @@
...
+function Converter:visit_prog(prog_ast)
+    assert(prog_ast._tag == "ast.Program.Program")
+    for _, tl_node in ipairs(prog_ast.tls) do
+        if tl_node._tag == "ast.Toplevel.Stats" then
+            for _, stat in ipairs(tl_node.stats) do
+                self:visit_stat(stat)
+            end
+        else
+            -- skip record declarations and type aliases
+            assert(typedecl.match_tag(tl_node._tag, "ast.Toplevel"))
```
VictorNogueiraRio, line 95: Is this else case really necessary?

pallene/assignment_conversion.lua
```diff
@@ -0,0 +1,347 @@
+local util     = require "pallene.util"
```
VictorNogueiraRio, line 1: Don't forget to add the copyright notice

srijan-paul: Aha, My bad. I typically tend to forget things like that

pallene/assignment_conversion.lua
srijan-paul: I just added it to make sure that we're not ever calling this function with the wrong argument.

VictorNogueiraRio: Okk

pallene/assignment_conversion.lua
```diff
@@ -0,0 +1,349 @@
...
+end
+
+function Converter:register_decl(decl)
+    if not self.ref_to_decl[decl] then
+        self.ref_to_decl[decl]        = {}
+        self.func_depth_of_decl[decl] = #self.func_stack
+    end 
+end
+
+function Converter:visit_prog(prog_ast)
```
hugomg, line 91: Why are some functions are called "visit" and some are called "convert"?

pallene/assignment_conversion.lua
```diff
@@ -0,0 +1,349 @@
...
+end
+
+function Converter:visit_func(func)
+    assert(func._tag == "ast.FuncStat.FuncStat")
+    local lambda = func.value
+    assert(lambda and lambda._tag == "ast.Exp.Lambda")
+    self:visit_lambda(lambda)
+end
+
+function Converter:convert_exps_of_stat(stat)
```
hugomg, line 183: Having a function that only works on some kinds of stats (the ones with an `.exps` field) doesn't smell right. IMO, it's ok to copy paste this for loop in the ast.Stat.Return and the ast.Stat.ForIn, etc.

pallene/assignment_conversion.lua
```diff
@@ -0,0 +1,349 @@
...
+    return var
+end
+
+-- If necessary, transforms `exp` or one of it's subexpression nodes in case they 
+-- reference a mutable upvalue.
+-- Recursively visits all sub-expressions and applies an `ast.Var.Name => ast.Var.Dot`
+-- transformation wherever necessary, returning the new transformed node.
+-- Note that for the transformed node to be updated at the call site, this method must
+-- be called like so:
+-- `exp = self:convert_exp(exp)`
```
hugomg, line 311: Do we ever return an `exp` that is not the same that we received as an argument? Maybe I'm missing it but I can't see where that happens. If this really does not happen, then we could return nothing instead.

I also ask the same question about the `convert_var`.

hugomg (changes requested):
While we did decide to leave uninitialized variables for later, on second thought I think we should stop to think if they will pose a problem.

What should be the assignment conversion for something like the following?
```
local x
if b then
    x = 1
else
    x = 2
end
```

Do we need to be able to represent an empty uninitialized record?

```
local x = <record with uninitialized fields>
if b then
    x.value = 1
else
    x.value = 2
end
```

Is this something that looks nice as an AST? One possibility would be to give up on using an AST transformation. Instead of transforming the AST, add some special annotations (similarly to the checker) and then in the to_ir use these annotations to know if a variable needs to be boxed or not.

Some pros:
1. Doesn't have the empty record problem
2. Doesn't have to mutate the Vars in place

Some cons:
1. The to_ir becomes more tightly coupled to the assignment conversion step.

I don't know what is the right answer here, what do you think?

pallene/assignment_conversion.lua
srijan-paul: My reasoning was that it becomes more clear to the caller how to call the function. So something called `visit_foo` can be called like `self:visit_foo(my_foo)` whereas `convert_foo` should be called as `my_foo = self:convert_foo(my_foo)`. (Since `convert_*` transforms it's parameter and returns the modified one.)

pallene/assignment_conversion.lua
srijan-paul: You're right. This was a misjudgement on my end. I changed that and the names to`visit_exp` and `visit_var`

srijan-paul:
> Do we need to be able to represent an empty uninitialized record?

I think I prefer this. The downside here is in the ToIR, we need to be sure that all the uninitialized records are from the assignment conversion transformations and not from user written code. But the good part is we don't have to introduce any new annotations that aren't standard to the AST.

hugomg (changes requested):

hugomg (approved):

PR #401, opened 2021-06-25
Tweaks

hugomg:
Some tweaks that were discussed in  #399.

PR #402, opened 2021-06-28
Implemented mutable parameter capturing.

srijan-paul:
Parameters can now be captured and mutated in closures.
Changes are primarily in `assignment_conversion.lua`.

For each parameter, we keep track of which lambda it belongs in (by using a `lambda_of_param` map).
At the end of the AST traversal, if it is seen that a parameter has been captured, then we take it's corresponding lambda and insert a 'proxy' declaration in it's `body`.

Next, we replace all references made to the parameter with this proxy declaration's `.value` field.

```
function foo(param)
  local $param = { value = param }
end
```

### DIFF
--- a/pallene/assignment_conversion.lua
+++ b/pallene/assignment_conversion.lua
@@ -147,8 +147,8 @@ function Converter:apply_transformations()
                 { "value" } ,
                 { value = decl._type }
             )
-            self:add_box_type(decl.loc, typ)
 
+            local type_node = self:add_box_type(decl.loc, typ)
             local init_exp_update = self.update_init_exp_of_decl[decl]
 
             if init_exp_update then
@@ -177,17 +177,18 @@ function Converter:apply_transformations()
                 --- end
                 local param = ast.Exp.Var(decl.loc, ast.Var.Name(decl.loc, decl.name))
                 param.var._def = checker.Def.Variable(decl)
-                param.var._type = decl._type
+                param.var._type = assert(decl._type)
                 param._type = decl._type
-                local decl_lhs = ast.Decl.Decl(decl.loc, "$"..decl.name, typ)
+
+                local decl_lhs = ast.Decl.Decl(decl.loc, "$"..decl.name, type_node)
                 local decl_rhs = ast.Exp.InitList(decl.loc, {{ name = "value", exp = param }})
                 decl_rhs._type = typ
                 decl_lhs._type = typ
 
-                local new_node = ast.Stat.Decl(decl.loc, { decl_lhs }, { decl_rhs })
+                local stat = ast.Stat.Decl(decl.loc, { decl_lhs }, { decl_rhs })
 
                 local lambda = self.lambda_of_param[decl]
-                table.insert(lambda.body.stats, 1, new_node)
+                table.insert(lambda.body.stats, 1, stat)
                 proxy_decl_of_param[decl] = decl_lhs
 
             else
--- a/pallene/assignment_conversion.lua
+++ b/pallene/assignment_conversion.lua
@@ -7,6 +7,7 @@ local util     = require "pallene.util"
 local typedecl = require "pallene.typedecl"
 local types    = require "pallene.types"
 local ast      = require "pallene.ast"
+local checker  = require "pallene.checker"
 
 local converter = {}
 
@@ -69,6 +70,7 @@ function Converter:init()
     self.mutated_decls           = {} -- list of ast.Decl
     self.captured_decls          = {} -- { ast.Decl }
     self.box_records             = {} -- list of ast.Toplevel.Record
+    self.lambda_of_param         = {} -- { ast.Decl => ast.Lambda }
 
     -- used to assign unique names to subsequently generated record types
     self.typ_counter = 0
@@ -88,6 +90,7 @@ function Converter:add_box_type(loc, typ)
     local dummy_node = ast.Toplevel.Record(loc, types.tostring(typ), {})
     dummy_node._type = typ
     table.insert(self.box_records, dummy_node)
+    return dummy_node
 end
 
 function Converter:register_decl(decl)
@@ -130,6 +133,8 @@ end
 -- Goes over all the ast.Decls inside the AST that have been captured by some nested
 -- function, transforms the decl node itself and all the references made to it.
 function Converter:apply_transformations()
+    local proxy_decl_of_param = {} -- { ast.Decl => ast.Decl }
+
     for _, decl in ipairs(self.mutated_decls) do
         if self.captured_decls[decl] then
             assert(not decl._exported_as)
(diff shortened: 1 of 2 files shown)


srijan-paul:
I've left some of the change that have already been made in #401 , since those would be taken care of once I rebase it after 401 is merged.

hugomg:
Can we try to rebase it now?

Em 28 de junho de 2021 06:41:40 BRT, srijan_ ***@***.***> escreveu:

srijan-paul:
> Can we try to rebase it now?

Yeah, I'll do that.

pallene/assignment_conversion.lua
```diff
@@ -143,17 +148,48 @@ function Converter:apply_transformations()
...
+                --- we create a variable to represent the boxed parameter which can be captured.
+                --- ```
+                --- function m.foo(n: integer)
+                ---   local $n: $T = { value = n }
+                ---   -- capture and mutate $n
+                --- end
+                local param = ast.Exp.Var(decl.loc, ast.Var.Name(decl.loc, decl.name))
+                param.var._def = checker.Def.Variable(decl)
+                param.var._type = decl._type
+                param._type = decl._type
```
hugomg, line 181: This part is tricky. I think we could use some `assert(decl._type)`, for defensive programming.

pallene/assignment_conversion.lua
```diff
@@ -143,27 +148,69 @@ function Converter:apply_transformations()
...
+                local decl_lhs = ast.Decl.Decl(decl.loc, "$"..decl.name, typ)
```
hugomg, line 182: If we provide a type here, it should be an `ast.Type`, not a `types.T`. We might be able to get away with setting it to `false` though.

srijan-paul:
> Oh wait, I thought that other one was already merged. Sorry about that!

Yeah, it is merged now actually. I rebased it 👍

pallene/assignment_conversion.lua
```diff
@@ -143,27 +148,69 @@ function Converter:apply_transformations()
...
+                local decl_rhs = ast.Exp.InitList(decl.loc, {{ name = "value", exp = param }})
+                decl_rhs._type = typ
+                decl_lhs._type = typ
+
+                local new_node = ast.Stat.Decl(decl.loc, { decl_lhs }, { decl_rhs })
```
hugomg, line 187: We're creating several nodes in this section of the code. We could call this `stat` or `new_stat` instead.

pallene/assignment_conversion.lua
```diff
@@ -143,27 +148,69 @@ function Converter:apply_transformations()
...
+
+                local lambda = self.lambda_of_param[decl]
+                table.insert(lambda.body.stats, 1, new_node)
```
hugomg, line 190: Two things that I'm not sure are a problem, but might be:

1. I suspect that the created statements might be in the opposite order of the parameters
2. This `table.insert` might result in O(n^2) behavior if we have many captured parameters

We might be able to avoid this by keeping the created statements in a list and then inserting them all at once, at the end of the program.

pallene/assignment_conversion.lua
```diff
@@ -143,27 +148,69 @@ function Converter:apply_transformations()
...
                 local update  = node_update.update_fn
 
-                local loc = old_var.loc
-                local new_node = ast.Var.Dot(loc, ast.Exp.Var(loc, old_var), "value")
+                local dot_exp
+                if proxy_decl_of_param[decl] then
+                    -- references to captured parameters get replaced by references to `value` field of
+                    -- their proxy variables.
+                    local proxy_var = ast.Var.Name(old_var.loc, "$"..decl.name)
+                    proxy_var._def = checker.Def.Variable(proxy_decl_of_param[decl])
```
hugomg, line 207: Technically this proxy_var and this def are the same for all the places that use the captured parameter. Should we reuse them instead of recreating them every time?

spec/execution_tests.lua
```diff
@@ -2509,6 +2509,13 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
 
                 return set, get
             end
+
+            function m.counter(x: integer): (() -> integer)
+                return function ()
+                    x = x + 1
+                    return x
+                end
+            end
```
hugomg, line 2518: Can we have a test case with more than one captured parameter?

hugomg (changes requested):
Having to `require` the checker makes me a bit nervous, but I think it's an unavoidable consequence of having to reset the `_type` and `_defs`... Would it be better to not create the proxy vars and then tell the `to_ir`to treat the parameters specially? Or would that be even messier than what we have now?

pallene/assignment_conversion.lua
srijan-paul: I decided to keep it that way because in a real AST those nodes would be repeated, right?
But I do agree that in this case we can use the same node. I'll change that 👍

srijan-paul:
> Would it be better to not create the proxy vars and then tell the to_irto treat the parameters specially? Or would that be even messier than what we have now?

I think that would be trickier (or perhaps just more baggage). Currently, mutable upvalues are self contained in the `assignment_conversion` pass, intertwining this between the `to_ir` and `assignment_conversion` sounds like it might not be too great. But I'm not 100% sure either.

pallene/assignment_conversion.lua
srijan-paul: > I suspect that the created statements might be in the opposite order of the parameters

Since we're merely referencing the parameters by name while creating the record boxes, this might not be a problem. What do you think?

pallene/assignment_conversion.lua
srijan-paul: I think I will use the box type returned by `add_box_type`, that should be fine too I imagine.

pallene/assignment_conversion.lua
srijan-paul: About the`table.insert`, I think the older method of doing assignment conversion at the end of each function was more suited to this. If we want to store all proxy statements in a table and `table.concat` it with `lambda.body.stats`, we would have to go over the `self.mutated_decls` lists again, and we'd have to keep track of which mutated decls are updated and which aren't. And a map that tells us which list of proxy decl stats belong to which lambda.

pallene/assignment_conversion.lua
hugomg: Since we're past the type checker, I think it's OK to not add the `ast.Type`. As long as the `_type` is there, everything should still work.

pallene/assignment_conversion.lua
hugomg: The reason I'm thinking about the order of the statements, is that it can affect the order of the local variables that are created, which in turn can affect the `gc.lua`. But there's a decent chance that it doesn't really matter.

>  And a map that tells us which list of proxy decl stats belong to which lambda.

I think we would still need that map even if we used the `exit_lambda`? My gut instinct is to keep doing it without the exit_lambda. But if it really is easier to go back to exit lambda that would be ok.

hugomg:
> I think that would be trickier (or perhaps just more baggage).

I think I agree. There are certainly benefits to making the AST transformation self-contained. Furthermore, while this couples the ast transformation to the type checker, it is coupled in the same way that the to_ir is coupled to the typechecker (via the `_type` and `_def` nodes). From this point of view it doesn't sound bad.

Let's keep the AST transformation self-contained. We can always change our mind later if something unexpected happens.

pallene/assignment_conversion.lua
hugomg: I wonder if we could replace the `proxy_decl_of_param` by a `proxy_var_of_param`, or something of the sort.

pallene/assignment_conversion.lua
srijan-paul: This did not take as many changes as I thought it would!

pallene/assignment_conversion.lua
srijan-paul: Another reason I think using the same node might not be the best is because of the `loc` field. I think it's best to keep the `loc`s as close as possible to the original var nodes.

hugomg (changes requested):
Just a handful of questions:

1. I think it would be cleaner if the apply_transformations and the box_records stuff happened in the main `convert` function instead of inside the visit_program. The visit method should only visit things.
2. The visit method currently returns the AST. It could return nothing.
3. Do we need to have a separate method for the `update_refs`? I think it's OK to leave everything inside the apply_transformation. If you are worried about the function getting too long, a couple of comments between the code "paragraphs" should do the trick.

srijan-paul:
>  If you are worried about the function getting too long, a couple of comments between the code "paragraphs" should do the trick.

Yeah, that was my primary concern. I removed that function now.

hugomg (commented):
Great! Next step is making sure that uninitialized variables work, correct?

I'm going to pay special attention to the ast.Stat.Decl case in the visit_stat. Currently we take care of the case when `#exps > #decls` but we'll also need to to the right thing when `#decls > #exps`.

hugomg (approved):

srijan-paul:
> Great! Next step is making sure that uninitialized variables work, correct?

Yes, that is what I have planned for tomorrow :grin:

Issue #398, opened 2021-06-22
Remove the metatables for upvalue records


srijan-paul:
We have to remember to remove the metatables for the artificial record types that are introduced by the assignment conversion pass.

PR #404, opened 2021-07-03
Implement upvalues that aren't initialized upon declaration.

srijan-paul:
Using a new `ast.Exp.UpvalueRecord`, the IR can now have records that aren't initialized upon declaration.
This dummy node is added to the corresponding `ast.Decl` of uninitialized upvalues in the `assignment_conversion` pass.


hugomg (approved):
Great!

For the next step, I suggest issue #398. And start thinking about the specification for the "grouped boxes" optimization.

PR #405, opened 2021-07-05
Remove metatables for upvalue boxes.

### LINKED ISSUE #398: Remove the metatables for upvalue records

srijan-paul:
Fixes #398.
Done using a simple type annotation `_is_upvalue_box` to the `types.T` of upvalue records.


hugomg (changes requested):
Could we make the `is_upvalue_box` into a regular field (called `has_metatable`), instead of an annotation?

By the way, if you want Github to recognize the issue number you need to [literally write](https://github.blog/2013-01-22-closing-issues-via-commit-messages/) "Fixes \#xxx" or "Closes \#xxx".

srijan-paul:
> Could we make the is_upvalue_box into a regular field (called has_metatable), instead of an annotation?

I thought about that, but then every call to the c'tor for `types.T.Record` would need an extra `false` argument. I'm wondering if that would make the code a bit iffy to read.

hugomg:
I think the extra `false` in the constructor is better than adding an annotation after the fact. It helps ensure that this field is always created

Another thing I remembered: we should add some `assert` statements in places that currently assume that a record object has a metatable. For example, the `test_tag` method should raise an exception at compile time, if we try to use it on a record type that doesn't have a metatable.

pallene/coder.lua
```diff
@@ -824,18 +824,26 @@ function RecordCoder:declarations()
     end
 
     -- Constructor
+    local set_metatable
+    if self.record_typ._is_upvalue_box then
```
hugomg, line 828: Why does this have an underscore?

hugomg (changes requested):
We still need to add assertions in the places that use the metatable. The `test_tag` is one, but there may be others.

hugomg:
Hi Srijan,


I talked to Roberto and he thinks it's OK to remove the metatables. We should proceed with this pull request.

Generally speaking, we don't have to protect against other C files misbehaving, because C has many ways to do unsafe things that crash the program. We only need to have run-time checks to protect against things that plain Lua code can do. He said that when regular C code uses the Lua-C API, it's customary assume that the types of the upvalues are correct, without a run-time check. There's an example of this in the Programming in Lua book: 

> This complete example has an interesting subtlety. At first, it may seem that dir_gc should check whether its argument is a directory and whether it has not been closed already. Otherwise, a malicious user could call it with another kind of userdata (a file, for instance) or finalize a directory twice, with disastrous consequences. However, there is no way for a Lua program to access this function: it is stored only in the metatable of directories, which in turn are stored as upvalues of the iteration functions. Lua programs cannot access these directories.

I think that one way to implement this, would be to use `unchecked_get_slot` instead of `get_stack_slot` when we get the upvalue slots in the entry point. Just make sure to add a comment explaining why we're using the unsafe function.

hugomg (changes requested):
I suspect that we are still generating the metatables for the record types that do not need them. We need to not create them in the first place.

I would also appreciate if we could double check the generated C code, to ensure that we did not forget to remove anything. Thanks!

srijan-paul:
> I suspect that we are still generating the metatables for the record types that do not need them. We need to not create them in the first place.

I believe [this](https://github.com/pallene-lang/pallene/pull/405/files#diff-ba37d7a37e782afcdf924f13a5db09a2023a937736a0af2e6411a68f258e343bR833) line in the diff avoids this problem. From what I can tell, the C code seems to be clean right now on the examples I tested with. Should I share a paste and related pallene code to be sure?

hugomg:
The line you linked refers to the constructor for the Box record, which indeed does not add a metatable to the box.

However, the problem is that we are still allocating a metatable for the box datatype and saving it as of of the fields of the `G`. This metatable is never used, and shouldn't need to be created in the first place.

```lua
local m = {}
function m.counter(): () -> integer
    local x = 0
    return function()
        x = x + 1
        return x
    end
end
return m
```

```c
int luaopen_foo(lua_State *L)
{
    luaL_checkversion(L);
    
    /* Constants */
    
    lua_newuserdatauv(L, 0, 3);
    int globals = lua_gettop(L);
    
// I believe that the following block of code
// is allocating a metatable that we never use:
    lua_newtable(L);
    lua_pushstring(L, "__metatable");
    lua_pushboolean(L, 0);
    lua_settable(L, -3);
```

hugomg:
While looking at the generated C code, I spotted a separate problem involving the Lua entry points. But I think we can leave that for another PR. I opened a separate issue so we don't forget (#409).

srijan-paul:
> However, the problem is that we are still allocating a metatable for the box datatype and saving it as of of the fields of the G

Ah, I didn't notice that! I think I misunderstood the problem. I'll fix this one.
And good point on the internal closures being added to `G`!

hugomg (approved):
Great!

PR #415, opened 2021-07-15
Linter: check if copyright headers are present

### LINKED ISSUE #400: Have the linter detect a missing copyright comment
It would be nice if the linter could detect a missing copyright comment, so that this could be caught in the continuous integration.

hugomg:
When we create new source code files, we sometimes forget to add a copyright header. Now the linter should catch that mistake if it happens. Closes #400.

### DIFF
--- a/lint-project
+++ b/lint-project
@@ -13,7 +13,7 @@ error() {
 }
 
 echo "=== Lua Lint ==="
-luacheck pallene/ spec/ "$@" || exit 1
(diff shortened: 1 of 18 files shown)


hugomg:
This commit rewrites the `lint-project` script from scratch. Instead of using recursive grep, use a for loop to iterate the files one by one.

For the review, it might be easier to look at the entire file before and after, instead of the line-by-line diff.

srijan-paul (approved):

PR #416, opened 2021-07-16
Add compile-time assertion for luaS_newudata

### LINKED ISSUE #412: Be careful about the limit of uservalues in  Userdata objects
Lua has a large limit of the maximum number of upservalues associated with an userdata object. The `nuvalue` in the `Udata` struct is an `unsigned short`. We should add some assertions in the compiler to ensure that records and the constant table don't go over this limit.

If possible, also provide some friendly user-facing error messages.

Currently, I think the most likely way to hit this limit  is if a program with a very large number of string constants.

hugomg:
To avoid integer overflow, we must ensure that the parameter to luaS_newudata is not too large. This can happen if a record type has an unreasonably large number of GC fields, or if a program has too many string literals.

Because USHRT_MAX is platform specific I chose to put the compile-time assertion in the C code instead of in the Lua code.

I also decided to not bother with a nice user-friendly error message. That would be harder to do with the assertion in the generated C code. I think this is OK because this sort of error should only happen for pathological programs.

Closes #412.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -780,6 +780,9 @@ end
 function RecordCoder:declarations()
     local declarations = {}
 
+    assert(self.prim_count >= 0)
+    assert(self.gc_count >= 0)
+
     -- Comment
     table.insert(declarations, C.comment(self.record_typ.name) .. "\n")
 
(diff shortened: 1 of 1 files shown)


srijan-paul:
One easy way to give the assertion an error message would be `#define ASSERT(cond, msg) assert((cond) && msg)`

srijan-paul (approved):

Issue #137, opened 2019-12-10
Find a better way to test the Pallene garbage collector


hugomg:
IIRC, Lua has an internal test setting that causes the garbage collector to run at every location where it could potentially run. That is, every "condGC" calls the garbage collector.

We should find a way to use this to test the Pallene garbage collector.

hugomg:
Another thing that would be nice, is to force a stack reallocation after every check stack.

gabrielsferre:
We can make so that "condGC" always calls the garbage collector by defining the macro "HARDMEMTESTS" in C. The problem is to decide how to set the Pallene compiler up so that it uses this macro for testing the garbage collector.

The two obvious ways to do this that I can think of are either adding a CLI flag  in "pallenec" for defining this macro in the C code or changing the "run-tests" script so that it has a mode in which it defines the macro using the "CFLAGS" env variable. I tend to like the latter idea better because it doesn't mess with the compiler, and if someone wants to test the GC outside the "run-tests" tests, they can just set CFLAGS manually by using a command like "CFLAGS='-O4 -DHARDMEMTESTS'  pallenec file.pln" or something.

hugomg:
I agree.

Have you managed to confirm whether the hard mem tests work using cflags? If they do we can update the run-scripts and/or the CONTRIBUTING docs accordingly.

gabrielsferre:
> Have you managed to confirm whether the hard mem tests work using cflags?

I did, they're working.

I'll make the changes and put them into a PR with the other GC changes. Is that ok?

hugomg:
Great!

Maybe this is better as two separate prs. One that changes the gc stuff to use basic blocks, and a seconf for the hardmemtests.

gabrielsferre:
Ok, I'll do that.

PR #417, opened 2021-07-16
Fix a memory corruption bug with luaD_checkstack

hugomg:
If the stack grows after the luaD_checkstack, we must update the `base` pointer accordingly.

This fixes a nasty memory corruption bug. When the checkstack moved the stack we would use outdated values of `base` and `base_offset`.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -428,17 +428,20 @@ function Coder:pallene_entry_point_definition(f_id)
     end
 
     local prologue = {}
+    table.insert(prologue, self:savestack())
 
     local max_frame_size = self.gc[func].max_frame_size
     local slots_needed = max_frame_size + self.max_lua_call_stack_usage[func]
     if slots_needed > 0 then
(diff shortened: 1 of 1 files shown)


hugomg:
Before:

```c
luaD_checkstack(L, 100);
```

After:

```c
luaD_checkstackaux(L, 100,
    (void)0,
    base = restorestack(L, base_offset));
```

The difference between the luaD_checkstack and luaD_checkstackaux macros is that the aux version receives an extra argument, which is a code snippet that runs if the stack is relocated.

srijan-paul (approved):
Should we add a test that is guaranteed to grow the stack, just to make sure we can avoid problems like these in the future?

hugomg:
I don't think a normal test would help much here. I is hard to create a test that fails cleanly when this is wrong.

Maybe a better way to test is with #137. Force the stack to move after every function call and run under valgrind to catch invalid memory writes.

PR #419, opened 2021-07-17
Lua 5.4.3

hugomg:
This PR updates our bundled Lua version to 5.4.3

After it is merged, remember to recompile your project with `make linux-readline -j4`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,6 @@ jobs:
         steps:
             - uses: actions/checkout@v2
             - uses: leafo/gh-actions-lua@v8.0.0
-              with:
-                  luaVersion: "5.4.3"
             - uses: leafo/gh-actions-luarocks@v4.0.0
 
             - name: Install Luacheck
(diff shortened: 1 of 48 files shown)


hugomg:
@srijan-paul This PR, and some of the other ones I'm planning for the near future, might change some of the internals of the interpreter and the build system. They might be disruptive to the experiments to the current round of experiments, which relies on  hand-crafted code. Feel free to run the experiments on the version of Pallene before these disruptive PRs. If you want you can add a comment referencing the SHA identifier for the Pallene version that you're running the experiments in.

srijan-paul:
> Feel free to run the experiments on the version of Pallene before these disruptive PRs

Noted.
This PR just changes the version of Lua right?

I can't tell which changes are from the Lua interpreter source and which ones are changes to Lua required by Pallene.

srijan-paul (approved):

hugomg:
Yes, this change is only to update the version of Lua. As you said, it might be easier to look at the individual commits :)

We often have to make some tweaks to the coder in response to changes in the internal APIs, so it can break the generated C files from an older version of the compiler. In this case, it was a change to `luaH_newkey`. The most disruptive changes are yet to come, however. I'm planning on renaming the `lua` directory to another name, to work better with Luarocks. I'm also planning on moving the `runtime` files to inside the Lua interpreter, so they become part of the compiled interpreter, instead of a separate pallenelib library.

PR #422, opened 2021-07-21
Revert the changes we made to the luaconf in 5.4.3

hugomg:
Instead of making the l_likely public in the luaconf, what we should do is mark the Pallene. generated files as being "LUA_CORE". The "define LUA_CORE" is something that already existed in previous versions of Lua, and we probably should already have been using it before. However, that 5.4.3 thing with the l_likely was the first time that this came up for Pallene.


hugomg:
This removes the additional patch that I made to luaconf when I ported to 5.4.3. Now it only has the patches that it used to have in version 5.4.2.

srijan-paul (approved):

PR #423, opened 2021-07-22
Limit the maximum number of function parameters / arguments.

### LINKED ISSUE #410: Add a limit for the number of function parameters
Suggested limit: 200 function parameters

Lua itself does not allow functions with that many parameters, due to the limit on the number of local variables (MAXVARS in lparser.c). Furthermore, it is not very useful to define functions with that many parameters, because Lua would have a hard time calling them. When Lua function calls have that many arguments, they start hitting the limit of VM registers per function (255).

The compiler should give a clean user-facing error message if the program has a function with too many arguments. The parser might be a good place to do this.

srijan-paul:
Fixes #410 
The error message is similar to that thrown by Lua.
I'm not sure if it would be better to have a long snippet with too many parameters instead of pragmatically generating a string in the test cases.

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -13,6 +13,11 @@ local util = require "pallene.util"
 
 local Parser = util.Class()
 
+-- The Lua VM might have a hard time calling functions with too many arguments.
+-- Lua itself does not allow more than 200 upvalues, function parameters or local variables 
+-- since it uses 1-byte long unsigned numbers to store stack offsets for locals.
+local MaxParams = 200
+
 function Parser:init(lexer)
     self.lexer = lexer
     self.prev = false -- Token
@@ -22,7 +27,6 @@ function Parser:init(lexer)
     self.region_depth = 0     -- Are we inside a type annotation?
     self.type_regions = {}    -- Sequence of pairs. Ranges of type annotations in program.
     self.comment_regions = {} -- Sequence of pairs. Ranges of comments in the program.
-    self.max_params = 200
     self:_advance(); self:_advance()
 end
 
@@ -784,9 +788,9 @@ function Parser:FuncArgs()
         local open = self:e("(")
         local exps = self:peek(")") and {} or self:ExpList1()
         local _    = self:e(")", open)
-        if #exps > self.max_params then
-            self:syntax_error(exps[self.max_params + 1].loc,
-                "too many arguments (limit is %d)", self.max_params)
+        if #exps > MaxParams then
+            self:syntax_error(exps[MaxParams + 1].loc,
+                "too many arguments (limit is %d)", MaxParams)
         end
         return exps
     end
@@ -795,9 +799,9 @@ end
 function Parser:FuncParams()
     local oparen = self:e("(")
     local params = self:DeclList()
-    if #params > self.max_params then
-        self:syntax_error(params[self.max_params + 1].loc,
-            "too many parameters (limit is %d)", self.max_params)
+    if #params > MaxParams then
+        self:syntax_error(params[MaxParams + 1].loc,
+            "too many parameters (limit is %d)", MaxParams)
     end
     local _ = self:e(")", oparen)
     return params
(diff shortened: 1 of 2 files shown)


spec/parser_spec.lua
```diff
@@ -419,6 +419,19 @@ describe("Parser /", function()
             ]], "local function name has a ':'")
         end)
 
+        it("disallow too many function parameters", function ()
+           local params = ''
+           for i = 1, 200 do
+              params = params..'a'..i..', '
+           end
+           params = params..'a201'
```
hugomg, line 427: It's better to avoid repeated string concatenation inside a loop, that's O(n^2)  for Lua. I suggest using table.concat, which should also allow you to stop special-casing the last arguments.

Also, our coding style favors double-quoted strings.

```lua
local xs = {}
for i = 1, 201 do
    xs[i] = "a"..i
end
local params = table.concat(xs, ", ")
```

spec/parser_spec.lua
```diff
@@ -419,6 +419,19 @@ describe("Parser /", function()
...
+           assert_toplevel_error(string.format([[
+                function m.f(%s)
+                    return 1
+                end
+           ]], params), "too many parameters (limit is 200)")
```
hugomg, line 432: I'm very much in favor of generating the argument list programatically, instead of writing it by hand. However, in other tests that do this kind of thing we use string concatenation instead of string.format. For example, see the "maxlenfield" test cases in the execution tests.

    function m.f(]]..params..[[)

pallene/parser.lua
```diff
@@ -22,6 +22,7 @@ function Parser:init(lexer)
     self.region_depth = 0     -- Are we inside a type annotation?
     self.type_regions = {}    -- Sequence of pairs. Ranges of type annotations in program.
     self.comment_regions = {} -- Sequence of pairs. Ranges of comments in the program.
+    self.max_params = 200
```
hugomg, line 25: Since this is a constant, I think  it would be better to define it as a toplevel local variable, instead of as an instance variable.

While we are at it, it might also be a good idea to add a comment next to this variable, describing why the limit exists:
* Lua has a hard time calling functions with close to 255 arguments
* Lua itself does not allow functions with more than 200 local vars

spec/parser_spec.lua
```diff
@@ -419,6 +419,19 @@ describe("Parser /", function()
             ]], "local function name has a ':'")
         end)
 
+        it("disallow too many function parameters", function ()
```
hugomg, line 422: This looks like it has an indentation of only 3 spaces?

spec/parser_spec.lua
```diff
@@ -555,6 +568,17 @@ describe("Parser /", function()
...
 
+        it("disallows too many arguments", function ()
+           local args = ''
+           for i = 1, 200 do
+              args = args .. i .. ', '
+           end
+           args = args ..'201'
+            assert_statements_error(string.format([[
+                f(%s)
+            ]], args), "too many arguments (limit is 200)")
```
hugomg, line 579: The same comments for the other test case apply here (indentation, table.concat, and string.format)

hugomg (changes requested):
I had some requests about coding style, but other than that the functionality in this PR is exactly what we need.

spec/parser_spec.lua
srijan-paul: Ah! I tweaked my `vimrc` and ended up messing with something that may have led to this😅

hugomg: While you're checking the vimrc, it might be a good idea to verify if the editorconfig plugin is working.

By the way, this 3-spaces indentation reminds me of Hisham. He's not afraid to swim against the current, and argued that 3 spaces is good for Lua because it aligns the "end" tokens.

hugomg (approved):

hugomg:
Oof. Looks like github's "update branch" button actually does a merge instead of rebase. I'll try to fix it.

spec/parser_spec.lua
srijan-paul: Good point, I totally forgot to setup editorconfig on the WSL.
I prefer tabs personally because I can render them at any width, so I do 2 space-wide tabs for Lua generally, but 
looking at it like that 3-space tabs sounds like a good argument.

srijan-paul:
> 
> 
> Oof. Looks like github's "update branch" button actually does a merge instead of rebase. I'll try to fix it.

I could just do a rebase + force-push from my local branch and that should work too I think.

hugomg:
That's the fix I did. However, it's important to avoid two persons trying to do that at the same time ;)

Issue #153, opened 2019-12-22
We should figure out how to approach custom versions of Lua


hugomg:
Lua is a flexible language. There are multiple ways to build a custom version of it by tweaking the luaconf.h or patching its source code, and these changes all affect the API between the Lua VM and C extension modules.

There are two different directions Pallene can take in this situation. We could require that Pallene code only works in our own custom version of Lua, or we should make it so that Pallene can also be used by the various custom versions of Lua. Currently, Pallene is halfway between these two options, and I think we should pick one side or the other:

1. Pallene modules can only run against versions of Lua that edit the luaconf.h to make all the private functions public.
2. Pallene modules can only run against versions of Lua that make all the internal ".h" header files public (lobject.h, lvm.h, etc)
3. Pallene assumes that `LUAI_MAXSHORTLEN` is 40
4. Pallene assumes that `LUA_NUMBER` is `double`. (This affects how to print floating-point literals in C.lua)

Option 1 wold be to require Pallene to only work with our custom version of Lua 5.4. We would need to modify the installation scripts and installation instructions to have a way to install our custom version of Lua.

Option 2 would be to allow Pallene to work with any suitably-configured version of Lua 5.4. We would need to describe what are the required settings in the luaconf.h and how to make the header files like `lobject.h` and `lvm.h` visible to Pallene. Pallene should also produce a clean error when something that it assumes is not what it expects (e.g., LUA_NUMBER is not `double`). These can be compile-time errors, or run-time errors at the start of the `luaopen` function.

Option 3 would be to find a way to allow Pallene modules to be used with a totally unmodified version of Lua 5.4, without requiring any changes to luaconf or that any internal header files be made public. In order for this to work, the Pallene library code would need to bundle a copy of all the necessary Lua interpreter functions. We would also need to find a way to ensure that this bundled version of Lua is compatible with the Lua interpreter that is calling it.

Whatever solution we end up with, I think this is going to be a key step to solve #16.

gligneul:
I wish we could link to any Lua release during the compilation of a Pallene module. This could be an optional command line argument that receives the path to the Lua source code. Then, we copy the internal functions that we need so we don't need to modify the interpreter source code to expose them.

hugomg:
I think what you describe would fit with option 2. AFAIK, the only modification we have right now to the Lua source code is that we change the luaconf.h to make the private functions be public.

[...]

Issue #16, opened 2018-10-02
Allow the compiler to be called from any directory


hugomg:
Currently, `pallenec` only works if the current directory is the root project directory. If we try to call it elsewhere it fails for the following reasons:

- [ ] The C compiler can't find the headers from our bundled version of Lua
- [ ] The `require` can't find the modules in the titan-compiler subdirectory. (LUA_PATH issue)
- [ ] The C linker (called in c_compiler.lua) cannot find our runtime library.

gligneul:
Remainder: update readme when done.

gligneul:
I don't this this is an issue for the dev version.
This issue could be solved if install Pallene properly with Luarocks.

hugomg:
Even for the dev version, being effectively restricted to only running code inside the root directory is a big limitation. It is certainly affecting how we write our benchmark and test scripts. (See #17)

I agree that properly installing pallene via luarocks is the way to go. In fact, figuring out how to do so is exactly why I opened this issue in the first place :)

hugomg:
I've been thinking about this problem and I think the tricky bit is going to be the Pallene runtime library. Currently we have a custom Lua interpreter and a runtime library (pallenelib). In theory we could merge one of them into the other but in any case we will need a runtime library component and that brings two problems

1. The runtime library is not a Lua module that you load with `require`, and Luarocks is not designed to install something like that.
2. We currently statically link the pallene library but if we would like the option of dynamically linking it then it would need to be installed to /usr/local/lib instead of the Luarocks installation prefix. Linux doesn't like when your shared library is hidden away inside the Luarocks directory hierarchy.

So what do you think about having a separate installer for the Pallene compiler and for the Pallene runtime library?

The Pallene compiler can continue as a Luarocks module. Luarocks knows how to install it and also takes care of installing our Lua dependencies.

For the Pallene runtime library, we would take our existing makefile and add a `make install` rule that would install the library and the pallene_core.h header file to the standard system directories. We should also try to configure things so that the local directory takes precedence during development, to make the version we are currently working on take precedence over the installed version.

hugomg:
To solve this issue for good I think we will need to clean up the situation with the pallene runtime library and the custom Lua interpreter.

I investigated what private Lua functions we are currently using. For the record, this is the list that I got after running `nm` on a large Pallene module packed with as many language features as I could:

* luaC_step
* luaD_growstack
* luaH_getn
* luaH_new
* luaH_resizearray
* luaL_error
[...]

PR #424, opened 2021-07-22
Move pallenelib to inside the Lua interpreter

hugomg:
This set of commits gets rid of the separate pallenelib library. This is part of the ongoing effort to simplify the build and installation process for Pallene (see #16, #153). The pallene_core files are now going to be part of our the custom installation of Lua.

Unfortunately, for some reason it is not working. When I compile the custom Lua, the
`pallene_*` symbols are not being exported by the custom Lua executable. We'll need to figure out what is the problem.

TODO:
- [x] Find out why the symbols are not being exported
- [x] Fix compilation warnings that appeared because of -Wextra
- [x] Update the patch file we use when updating to a new Lua version
- [x] Delete the "runtime" folder, and find a new place to store the patch file
- [x] Update the installation instructions in the README.


hugomg:
I think this might be one of those PRs where it is easier to understand if it is read one commit at a time.

vm/src/lua.c
```diff
@@ -657,3 +657,16 @@ int main (int argc, char **argv) {
...
 
+/*
+** PALLENE LINKER HACK
+**
+** When we build the final "lua" executable, bundling it with the "liblua.a" library, the linker
+** only includes symbols from the object files where at least one of the symbols of the module is
+** used by the executable itself. This is a problem for the Pallene runtime library, because those
+** symbols are only used by extension modules, which we expect to dynamically link. One hacky
+** workaround I found is to force the "lua" executable to reference one of the symbols from the
+** pallene_code, as done below.
```
srijan-paul, line 668: This was supposed to say `pallene_core` perhaps?

vm/src/pallene_core.c
```diff
@@ -159,6 +161,7 @@ void pallene_grow_array(lua_State *L, const char* file, int line, Table *arr, un
 
 void pallene_io_write(lua_State *L, TString *str)
 {
+    (void) L;
```
srijan-paul, line 164: I wonder what this does?

pallene/coder.lua
```diff
@@ -1265,7 +1265,7 @@ gen_cmd["GetTable"] = function(self, cmd, _func)
 
     return util.render([[
         {
-            static size_t cache = UINT_MAX;
+            static int cache = UINT_MAX;
```
srijan-paul, line 1268: should this be `INT_MAX` now?

srijan-paul (changes requested):
I'm wondering if it would be possible to keep the pallene files in `vm/pallene/` instead of `vm/src/`? and if that would be better?

README.md
```diff
@@ -70,18 +70,18 @@ $ echo 'eval `luarocks path`' >> ~/.bashrc
...
-Pallene must be run against a custom-built version of the Lua interpreter, as
-well as the Pallene runtime library. Both of these are written in C and must be
-compiled before the Pallene compiler can be used.
+Pallene must be run against a custom-built version of the Lua interpreter.
+This custom version of Lua 5.4 doesn't have to be the same one that you will use to run the compiler itself,
+or to install the Luarocks packages.
 
-These two components can be built through the Makefile we provide. The command
-to be used depends on your operating system:
+To compile the custom version of Lua, follow the instructions found the vm/doc/readme.html.
```
srijan-paul, line 79: > instructions found the

"Instructions found in the [path]" perhaps?

srijan-paul (changes requested):

README.md
hugomg: Apparently, Github displays the raw HTML if we link to that. I think I'm going to link to the lua.org website instead.

hugomg:
Because the Lua Makefile cds into the "src" directory do to it's thing, I think it's clearer if we put all our files there instead of on a separate file. But I'm glad you asked, because I found a separate bug --  I had forgotten to modify the parent Makefile to tell it to install the additional header files.

srijan-paul (approved):

PR #438, opened 2021-07-27
Avoid creating closures for lambdas in luaopen.

### LINKED ISSUE #409: The luaopen should not create closure objects for the lambdas
Currently the luaopen function is creating a closure object for every lambda in the program, and saving them to the `G`:

```lua
local m = {}
function m.counter(): () -> integer
    local x = 0
    return function()
        x = x + 1
        return x
    end
end
return m
```

```c
int luaopen_foo(lua_State *L)
{
    luaL_checkversion(L);
    
    /* Constants */
    
    lua_newuserdatauv(L, 0, 3);
    int globals = lua_gettop(L);
    
    lua_newtable(L);
    lua_pushstring(L, "__metatable");
    lua_pushboolean(L, 0);
    lua_settable(L, -3);
    
    lua_pushvalue(L, globals);
    lua_pushcclosure(L, function_02_lua, 1);
    lua_setiuservalue(L, globals, 2);
    
    lua_pushvalue(L, globals);
    lua_pushcclosure(L, function_03_lua, 1);
    lua_setiuservalue(L, globals, 3);
```

In this program, `function_03` refers to the lambda inside the `m.counter` function.There no reason to create this closure in the luaopen and save it to the `G` list. The closure is also invalid, by the way, with the wrong number of upvalues.

srijan-paul:
This fixes #409 
The current fix is having a set of function IDs that represent lambdas in an `ir.Module`, and then checking to see if a certain function is a lambda before creating a closure for it in luaopen.
It works, but would it be cleaner if we chose to not include the lambdas in `exported_functions` instead?

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -672,14 +672,13 @@ function Coder:init_upvalues()
     for _, func in ipairs(self.module.functions) do
         for cmd in ir.iter(func.body) do
             for _, v in ipairs(ir.get_srcs(cmd)) do
-                if v._tag == "ir.Value.Function" and not self.module.lambdas[v.id] then
+                if v._tag == "ir.Value.Function" then
                     table.insert(closures, v.id)
                 end
             end
         end
     end
     for _, f_id in ipairs(self.module.exported_functions) do
-        assert(not self.module.lambdas[f_id])
         table.insert(closures, f_id)
     end
 
@@ -1686,15 +1685,15 @@ function Coder:generate_module()
         table.insert(out, self:lua_entry_point_declaration(f_id) .. ";")
     end
 
-    table.insert(out, section_comment("Function Implementations"))
+    table.insert(out, section_comment("Pallene Entry Points"))
     for f_id = 1, #self.module.functions do
         table.insert(out, self:pallene_entry_point_definition(f_id))
     end
 
+    table.insert(out, section_comment("Lua Entry Points"))
     for f_id = 1, #self.module.functions do
         table.insert(out, self:lua_entry_point_definition(f_id))
     end
-
 
     table.insert(out, self:generate_luaopen_function())
 
@@ -1723,7 +1722,6 @@ function Coder:generate_luaopen_function()
                         str = C.string(upv.str)
                     }))
             elseif tag == "coder.Upvalue.Function" then
-                assert(not self.module.lambdas[upv.f_id])
                 table.insert(init_constants, util.render([[
                     lua_pushvalue(L, globals);
                     lua_pushcclosure(L, ${entry_point}, 1);
@@ -1756,7 +1754,6 @@ function Coder:generate_luaopen_function()
 
     local init_exports = {}
     for _, f_id in ipairs(self.module.exported_functions) do
-        if not self.module.lambdas[f_id] then
             local name = self.module.functions[f_id].name
             table.insert(init_exports, util.render([[
                 lua_pushstring(L, ${name});
@@ -1767,7 +1764,6 @@ function Coder:generate_luaopen_function()
                 name = C.string(name),
                 ix = C.integer(self.upvalue_of_function[f_id]),
             }))
-        end
     end
 
     for _, g_id in ipairs(self.module.exported_globals) do
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1707,6 +1707,7 @@ function Coder:generate_luaopen_function()
     for ix, upv in ipairs(self.upvalues) do
         local tag = upv._tag
         local is_upvalue_box = false
+        local is_lambda = false
         if tag ~= "coder.Upvalue.Global" then
             if     tag == "coder.Upvalue.Metatable" then
                 is_upvalue_box = upv.typ.is_upvalue_box
@@ -1723,18 +1724,21 @@ function Coder:generate_luaopen_function()
                         str = C.string(upv.str)
                     }))
             elseif tag == "coder.Upvalue.Function" then
-                table.insert(init_constants, util.render([[
-                    lua_pushvalue(L, globals);
-                    lua_pushcclosure(L, ${entry_point}, 1);
-                ]], {
-                    entry_point = self:lua_entry_point_name(upv.f_id),
-                    ix = C.integer(self.upvalue_of_function[upv.f_id]),
-                }))
+                is_lambda = self.module.lambdas[upv.f_id]
+                if not is_lambda then
+                    table.insert(init_constants, util.render([[
+                        lua_pushvalue(L, globals);
+                        lua_pushcclosure(L, ${entry_point}, 1);
+                    ]], {
+                        entry_point = self:lua_entry_point_name(upv.f_id),
+                        ix = C.integer(self.upvalue_of_function[upv.f_id]),
+                    }))
+                end
             else
                 typedecl.tag_error(tag)
             end
 
-            if not is_upvalue_box then
+            if not (is_upvalue_box or is_lambda) then
                 table.insert(init_constants, util.render([[
                     lua_setiuservalue(L, globals, $ix);
                     /**/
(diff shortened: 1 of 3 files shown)


hugomg (changes requested):
> It works, but would it be cleaner if we chose to not include the lambdas in exported_functions instead?

Yes, I suspect that might be a better way to do it. Can we try that method?

srijan-paul:
So lambdas are no longer added in exported functions but we still need to declare them up top and define them in a section, so I've kept the module.lambdas list.

hugomg (changes requested):
The logic inside `generate_module` that decides if a function needs to have a Lua entry point is getting more complicated. It generates a Lua entry point if:

1. The function is the `init` function (id = 1)
2. The function has an entry in `upvalue_of_function`, which can happen if:
   * The function is exported
   * The function is used as a higher order function
3. The function is a lambda

Maybe we should take a step back and try to simplify this a bit. What if we added a `has_lua_entry_point` field to the `ir.Function`? Then the coder wouldn't need to be clever about what functions need a Lua entry point. It would also let us avoid creating a forward declaration for functions that don't have a Lua entry point. Currently we create a forward declaration for every function. 

```lua
for f_id = 1, #self.module.functions do
    table.insert(out, self:lua_entry_point_declaration(f_id) .. ";")
end
```

To summarize:
* Don't create a new `module.lambdas` list
* Directly say in the IR which functions have a Lua entry point
* Don't forward declare a Lua entry point for functions that don't have one.

hugomg:
I suppose that the main problem I have with the current code is that the logic for deciding whether a function has a Lua entry point, happens when we are creating the entry point declarations. Which means that we are not applying that logic when creating the forward declarations.

I don't particularly care if this logic should live in the to_ir or the coder, but it should be in a single place and

hugomg:
Oops! There was a comment that I started typing and decided not to post. But then instead of pressing the button to cancel the comment, I accidentally pressed "close with comment". If you got a notification with the half-baked comment, or are confused why this issue is closed then reopened, that is the answer.

srijan-paul:
>  If you got a notification with the half-baked comment, or are confused why this issue is closed then reopened, that is the answer

Aha, alright :D

srijan-paul:
> 
    The function is the init function (id = 1)
    The function has an entry in upvalue_of_function, which can happen if:
        The function is exported
        The function is used as a higher order function
    The function is a lambda

About these, wouldn't we still have to differentiate between:
1. Functions that are lambdas but need a Lua entry point for CallDyn
2. Toplevel module functions that need a Lua entry point.

Since we would want to create closures in luaopen for 2 but not 1.

hugomg:
Perhaps. But afiter we break up the G (as part of the new calling convention) I think we will want to create the toplevel closures in the `init` instead of on the luaopen. In the end, the luaopen should only be responsible for initializing the constant table.

My main impressions so far are:

1. A list of lambdas doesn't feel right, because what the coder cares about is wether a function has a Lua entry point, not whether it is a lambda
2. Currently, we are forward-declaring Lua entry points for functions that don't have Lua entry points.

If things get too complicated, I would even be open to blindly creating a Lua entry point for every function. My order of preference

1. Only create Lua entry points when necessary, assuming it's not too complicated (best)
2. Create Lua entry points for every function, even if they are not needed
3. Complicated logic inside the luaopen, that is inconsistent to boot (worst)

hugomg:
Some interesting examples in the Penlight codebase:

* The `iter` function in `lexer.scan`. There is a single function with many mutable closures. [Link](https://github.com/lunarmodules/Penlight/blob/c3e45ed503f9a6c29a69c21435f357353774eb16/lua/pl/lexer.lua#L178).
* The mutually recursive functions inside `pretty.write`. But the functions don't escape, which means the local variables can be put in the same box. However, would require escape analysis. [Link](https://github.com/lunarmodules/Penlight/blob/c3e45ed503f9a6c29a69c21435f357353774eb16/lua/pl/pretty.lua#L252)

hugomg:
Another interesting pattern is module requires. Module tables are cached by the Lua module system and are guaranteed to outlive any closure. When we implement "require" in Pallene, we should consider storing the module tables in the constant table instead of as regular upvalues like Lua does.

```lua
-- From driver.lua:
local c_compiler = require "pallene.c_compiler"
local checker = require "pallene.checker"
local assignment_conversion = require "pallene.assignment_conversion"
local constant_propagation = require "pallene.constant_propagation"
local coder = require "pallene.coder"
local Lexer = require "pallene.Lexer"
local parser = require "pallene.parser"
local to_ir = require "pallene.to_ir"
local uninitialized = require "pallene.uninitialized"
local translator = require "pallene.translator"
```

I'm mentioning this because the grep scripts we're using to count upvalues don't distinguish between module tables and other upvalues. I don't think we need to spend time to teach the scripts about this, but it is something we should keep in mind when we look at the source code with our own eyes.

srijan-paul:
Currently it's creating lua entry point for every kind of function.
Most of the stuff has been worked out, the only part remaining is this rule here:
> - The function is used as a higher order function

As of now, there is a `has_lua_entry_point` field, which is true for every function.
If we can figure out which toplevel functions are also used has HOFs, we can simply set this field to false later and that would solve the problem.

What I'm wondering is, would that kind of analysis be worth the effort?

hugomg:
All this discussion is making me feel like we should stop bothering this and simply give a Lua entry point to everyone (option 2). 

The only reason to not create the Lua entry point is to make the generated C a bit prettier. Creating unused Lua entry points is not a problem for performance, because the C compiler is able to get rid of them. The only benefit is to make the generated C code a little bit easier to read. However, if that comes at the cost of extra complexity in the code generation then I don't think it's work it!

srijan-paul:
Clean C code would've been good to have :p
But yes, I also realized we'd have to deal with aliasing of functions with local vars etc to make sure there's no HOFs that we leave out. In that case, I think the current implementation can do. I think I'll remove the `has_lua_entry_point` in that case.

pallene/ir.lua
```diff
@@ -33,6 +33,7 @@ function ir.Module()
         globals            = {}, -- list of ir.VarDecl
         exported_functions = {}, -- list of function ids
         exported_globals   = {}, -- list of variable ids
+        lambdas            = {}, -- { function id }
```
hugomg, line 36: If we stop adding lambdas to the `exported_functions` list, can we get rid of this `module.lambdas` map?

pallene/to_ir.lua
```diff
@@ -145,9 +145,12 @@ function ToIR:resolve_variable(decl)
     return var
 end
 
-function ToIR:register_lambda(exp, name)
+function ToIR:register_lambda(exp, name, is_hof)
```
hugomg, line 148: If get rid of the `module.lambdas`, can we also get rid of this `is_hof` argument?

pallene/to_ir.lua
```diff
@@ -856,10 +859,9 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
         table.insert(cmds, ir.Cmd.CheckGC())
 
     elseif tag == "ast.Exp.Lambda" then
-        local f_id = self:register_lambda(exp, "$lambda")
+        local f_id = self:register_lambda(exp, "$lambda", true, true)
```
hugomg, line 862: Currently this function is receiving one argument too many. But would be two, if we get rid of the `is_hof`.

hugomg (changes requested):

pallene/coder.lua
```diff
@@ -1690,12 +1691,11 @@ function Coder:generate_module()
         table.insert(out, self:pallene_entry_point_definition(f_id))
     end
 
-    table.insert(out, section_comment("Exports"))
```
hugomg, line 1693: Maybe we should change section comment, instead of deleting it outright.

pallene/ir.lua
srijan-paul: I'm not sure if we can do that.
In `coder.lua`, we are recursively going through every single function and adding it to the "closures" list. (line 675)

hugomg (approved):
Great!

By the way, after our last phone call I realized that there was another reason why creating a Lua entry point for every function is the correct choice. In the new calling convention every function will need a Lua entry point!

srijan-paul:
Aah, that's a good point!

Issue #413, opened 2021-07-15
New calling convention for Pallene entry point


hugomg:
# Objectives

This issue is about the planned cleanup of the `G` and the new calling convention that would come together with it. The main objectives are:

1. Treat global variables and module functions as regular upvalues
    * They are only an upvalue of closures that actually use them, to avoid memory leaks.
    * We want to remove the special treatment of global vars in the IR and Coder.
2. Potentially speed up calls to functions with many upvalues. (We should measure this)

# The calling convention

The example uses the inner lambda in the following program:

```lua
local function f(): (integer)->integer
    local x = 1
    local y = 2
    return function(z, w)
        return x + y + z + w
    end
end
```

The old calling convention puts all the global stuff in a `G` userdata and passes each upvalue as a separate argument.

```c
// Old calling convention
static lua_Integer function_05(
    lua_State *L,     /* Lua state */
    Udata *G,         /* module variables, strings & metatables*/
    StackValue *base, /* Lua stack*/
    lua_Integer uv1,  /* upvalues*/
    lua_Integer uv2, 
    lua_Integer x1,   /* arguments */
    lua_Integer x2 
) {
```

I propose putting only the strings and metatables in the `G`, which will now be called the constant table  `K`. The global variables will become regular upvalues. Instead of having one function argument per upvalue, the Pallene entry point will receive a pointer to the upvalue array `U`. 

The reason we are currently using separate arguments for the upvalues was in the hope that we might not need to allocate a closure object in the first place. However, I 'm not sure if we really want to avoid these closures allocations... I also worry that all those extra arguments might spill into the C stack. Passing the upvalue array in a single argument might be simpler and more efficient.

```c
// New calling convention
static lua_Integer function_05(
    lua_State *L,     /* Lua state */
    StackValue *base, /* Lua stack*/
    TValue *K,        /* array of module strings & metatables */
    TValue *U,        /* array of upvalues */
    lua_Integer x1,   /* arguments */
    lua_Integer x2
) {
```

# TValue\*

The constant table argument may either be an `Udata *` (as today) or it may be a `TValue *` that points to the upvalue array of the userdata.

Similarly, the upvalue argument may be either a `CClosure *` or it may be a `TValue *` that points to the upvalue array of the closure.

I suspect that if we use `TValue *` the generated C code will look a bit prettier and it might slightly help with the `restrict` stuff. However, I expect that the performance would be similar either way.

# Restrict

[...]

Issue #354, opened 2021-03-06
Optimize CallDyn for Pallene->Pallene calls


hugomg:
Currently, all indirect function calls in Pallene use the Lua calling convention, which passes all arguments and return values via the Lua stack. We should be able to get better performance if we implement a fast path in CallDyn for Pallene-Pallene function calls.

The implementation will need to be able to:
1. determine if a given CClosure is a Pallene closure
2. check at run-time if the Pallene closure has the expected function signature
3. store the Pallene entry point in the CClosure.

The basic idea is to have a run-time check to see if the closure is a Pallene closure of the correct type. If it is, we can call the Pallene entry point, bypassing the Lua API and all the stack manipulation and tag checking that goes with it. Otherwise, we call the Lua entry point. If the closure is a non-Pallene closure then calling the Lua entry point is exactly what we want. If the closure is a Pallene closure with the wrong function signature then calling the Lua entry point for it will raise an appropriate error message, as it does today.

For item (1), one possibility is to have the first upvaule be some special value that only Pallene could put there. One promising idea that was suggested is to have it be a light userdata holding a pointer to the internal upvalue array of the closure. Normal C code that respects the Lua API cannot create such a pointer. There is also the bonus that this is a pointer that the code responsible for the run-time test already has at hand.

For item (2), one possibility is to encode the function signature as a string, using the low level IR type instead of the original Pallene type. A simple implementation would be one character per argument, representing the type tags of the various arguments. For example, a function type of `(string, bool)->integer` could be encoded as `"sb;i"`. A function of type `({integer}, integer->boolean)->{integer}` would have an IR type of `(table, function)->table`, which would be encoded as `"tf;t"`. Remember that at the IR level, there is a single type for all tables (corresponding to the the tag LUA_TTABLE) and a single type for all functions (corresponding to tag LUA_TFUNCTION).

For item (3), the options are either to store the entry point in a light userdata or a light C function. According to POSIX, either would be allowed. However, if we want to strictly follow the C standard then the only option is using a light C function. The standard does not allow casting a function pointer to a `void*`. On the other hand, casting a function pointer to an incompatible function pointer type and then back to the original type is totally kosher. The downside of using light c functions is that if some C code grabs the light C function using lua_getupvalue and tries to call it with lua_call, it will result in undefined behavior.

-------------------------

Finally, some miscellaneous thoughts:

[...]

PR #440, opened 2021-07-30
Pass a single parameter representing upvalues to Pallene entry point.

srijan-paul:
A small change to the calling convention.
Instead of the Lua entry point taking upvalues from `func->upvalue` and passing them to the Pallene entry point as arguments, now it only passes a `TValue*` and the Pallene entry point extracts out the upvalues.
So Instead of:
```c
Udata* uv1 = uvalue(&func->upvalue[1]);
int uv2 = uvalue(&func->upvalue[2]);
function_0x(L, G, uv1, uv2, args...);
```
We now have:
```c
function_0x(L, G, func->upvalue, args...);
```

Meanwhile the corresponding Pallene entry point would look like so:
```c
static void function_0x(lua_State* L, Udata* G, TValue* U, params...) {
    Udata* uv1 = uvalue(&func->upvalue[1]);
    int uv2 = ivalue(&func->upvalue[2]);
    // stuff
}
```

As described in #413.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -53,6 +53,8 @@ function Coder:init(module, modname, filename)
     self.modname = modname
     self.filename = filename
 
+    self.current_func = nil
+
     self.upvalues = {} -- { coder.Upvalue }
     self.upvalue_of_metatable = {} -- typ  => integer
     self.upvalue_of_string    = {} -- str  => integer
@@ -309,7 +311,9 @@ end
 
 -- @returns the C parameter name for the upvalue u_id
 function Coder:c_upval(u_id)
-    return "uv" .. u_id
+    assert(self.current_func)
+    local typ = self.current_func.captured_vars[u_id].decl.typ
+    return lua_value(typ, string.format("&U[%d]", u_id))
 end
 
 -- @returns the C return variable name for the variable ret_i
@@ -354,13 +358,6 @@ function Coder:prepare_local_var(func, v_id)
     return typ, c_name, (p_name and " "..C.comment(p_name) or "")
 end
 
--- The information of a C local var for a given Pallene captured var.
-function Coder:prepare_captured_var(u_id, decl)
-    local c_name = self:c_upval(u_id)
-    local typ    = decl.typ
-    local p_name = decl.name
-    return typ, c_name, (p_name and " "..C.comment(p_name) or "")
-end
 
 --
 -- # Pallene entry point
@@ -426,6 +423,8 @@ function Coder:pallene_entry_point_definition(f_id)
     local func = self.module.functions[f_id]
     local arg_types = func.typ.arg_types
 
+    self.current_func = func
+
     local name_comment = func.name
     if func.loc then
         name_comment = name_comment .. " " .. func.loc:show_line()
@@ -459,17 +458,6 @@ function Coder:pallene_entry_point_definition(f_id)
         table.insert(prologue, decl..initializer..";"..comment)
     end
 
-    for u_id, upval in ipairs(func.captured_vars) do
-        local typ, c_name = self:prepare_captured_var(u_id, upval.decl)
-        local decl = C.declaration(ctype(typ), c_name)..";";
-        local src  = string.format("&U[%d]", u_id)
-        -- Since upvalue boxes do not have metatables, type checking them at runtime is not possible.
-        -- Moreover, since upvalues are only passed around internally by Pallene, it is ok to assume that
-        -- their types will be correct. So we use the `unchecked_get_slot` here instead.
-        local init = unchecked_get_slot(typ, c_name, src)..C.comment(upval.decl.name)
-        table.insert(prologue, decl)
-        table.insert(prologue, init)
-    end
 
     local body = self:generate_cmd(func, func.body)
 
@@ -488,13 +476,13 @@ function Coder:pallene_entry_point_definition(f_id)
     }))
 end
 
-function Coder:call_pallene_function(dsts, f_id, base, xs, upvalue_list)
+function Coder:call_pallene_function(dsts, has_upvalue, f_id, base, xs)
 
     local args = {}
     table.insert(args, "L")
     table.insert(args, "G")
     table.insert(args, base)
-    table.insert(args, upvalue_list or "NULL")
+    table.insert(args, has_upvalue and "func->upvalue" or "NULL")
     for _, x in ipairs(xs) do
         table.insert(args, x)
     end
@@ -537,6 +525,9 @@ function Coder:lua_entry_point_definition(f_id)
     local fname = func.name
     local arg_types = func.typ.arg_types
     local ret_types = func.typ.ret_types
+    local captured_vars = func.captured_vars
+
+    self.current_func = func
 
     -- We unconditionally initialize the G userdata here, in case one of the tag checking tests
     -- needs to use it. We don't bother to make this initialization conditional because in the case
@@ -585,8 +576,9 @@ function Coder:lua_entry_point_definition(f_id)
         table.insert(ret_decls, C.declaration(ctype(typ), ret)..";")
     end
 
-    local call_pallene = self:call_pallene_function(ret_vars, f_id, "L->top", arg_vars,
-        "func->upvalue")
+    local call_pallene = self:call_pallene_function(ret_vars, #captured_vars >= 1, f_id,
+        "L->top", arg_vars)
+
 
     local push_results = {}
     for i, typ in ipairs(ret_types) do
@@ -1403,7 +1395,8 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
     local top = self:stack_top_at(func, cmd)
 
     local parts = {}
-    table.insert(parts, self:call_pallene_function(dsts, cmd.f_id, top, xs))
+    table.insert(parts, self:call_pallene_function(dsts, #func.captured_vars >= 1,
+        cmd.f_id, top, xs))
     table.insert(parts, self:restorestack())
     return table.concat(parts, "\n")
 end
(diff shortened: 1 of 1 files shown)


pallene/coder.lua
```diff
@@ -392,10 +392,10 @@ function Coder:pallene_entry_point_declaration(f_id)
     table.insert(args, {"Udata *"     , "G",    ""})
     table.insert(args, {"StackValue *", "base", ""})
 
-    for i, upval in ipairs(func.captured_vars) do
-        local typ, c_name, comment = self:prepare_captured_var(i, upval.decl)
-        table.insert(args, {ctype(typ), c_name, comment})
+    if #func.captured_vars >= 1 then
+        table.insert(args, {"TValue*", "U", C.comment("upvalues")})
```
hugomg, line 396: Having the `U` argument be optional migth be a bit awkward when we implement indirect Pallene calls in #354. I think we'll have 3 options in that case:

1. The function doing the indirect Pallene call will have to include an if statement to see if it calls passing the U or not (urgh!)
2. We'll need to create an "calldyn pallene entry point", which always receives the U, and which lives between the lua entry point and the "callstatic pallene entry point"
3. We could have the pallene entry point always receive the U parameter.


For simplicity, I'm inclined to go with option (3) at first, and then open a github issue asking if it would be better to switch to option (2). Since option (2) is more complicated, I would only like to do it if we have benchmarks saying it's faster.

pallene/coder.lua
```diff
@@ -462,6 +462,15 @@ function Coder:pallene_entry_point_definition(f_id)
...
     end
 
+    for u_id, upval in ipairs(func.captured_vars) do
+        local typ, c_name = self:prepare_captured_var(u_id, upval.decl)
+        local decl = C.declaration(ctype(typ), c_name)..";";
+        local src  = string.format("&U[%d]", u_id)
+        local init = unchecked_get_slot(typ, c_name, src)..C.comment(upval.decl.name)
+        table.insert(prologue, decl);
+        table.insert(prologue, init);
+    end
```
hugomg, line 472: Instead of eagerly fetching all the upvalues at the start of the function, could we instead fetch them lazily, only when we use them? This is what we currently do with the stuff in the `G`. See the `ir.Value.Function` and the `GetGlobal` for example.

If we say that the `U` is a `restrict` pointer, I think the C compiler should be able to optimize it to only fetch the upvalue once, even if we are fetching it "lazily".

before:

    Udata* uv1 = uvalue(&U[1]);
    int uv2 = ivalue(&U[2]);
    stuff(uv1, uv2);

after:

    stuff(uvalue(&U[1]), ivalue(&U[2]));

pallene/coder.lua
```diff
@@ -462,6 +462,15 @@ function Coder:pallene_entry_point_definition(f_id)
         table.insert(prologue, decl..initializer..";"..comment)
     end
 
+    for u_id, upval in ipairs(func.captured_vars) do
+        local typ, c_name = self:prepare_captured_var(u_id, upval.decl)
+        local decl = C.declaration(ctype(typ), c_name)..";";
+        local src  = string.format("&U[%d]", u_id)
+        local init = unchecked_get_slot(typ, c_name, src)..C.comment(upval.decl.name)
```
hugomg, line 469: We should keep the comment explaining why the unchecked_get_slot is safe.

hugomg (changes requested):
Another thing that we could do is swap the order of the `G` and the `base`. That should be simple enough that we don't have to put on a separate PR.

pallene/coder.lua
srijan-paul: I agree overall, but is it okay to pass `NULL` when we're doing a "CallStatic" for instance?

pallene/coder.lua
srijan-paul: Is the C compiler guaranteed to optimize this? (I'm not well aware :p)
If not, then maybe we could use a benchmark.

hugomg: I agree that some experiments wouldn't hurt. It might be more direct to look at the generated assembly instead of a performance benchmark, though. In #413 I link to some [tests](https://godbolt.org/z/zPhTja9sG) I ran on a simplified non-Pallene program. It suggested that, with restrict, Clang could optimize something similar to this but gcc did not.

That said, my initial instinct is to be optimistic and try to leave the optimization to C compiler. I also get a bit nervous creating all those extra variables; they add pressure to the register allocator. One thing that I like about the new calling convention is that it doesn't have as many variables (although admitedly I don't have hard performance data about that).

pallene/coder.lua
hugomg: That's another idea, let's call it option (4) :)

hugomg: But for this PR I think we should still go with option (3), which is the simplest. We can reconsider options 2,3,4 in a future PR.

pallene/coder.lua
srijan-paul: Sounds good enough to me!
I don't know the internals of C compilers to make a more educated guess. I couldn't have figured out the problem with the register allocator you mentioned, so I'll believe this works the way we think it does.

pallene/coder.lua
srijan-paul: I went with (4) for now, should I revert it back to (3). (Is there a benefit to doing that?)

pallene/coder.lua
srijan-paul: This might require some changes, since we'll have to know the type of an upvalue id. Would it be ok to maintain a map of `upvalue_id -> type` for each `ir.Function`?

hugomg: The only thing you need to know about the register allocation, is that in assembly language there are a finite number of CPU registers. If two C variables are "alive" at the same time, then they can't be stored in the same CPU register. If at some point in the program there are too many variables "alive", some have to be stored in the stack instead of in registers (spilling).

From this point of view, it's not ideal to initialize a lots of variables at the start of the function, only to use their value much later. That results in many variables with overlapping lifetimes. If we decide that we can't trust the C compiler to optimize the upvalues, I expect that it would be better to create the local variables closer to the point where they will be used, instead of at the start of the function. But that's more complicated than doing it the lazy way ;)

pallene/coder.lua
hugomg: I suppose that the only benefit of 3 instead of 4 is being marginally simpler, one less conditional.

From the point of view of code review, the only thing I have a stronger opinion about is that I don't like the default argument in `call_pallene_function`. For example, it's weird that the argument order of the `xs` and the `upvalue_list` is different than the order they appear in the pallene entry point.

pallene/coder.lua
hugomg: > This might require some changes, since we'll have to know the type of an upvalue id

I expect that it would work similar to what we're already doing. I think we're getting the type from the `UpvalInfo` (captured_vars)?

srijan-paul: The reason I think we may need a map is because upvalues can be referenced in all kinds of places with `ir.Value.Upvalue(id)`, so for a given upvalue ID, we'd want to know what type it is and do something similar to `c_value` by casting it to the correct type (from U[id]).

hugomg: Exactly. But I think we already have that map? As I said before, I think it is the `captured_vars`.

srijan-paul: Yeah, that's what I meant xD
Now I see what caused the confusion. I should have phrased it better.

pallene/coder.lua
```diff
@@ -1408,7 +1405,7 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
     local top = self:stack_top_at(func, cmd)
 
     local parts = {}
-    table.insert(parts, self:call_pallene_function(dsts, cmd.f_id, top, xs))
+    table.insert(parts, self:call_pallene_function(dsts, false, cmd.f_id, top, xs))
```
hugomg, line 1408: I don't understand why the callstaic always says the function has no upvalues.

pallene/coder.lua
srijan-paul: Would it be okay to introduce a second parameter to `c_value` for this purpose (`func`)? That way c_value can pass that to c_upval and c_upval and return the appropriately casted string.

pallene/coder.lua
srijan-paul: Ah, I didn't request a review because I was still experimenting.
(If I did request a review by mistake, sorry about that, I also haven't fixed the eager upvalue fetching yet)

pallene/coder.lua
hugomg: I'd hate to need to pass an extra func parameter to the hundreds of places where we call c_value. Maybe we should store that in the `self`, as an instance variable of the Coder class.

Currently, we're operating under the assumption that all the upvalues are living in the `G`, and that all functions in the module have the same upvalues. That's why the `upvalue_of_foo` instance variables are defined at the module level... But now that each function has a different set of upvalues, I think it makes sense to keep a reference to the current function as an instance variable...

In a later PR we can refactor the places that currently receive the `func` as a parameter, and also switch them to using an instance variable.

pallene/coder.lua
hugomg: Don't worry. Sometimes I think you requested a review even if you haven't. Those github notifications all look the same 🤷

pallene/coder.lua
```diff
@@ -309,7 +311,9 @@ end
 
 -- @returns the C parameter name for the upvalue u_id
 function Coder:c_upval(u_id)
-    return "uv" .. u_id
+    assert(self.current_func)
+    local typ = self.current_func.captured_vars[u_id].decl.typ
+    return lua_value(typ, string.format("&U[%d]", u_id))
```
hugomg, line 316: Did we accidentally delete the comment that justifies why it is safe to unbox the upvalue without a tag check?

hugomg (changes requested):

hugomg (approved):
Looks good! Could you please rebase it?

srijan-paul:
> Looks good! Could you please rebase it?

Done 👍

PR #445, opened 2021-08-02
Limit upvalue and local-var count to 200

### LINKED ISSUE #411: Add a limit for the number of captured variables
Suggested limit: 200 captured variables / upvalues

The Lua interpreter has a limit of at most 255 upvalues in a C closure (see the nupvalues in the ClosureHeader of lobject.h). Currently, we run into an `assert` in the coder if this limit is reached. We should have a friendly user-facing error instead.

Maybe the right place to do this is in the `to_ir`? The answer might also change if we decide to change how we represent the `G`.

srijan-paul:
This PR attempts to fix #411. 

Upvalue limits are enforced in the `to_ir` pass, because we do not have enough information about upvalues to report an error in any of the prior passes.

In contrast, local variable limits are enforced in the type checking pass because by the time we get to the IR, it gets difficult to differentiate between local variables inserted by the compiler, and those mentioned by the user. One way to do it there could be:

When calling `add_local`, check the name of the variable being added, if it is falsy or starts with "$", then it's an internal compiler variable. However, the checker already maintains a symbol table, so it took very few changes to get it working in that pass.

The IR generator is now capable of throwing errors like the previous passes.
There is also an accompanying "to_ir_spec" (with just one test at the moment).

### DIFF
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -31,6 +31,9 @@ declare_type("Var", {
 
 local ToIR = util.Class()
 
+-- The Lua interpreter uses a byte sized number to refer to upvalues, and
+-- therefore supports a maximum of 255. To make sure we don't overflow that limit,
+-- we enforce a max upvalue count of 200 for Pallene.
 local MaxUpvalueCount = 200
 
 local function ir_error(loc, fmt, ...)
@@ -44,7 +47,6 @@ function to_ir.convert(prog_ast)
         return ToIR.new():convert_toplevel(prog_ast.tls)
     end)
 
-    local ir_module
     if not ok then
         if ret.tag == "to_ir" then
             return false, { ret.msg }
@@ -53,7 +55,7 @@ function to_ir.convert(prog_ast)
         end
     end
 
-    ir_module = ret
+    local ir_module = ret
     ir.clean_all(ir_module)
     return ir_module, {}
 end
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -7,6 +7,7 @@ local ir = require "pallene.ir"
 local types = require "pallene.types"
 local util = require "pallene.util"
 local typedecl = require "pallene.typedecl"
+local trycatch = require "pallene.trycatch"
 
 local to_ir = {}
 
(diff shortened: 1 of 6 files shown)


pallene/to_ir.lua
```diff
@@ -30,11 +31,31 @@ declare_type("Var", {
...
 function to_ir.convert(prog_ast)
     assert(prog_ast._tag == "ast.Program.Program")
-    local module = ToIR.new():convert_toplevel(prog_ast.tls)
-    ir.clean_all(module)
-    return module, {}
+    local ok, ret = trycatch.pcall(function()
+        return ToIR.new():convert_toplevel(prog_ast.tls)
+    end)
+
+    local ir_module
```
hugomg, line 47: We don't need to forward-declare this variable.

pallene/to_ir.lua
```diff
@@ -30,11 +31,31 @@ declare_type("Var", {
 
 local ToIR = util.Class()
 
+local MaxUpvalueCount = 200
```
hugomg, line 34: We should have a commend explaining why this limit is needed.

hugomg (changes requested):
The limit on upvalues is important, because Lua can't handle closures with more upvalues than that.

However, do we really need to limit the local variables? Lua does this because the bytecode can only reference 255 variables. However, Pallene doesn't have this restriction. 

I also had the impression that we had to write some unnatural-looking code in order to implement the local var limit, checking if the variable name has `$` and so on. I'm inclined to  leave checker.lua and symtab.lua unmodified, not limiting the number of local variables.

srijan-paul:
> However, do we really need to limit the local variables?

I agree that the code that enforces this limit is a bit unconventional.
I was thinking about the Lua backend. I understand that transpiling to Lua is not the primary purpose of Pallene, but it could perhaps be good to at least throw in a "warning" when the user transpiles such code to Lua?

Or we could just not enforce this limit, like you said.
What do you think?

hugomg:
If we want to implement such a warning, perhaps it would be more appropriate to implement it in the Lua backend instead of in the checker. I don't think we need to restrict how many variables the C backend can accept, just because of a limitation of the Lua backend.

However, because programs with more than 200 local variables are so rare, I don't know if implementing the warning is worth the trouble. The status quo is that the Lua interpreter complains when it tries to run the generated code, which is not that bad. This is different from the status quo for too many upvalues, where the Pallene compiler would crash with an incomprehensible error message.

hugomg (approved):
Great!

There diff has a couple of blank lines that we don't need, but I can take care of that.

srijan-paul:
Thanks!

Issue #414, opened 2021-07-15
Use CallStatic for non-toplevel functions


hugomg:
Currently, I believe we only use CallStatic for toplevel functions. We should also be able to use it for non-toplevel functions, if the said function is never reassigned. 

There are a couple of situations where this can happen:
1. The function is declared using a function statement
    ```lua
    function f() ... end
    ```
2. A local variable, which is never reassigned, is initialized with a lambda expression
    ```lua
    local f = function() ... end
    ```
3. A function call is directly calling a lambda expression
    ```lua
    (function (x,y,x) ... end)(1,2,3)
    ```

I think that if we implement at least the first case, that would be OK. Optimizing the second case would be a nice bonus, if it's not too complicated.

Optimizing things like the third case could be done in theory. However, perhaps in this kind of case it would be better to inline the function instead.

It is possible that we may need to implement #413 before this, but I am not sure.

srijan-paul:
Revisiting this issue, I'm curious about the first case.
Functions declared as `local function f() ... end` are already using `CallStatic`, and function statements without forward declarations aren't permitted, so is it not already optimized?

hugomg:
You might be right. I'll try to double check next week. This week I'm busy with something else, unfortunately.

hugomg:
I checked again and found that

1. `local function f` is is called with CallStatic
2. `local f = function` is called with CallDyn

I will leave this issue open until we optimize case 2, but it's a low priority because I expect that most programmers would write the first version. Let's not worry about case 3, because we're not Javascript.

PR #450, opened 2021-08-07
Fixed bug with missing 'func'

### LINKED ISSUE #448: `func` doesn't always exist in the Pallene entry point
The following program produces a C compiler error. It might be related to #440 and/or #414... We should find out what is the problem is and why the current tests did not catch it.
 

```lua
local m = {}

record Stream
    first: integer
    rest: () -> any
end

local function count_from(n: integer): Stream
    return {
        first = n,
        rest = function() return count_from(n+1) end,
    }
end

return m
```

```
/tmp/lua_cn9Wzz: In function ‘function_03’:
/tmp/lua_cn9Wzz:129:38: error: ‘func’ undeclared (first use in this function); did you mean ‘Pfunc’?
  129 |     x2 = function_02(L, base + 0, G, func->upvalue, x3);
      |                                      ^~~~
      |                                      Pfunc
/tmp/lua_cn9Wzz:129:38: note: each undeclared identifier is reported only once for each function it appears in
internal error: compiler failed
compilation line: cc -I./vm/src -I./runtime -std=c99 -g -fPIC -Wall -Wundef -Wpedantic -Wno-unused -O2 -fverbose-asm -O2  -x c -o '/tmp/lua_ZIUovs' -S '/tmp/lua_cn9Wzz'
```

The error happens inside the Pallene entry point for the lambda:

```c
/* $lambda foo.pln:11 */
static TValue function_03(
    lua_State *L,
    StackValue *base,
    Udata *G,
    TValue * restrict U /* upvalues */
) {
    
    TValue x1 = {{0},0};
    Udata *x2;
    lua_Integer x3;
    
    x3 = intop(+, ivalue(&U[1]), 1);
    x2 = function_02(L, base + 0, G, func->upvalue, x3);   // <---- HERE
    base = restorestack(L, base_offset);
    setuvalue(L, &x1, x2);
    return x1;
}
```

srijan-paul:
Fixes #448 
`CallStatic` is always used in pallene-to-pallene calls. It is worth noting that currently all calls to closures that may have upvalues is always done via CallDyn, so we can safely pass `NULL` in every CallStatic for the `TValue* U` parameter.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -493,7 +493,6 @@ function Coder:call_pallene_function(dsts, f_id, base, xs, cclosure)
     -- we can simply pass NULL as it's `upvalues` parameter as it will be unused anyway.
     local upvals
     if n_upvalues >= 1 then
-        assert(cclosure)
         upvals = cclosure.."->upvalue"
     else
         upvals = "NULL"
@@ -1411,15 +1410,9 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
 
     local parts = {}
 
-    -- If the function being called is a module function, then it has a cclosure
-    -- assosciated with it. Toplevel local functions cannot be called with `CallStatic`
-    -- as of now.
-    local upvalues = false
-    if self.upvalue_of_function[cmd.f_id] then
-        local cclosure = self:function_upvalue_slot(cmd.f_id)
-        upvalues = string.format("clCvalue(%s)->upvalue", cclosure)
-    end
-    table.insert(parts, self:call_pallene_function(dsts, cmd.f_id, top, xs, upvalues))
+    local f_id = cmd.src_f.id
+    local cclosure = string.format("clCvalue(%s)", self:function_upvalue_slot(f_id))
+    table.insert(parts, self:call_pallene_function(dsts, f_id, top, xs, cclosure))
     table.insert(parts, self:restorestack())
     return table.concat(parts, "\n")
 end
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -479,7 +479,7 @@ function Coder:pallene_entry_point_definition(f_id)
     }))
 end
 
-function Coder:call_pallene_function(dsts, f_id, base, xs, cclosure)
+function Coder:call_pallene_function(dsts, f_id, base, cclosure, xs)
 
     local func       = self.module.functions[f_id]
     local n_upvalues = #func.captured_vars
@@ -591,7 +591,7 @@ function Coder:lua_entry_point_definition(f_id)
         table.insert(ret_decls, C.declaration(ctype(typ), ret)..";")
     end
 
-    local call_pallene = self:call_pallene_function(ret_vars, f_id, "L->top", arg_vars, "func")
+    local call_pallene = self:call_pallene_function(ret_vars, f_id, "L->top", "func", arg_vars)
 
 
     local push_results = {}
@@ -1412,7 +1412,7 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
 
     local f_id = cmd.src_f.id
     local cclosure = string.format("clCvalue(%s)", self:function_upvalue_slot(f_id))
-    table.insert(parts, self:call_pallene_function(dsts, f_id, top, xs, cclosure))
+    table.insert(parts, self:call_pallene_function(dsts, f_id, top, cclosure, xs))
     table.insert(parts, self:restorestack())
     return table.concat(parts, "\n")
 end
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -479,13 +479,25 @@ function Coder:pallene_entry_point_definition(f_id)
     }))
 end
 
-function Coder:call_pallene_function(dsts, has_upvalue, f_id, base, xs)
+function Coder:call_pallene_function(dsts, f_id, base, xs, cclosure)
+
+    local func       = self.module.functions[f_id]
+    local n_upvalues = #func.captured_vars
 
     local args = {}
     table.insert(args, "L")
     table.insert(args, base)
     table.insert(args, "G")
-    table.insert(args, has_upvalue and "func->upvalue" or "NULL")
+
+    local upvals
+    if n_upvalues >= 1 then
+        assert(cclosure)
+        upvals = cclosure.."->upvalue"
+    else
+        upvals = "NULL"
+    end
+    table.insert(args, upvals)
+
     for _, x in ipairs(xs) do
         table.insert(args, x)
     end
@@ -528,7 +540,6 @@ function Coder:lua_entry_point_definition(f_id)
     local fname = func.name
     local arg_types = func.typ.arg_types
     local ret_types = func.typ.ret_types
-    local captured_vars = func.captured_vars
 
     self.current_func = func
 
@@ -579,8 +590,7 @@ function Coder:lua_entry_point_definition(f_id)
         table.insert(ret_decls, C.declaration(ctype(typ), ret)..";")
     end
 
-    local call_pallene = self:call_pallene_function(ret_vars, #captured_vars >= 1, f_id,
-        "L->top", arg_vars)
+    local call_pallene = self:call_pallene_function(ret_vars, f_id, "L->top", arg_vars, "func")
 
 
     local push_results = {}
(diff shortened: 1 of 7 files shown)


pallene/coder.lua
```diff
@@ -1398,8 +1398,12 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
     local top = self:stack_top_at(func, cmd)
 
     local parts = {}
-    table.insert(parts, self:call_pallene_function(dsts, #func.captured_vars >= 1,
-        cmd.f_id, top, xs))
+
+    -- CallStatic is used in pallene-pallene calls. Any calls to higher order functions or
+    -- closures is always done using CallDyn at the moment, meaning CallStatic will never call
+    -- a closure with upvalues.
```
hugomg, line 1404: I'm not sure that this assumption will still hold after next week. Once we split the `G`, the global variables will become upvalues.

Instead of the `call_pallene_function` receiving a `has_upvalue`, it should receive a string with a `C` expression for the `CClosure*`. In the Lua entry point that is the `func`, but in the Pallene entry point it will be something else.  (Currently, the `"func"` in the `call_pallene_function` assumes that we're inside a Lua entry point, which doesn't smell right)

I think that the `call_pallene_function` should be able to know how many upvalues the function has, based on the `f_id`. We might be able to do the NULL optimization inside the `call_pallene_function`, instead of in the places that call it.

hugomg (changes requested):
I have a feeling that this solution will break after we split the `G`. Do you agree?

srijan-paul:
> I have a feeling that this solution will break after we split the G. Do you agree?

Yes, since toplevel functions would also be closures.

pallene/coder.lua
```diff
@@ -479,13 +479,25 @@ function Coder:pallene_entry_point_definition(f_id)
...
+    local func       = self.module.functions[f_id]
+    local n_upvalues = #func.captured_vars
 
     local args = {}
     table.insert(args, "L")
     table.insert(args, base)
     table.insert(args, "G")
-    table.insert(args, has_upvalue and "func->upvalue" or "NULL")
+
+    local upvals
```
hugomg, line 492: If I were not familiar with this optimization, I would have guessed that the reason for the NULL is because the upvalue array doesn't exist when there are no captured variables. Which is not the real reason...

Maybe we should add a comment saying that this NULL thing an optimization.

pallene/coder.lua
```diff
@@ -479,13 +479,25 @@ function Coder:pallene_entry_point_definition(f_id)
...
+    if n_upvalues >= 1 then
+        assert(cclosure)
```
hugomg, line 494: If the `cclosure`  is no longer optional (see my other comments) then I think we should also remove this assert.

pallene/coder.lua
hugomg: At some point, the call static functions certainly will start having upvalues too. Either after we split the `G`, or after we implement #414, whichever comes first.

At the very least, we need to rewrite this comment. While this might be true today, it certalny won't be true anymore in the near future. However, I feel that maybe we should bite the bullet and try to allow upvalues here as well.

The impression I'm having here is that the only reason we added the assert, is to avoid having to come up with a reference to the cclosure. However, getting that cclosure is going to be a pre-requesite for #413. I suppose we might as well do it now...

pallene/coder.lua
```diff
@@ -1398,8 +1408,12 @@ gen_cmd["CallStatic"] = function(self, cmd, func)
...
 
     local parts = {}
-    table.insert(parts, self:call_pallene_function(dsts, #func.captured_vars >= 1,
-        cmd.f_id, top, xs))
+
+    -- CallStatic is used in pallene-pallene calls. Any calls to higher order functions or
+    -- closures is always done using CallDyn at the moment, meaning CallStatic will never call
+    -- a closure with upvalues.
+    assert(#self.module.functions[cmd.f_id].captured_vars == 0)
+    table.insert(parts, self:call_pallene_function(dsts, cmd.f_id, top, xs))
```
hugomg, line 1416: I recommend using optional arguments sparingly. In a dynamic language, it's harder to tell if a missing argument is intentional or if it is an accident. In this case I would prefer passing an explicit `false` for the missing argument.

That said, as I said elsewhere, I think it would be even better if the `cclosure` argument were always a string. We want to allow CallStatic for functions that have captured variables.

hugomg (changes requested):

pallene/coder.lua
srijan-paul: About the `CallStatic` being able to pass upvalues, we'd also have to be able to find a toplevel function's `upvalue` list, provided we know it's `id`. To that end, I have a small question:

Is it always guaranteed that  `function_n` is the `n`th value associated with the `globals` userdata?
If so, then what do you think about exposing the index `globals` and then getting the cclosure from that table?
I am also wondering the amount of overhead this might add to calling toplevel functions that capture globals as upvalues, since we're doing an extra push and pop operation.

```c
int type = lua_getiuservalue(L, globals, n);
assert(type == LUA_TFUNCTION);
// get the `upvalue` list from stack top, then pop from the stack?
lua_pop(L, 1)
```

hugomg: > Is it always guaranteed that function_n is the nth value associated with the globals userdata?

I think we can get this info from `upvalue_of_function`.

I don't remember if we are already creating a closure for every toplevel function. In the past we did not, but I don't remember if we recently changed this.

> I am also wondering the amount of overhead this might add to calling toplevel functions that capture globals as upvalues, since we're doing an extra push and pop operation.

We should't need to use a push and pop operation to get a value from the `G`. We can access the uservalue array directly.

Look at the places in the code that use with ir.Value.Function to see how we're currently dealing with this.

srijan-paul: > I don't remember if we are already creating a closure for every toplevel function. 

We don't create a `CClosure` object for toplevel functions declared like `local function foo`, but we do for `local foo: () -> () = ...`.  I realize now that it might be better to start by fixing this inconsistency first.

However, once that's done, we would still only be able to refer to local functions inside of `$init`.

srijan-paul:
To truly support `CallStatic` being capable of passing cclosures, we might have to implement the `G` splitting first.
I think for this PR we could:
- Only pass the upvalue list for module functions in `gen_cmd["CallStatic"]`. Which won't be used either way I believe.
- For toplevel local functions, we have to first decide upon a consistent behaviour. I think that will mean creating `CClosure`s for them in $init. In other words, treating them as regular toplevel locals?
- Once all toplevel local functions have a cclosure assosciated with them, we have to make sure that other functions can find them, by capturing them as upvalues. This is where it might make sense to do the `G` splitting first.

pallene/coder.lua
```diff
@@ -479,13 +479,26 @@ function Coder:pallene_entry_point_definition(f_id)
     }))
 end
 
-function Coder:call_pallene_function(dsts, has_upvalue, f_id, base, xs)
+function Coder:call_pallene_function(dsts, f_id, base, xs, cclosure)
```
hugomg, line 482: What do you think about reordering the arguments of `call_pallene_function` so it's more similar to the order of the entry point? Now that `cclosure` is no longer optional, it doesn't have to be the last argument.

pallene/print_ir.lua
```diff
@@ -217,7 +217,7 @@ local function Cmd(cmd)
     elseif tag == "ir.Cmd.NewRecord"  then rhs = "new ".. cmd.rec_typ.name .."()"
     elseif tag == "ir.Cmd.GetField"   then rhs = Field(cmd.src_rec, cmd.field_name)
     elseif tag == "ir.Cmd.SetField"   then rhs = Val(cmd.src_v)
-    elseif tag == "ir.Cmd.CallStatic" then rhs = Call(Fun(cmd.f_id),  Vals(cmd.srcs))
+    elseif tag == "ir.Cmd.CallStatic" then rhs = Call(Fun(cmd.src_f.id),  Vals(cmd.srcs))
```
hugomg, line 220: I get nervous when I see two field accesses in a row, without checking the type tag... In this case we might be able to avoid this if we use `Val(cmd.src_f)` instead of `Fun(cmd.src_f.id)`.

hugomg (changes requested):
Looks good! I only have two small requests....

Next step after those will be the SetUpvalue PR, right?

pallene/coder.lua
srijan-paul: Right. That's a good point.

hugomg (approved):
Perfect!

PR #452, opened 2021-08-10
Introduce `SetUpvalue` instruction in the IR

srijan-paul:
Prior to this, a list of to-be-captured upvalues was passed to the `ir.Cmd.NewClosure` instruction.
Now, a `SetUpvalue` command is introduced, and the labor of adding the upvalues to the CClosure is represented in the IR.

However, this also means that we will be extracting the ccl out of the TValue every single time.
I believe there is some room for optimization here, but I wonder if we should do that after #413 is taken care of.

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1366,14 +1366,6 @@ end
 gen_cmd["NewClosure"] = function (self, cmd, _func)
     local func = self.module.functions[cmd.f_id]
 
-    local capture_upvalues = {}
-    for i, upval_info in ipairs(func.captured_vars) do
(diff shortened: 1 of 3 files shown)


srijan-paul:
I just realized the diff is based off of the current HEAD and therefore contains some changes already reviewed in PR 451.
Its probably best to not look until that PR is merged and I've rebased this on master :p

hugomg (approved):
There's a chance that re-extracting the cclosure might not be a problem. Since the generated code for `SetUpvalue` does not contain any function calls, maybe the C compiler might be able to optimize that by itself. I'll open a separate issue for that.

PR #455, opened 2021-08-10
Remove obsolete "import" token

hugomg:
The current plan is to use `local foo = require` instead of `import`.

The diff might be slightly easier to read by looking at one commit at a time.

### DIFF
--- a/pallene/Lexer.lua
+++ b/pallene/Lexer.lua
@@ -49,7 +49,7 @@ local is_keyword = {}
 do
     local strs = [[
         and break do else elseif end for false function goto if in local nil not or repeat return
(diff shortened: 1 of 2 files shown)


pallene/Lexer.lua
```diff
@@ -49,7 +49,7 @@ local is_keyword = {}
 do
     local strs = [[
         and break do else elseif end for false function goto if in local nil not or repeat return
-        then true until while   as import record typealias
+        then true until while   as record typealias
```
srijan-paul, line 52: The double whitespace is intentional I guess.

srijan-paul (approved):

PR #456, opened 2021-08-12
Split G into U and K

srijan-paul:
Prior to this, G was a userdata table from which all globals were accessed.
After this PR, it will be renamed to `K` to represent the global constants table, and all toplevel `local` and module variables will be captured as upvalues instead.


srijan-paul:
This is a draft PR for now, I'll request a review once I have cleaned it up. 😅 
Currently I'm trying to make this work with minimum changes to the Coder.

Ideally, I want to avoid special casing the codegen for $init.

PR #457, opened 2021-08-13
Capturing non-toplevel FuncStats as upvalues.

srijan-paul:
Non-toplevel function statements of the form `local function f() end` can now be captured as upvalues.
Instead of using `ir.Value.Function` to refer to them, the IR now uses `ir.Value.LocalVar` or `ir.Value.Upvalue`.

Some tests entail this change. To ensure that function statements cannot be re-assigned to, and that the new representation is working well.

One thing that still needs to be addressed in this PR is mutually recursive closures.

If we try the following script:

```lua

function m.oddeven(n: integer): string
    local l_even, l_odd
    function l_even(x: integer): boolean
        if x == 0 then
            return true
        else
            return l_odd(x-1)
        end
    end
    function l_odd(x: integer): boolean
        if x == 0 then
            return false
        else
            return l_even(x-1)
        end
    end
    if l_even(n) then return "even" end
    return "odd"
end

```

We get an error saying 'l_odd' is used before initialization.


pallene/to_ir.lua
```diff
@@ -83,6 +83,7 @@ function ToIR:init()
     self.func_stack     = {} -- list of function to_ir.FuncInfo
     self.call_exps      = {} -- { ast.Exp.CallFunc }
     self.dsts_of_call   = {} -- { ast.Exp => { var_id } }
+    self.toplevel_funcs = {} -- { ast.FuncStat.FuncStat }
```
srijan-paul, line 86: This is a small hack for now. Since we will be getting rid of `ir.Value.Function` down the line anyway.
This table separates toplevel functions from non-toplevel functions.

srijan-paul:
I clicked request review again by mistake 😩 
I meant to do that after I solve the issue with recursive closures. So this is a WIP at the moment :|

srijan-paul:
Mutual recursion is fixed, however that part of the code is becoming more and more sensitive. 
Any change to the order/invariants might break stuff. But I also don't know if there are any significant refactors that can be made.

pallene/to_ir.lua
```diff
@@ -624,8 +642,40 @@ function ToIR:convert_stat(cmds, stat)
...
+            local exp = func.value
+            self:convert_func(exp)
+            if not self.toplevel_funcs[func] then
+                local dst  = self.loc_id_of_decl[func]
+                local f_id = self.fun_id_of_exp[exp]
+                table.insert(cmds, ir.Cmd.NewClosure(exp.loc, dst, f_id))
+            end
+        end
+
+        -- To support mutual recursion, upvalues are captured *after* the closures
```
hugomg, line 668: I would rephrase this as "initialized" instead of "captured".

hugomg (changes requested):
The code isn't the prettiest, but I don't see anything that makes me nervous. I think it's fine, specially if we consider that some of the hacky things should become less hacky once we finish the G split.

The only question I have is about the `closure-tests` file. We were planning to get rid of it in this PR. Can we already do that should we leave it for another PR?

srijan-paul:
> Can we already do that or should we leave it for another PR?

Ah yea, we can do that. done 👍

hugomg (approved):

hugomg:
Thanks!

Issue #221, opened 2020-05-31
Builtin functions should work with a `.` in the name


hugomg:
For example, instead of `io_write` and `math_sqrt` we should be able to write `io.write` and `math.sqrt`. Doing this for builtins will pave the way for allowing the same for non built-in modules in the future.

hugomg:
Today I went through the code to develop a plan of how we could implement this. I think I have a basic plan that might work.

For the parse we don't need to do anything. Something like `io.write` can continue to be parsed as an `ast.Var.Dot`.

```
{
    _tag = "ast.Var.Dot",
    exp = {
       _tag = "ast.Exp.Var", 
       var = {  _tag = "ast.Var.Name",  name = "io" }  },
    name = "write"
}
```

The type checker could then transform that `ast.Var.Dot` into a flat `ast.Var.Name`. This way the `to_ir` and the `coder` won't need to be modified and the changes can be restricted to the type checker.

```
{
    _tag = "ast.Var.Name",
    name = "io.write"
}
```

To do this, we will need to change the behavior of the `check_var` method. Instead of only annotating the var node with a `_type` and `_name` field, it will also have to return the (potentially modified) var node, similarly to how check_exp does it. That is, instead of just doing a `self:check_var(exp.var)`, the callers will have to do a `exp.var = self:check_var(exp.var)`.

We might want to add a `Module` case to the `checker.Name` tagged union, to represent names such as `io` and `string` that correspond to a module name. These names should only be allowed to be part of a Var.Dot. If someone tries to do `local x = io` then it should be a compile-time error, similarly to what happens when trying to assign a type name to a local variable.

I think this should cover most of it but if something unexpected happens then we can figure it out.

hugomg:
Implemented in #224

Issue #85, opened 2019-07-04
Recursive types


hugomg:
Currently, Pallene does not allow types to be recursive. The existing workaround is to use a `value` type where the recursion happens:

```
record List
   head : integer
   tail : value
```

In addition to not being clear about the intention that the `tail` should be a List, the value type here also allows users to write invalid values to the tail of the list node.

#### Obstacles to recursive types

Before we add sum types to the language, we would need to add a way to describe sum types, or a least nillable types. Without this it would be impossible to construct a list of finite length.

Furthermore, we would need to figure out what are the correct rules to implement type equality and consistency for recursive types. I imagine that we would want to make sure our recursive types are iso-recursive and not equi-recursive.

#### Mutually-recursive types

Once we figure out recursive types, it would be desirable to also allow mutually-recursive types. 
The syntax could be similar to the one I proposed for #84, that is, consecutive type declarations would be mutually-recursive. If we did this we would need to be careful to forbid invalid infinite loops with type aliases:

```
type a = b
type b = a
```

hugomg:
> Before we add sum types to the language, we would need to add a way to describe sum types, or a least nillable types.

Type unions / sums are a requirement for recursive data structures. However, there are some uses of recursive types that don't involve recursive data structures. For example, the following type which represents an infinite list of numbers.

```
record IntStream
    head: integer
    tail: () -> IntStream
end
```

PR #466, opened 2021-08-18
Add Stream Sieve benchmark from Typed Racket

### LINKED ISSUE #427: Add the Stream Sieve benchmark
Some gradual typing papers use the following benchmark with the naive prime sieve using lazy streams. Now that we have closures in Pallene, it might be interesting to add it to the benchmark list

I don't remember why the `stream_take` function is there. Maybe I added it for debugging?

```lua
-- Simple streams library for building infinite lists.
-- A stream is a cons of a value and a thunk that computes the next value
-- when applied.

-- (any, any->any) -> any
local function make_stream(v, f)
    return {v=v, f=f}
end

local function stream_head(st)
    return st.v
end

local function stream_tail(st)
    return st.f(st.v)
end

local function stream_get(st, n)
    for _ = 1, n-1 do
        st = stream_tail(st)
    end
    return stream_head(st)
end

local function stream_take(st, n)
    local elems = {}
    for i = 1, n do
        elems[i] = stream_head(st)
        st = stream_tail(st)
    end
    return elems
end

----------------------------------------------

-- Build a stream of integers starting from 1
local function count_from(n)
    return make_stream(n, function(i) return count_from(i+1) end)
end

-- Filter all multiples of n
local function sift(n, st)
    local hd = stream_head(st)
    local tl = stream_tail(st)
    if hd % n == 0 then
        return sift(n, tl)
    else
        return make_stream(hd, function() return sift(n, tl) end)
    end
end

-- Naive sieve of Erasthostenes
local function sieve(st)
    local hd = stream_head(st)
    local tl = stream_tail(st)
    return make_stream(hd, function() return sieve(sift(hd, tl)) end)
end

--
local prime_stream = sieve(count_from(2))

local function get_prime(n)
    return stream_get(prime_stream, n)
end

return {
    get_prime = get_prime
}
```

hugomg:
Closes #427.

This is a nice benchmark from the gradual typing literature but it has a problem... it hits the LUAI_MAXCCALLS limit due to too much recursion. I think I'll leave it here as a draft pull-request for now.

### DIFF
--- a/benchmarks/streamSieve/lua.lua
+++ b/benchmarks/streamSieve/lua.lua
@@ -1,0 +1,67 @@
+-- Naive stream-based prime sieve benchmark. (Based on a typed-racket benchmark)
+-- Based on https://github.com/bennn/gtp-benchmarks/tree/master/benchmarks/sieve
+
+local m = {}
+
+----------------------------------------------
+
+-- Simple streams library for building infinite lists.
+-- A stream is a cons of a value and a thunk that computes the next value.
+
+
+
+-- TODO: with recursive types, this could be (any -> Stream)
+
+
+local function make_stream(first, rest)
+    return {first=first, rest=rest}
+end
+
+local function stream_head(st)
+    return st.first
+end
+
+local function stream_tail(st)
+    return st.rest()
+end
+
+local function stream_get(st, n)
+    for _ = 1, n-1 do
+        st = stream_tail(st)
+    end
+    return stream_head(st)
+end
+
+----------------------------------------------
+
+-- Build a stream of integers starting from 1
+local function count_from(n)
+    return make_stream(n, function() return count_from(n+1) end)
+end
+
+-- Filter all multiples of n
+local function sift(n, st)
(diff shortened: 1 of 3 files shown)


hugomg:
There is a chance that this might stop blowing the stack if we implement tail recursion optimization (#221). It might also benefit from proper recursive types (#85), because there are a couple of places where the `as` cast makes the function no longer tail recursive.

hugomg:
Another pre-requesite is optimizing indirect Pallene->Pallene calls (#354). Tail-call optimization is not possible for functions that cross a Pallene<->Lua boundary.

hugomg:
Scratch that... It turns out that the benchmark isn't actually tail recursive. There are some function calls that are, but there are some that are not. In particular, the part in `sift` that happens before the `% n == 0` check.

srijan-paul:
Does this run in regular lua?

hugomg:
It works fine in Lua because the call stack limit is much higher (millions of calls, instead of hundreds of calls)

PR #467, opened 2021-08-18
Fixed `ir.Cmd.SetUpvalue` instruction.

srijan-paul:
This fixes the code codegen for `ir.Cmd.SetUpvalue` where it would set *all* upvalues of a closure instead of the ones it is supposed to set. 

Sample codegen for a closure with 3 integer upvalues.
Before:

```c
setivalue(&(clCvalue(&x4)->upvalue[1]), x1);
setivalue(&(clCvalue(&x4)->upvalue[2]), x2);
setivalue(&(clCvalue(&x4)->upvalue[3]), x3);
setobj(L, s2v(base + 0), &x4);
setivalue(&(clCvalue(&x4)->upvalue[1]), x1);
setivalue(&(clCvalue(&x4)->upvalue[2]), x2);
setivalue(&(clCvalue(&x4)->upvalue[3]), x3);
setobj(L, s2v(base + 0), &x4);
setivalue(&(clCvalue(&x4)->upvalue[1]), x1);
setivalue(&(clCvalue(&x4)->upvalue[2]), x2);
setivalue(&(clCvalue(&x4)->upvalue[3]), x3);
setobj(L, s2v(base + 0), &x4);
x5 = function_03(L, base + 1, K, clCvalue(&x4)->upvalue);
base = restorestack(L, base_offset);
```

after:

```c
setivalue(&(clCvalue(&x4)->upvalue[1]), x1);
setivalue(&(clCvalue(&x4)->upvalue[2]), x2);
setivalue(&(clCvalue(&x4)->upvalue[3]), x3);
setobj(L, s2v(base + 0), &x4);
x5 = function_03(L, base + 1, K, clCvalue(&x4)->upvalue);
base = restorestack(L, base_offset);
```

In addition, the ir.Cmd.Upvalue itself has had some changes.
The closure argument is now called `src_f`, and the upvalue index it's supposed to initialize is called `u_id`. (It cannot be called `dst` because `gc.lua` assumes every `dst` is a local var)

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1326,7 +1326,7 @@ gen_cmd["NewClosure"] = function (self, cmd, _func)
     })
 end
 
-gen_cmd["SetUpvalue"] = function(self, cmd, _func)
+gen_cmd["SetUpvalues"] = function(self, cmd, _func)
     local func = self.module.functions[cmd.f_id]
 
     assert(cmd.src_f._tag == "ir.Value.LocalVar")
@@ -1337,6 +1337,10 @@ gen_cmd["SetUpvalue"] = function(self, cmd, _func)
         local typ   = func.captured_vars[i].typ
         local c_val = self:c_value(val)
         local upvalue_dst = string.format("&(ccl->upvalue[%s])", C.integer(i))
+
+        -- Even though the CClosure is a heap object, it is safe to use `set_stack_slot` as
+        -- there are no operations in between the closure's creation and the upvalue initialization
+        -- that may trigger a GC Cycle.
         table.insert(capture_upvalues, set_stack_slot(typ, upvalue_dst, c_val))
     end
 
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -313,7 +313,7 @@ end
 -- @returns the C parameter name for the upvalue u_id
 function Coder:c_upval(u_id)
     assert(self.current_func)
-    local typ = self.current_func.captured_vars[u_id].decl.typ
+    local typ = self.current_func.captured_vars[u_id].typ
     -- Since upvalue boxes do not have metatables, type checking them at runtime is not possible.
     -- Moreover, since upvalues are only passed around internally by Pallene, it is ok to assume that
     -- their types will be correct. So we directly cast it using `lua_value` without a tag check.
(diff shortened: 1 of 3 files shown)


srijan-paul:
But this makes me wonder, instead of having the `SetUpvalue` set individual upvalues, it could process all upvalues for a closure in a batch. Since that seems to work even with the tests we have now for mutual and self recursion. The IR will be smaller that way too.

That said, this scheme of individual upvalue setting has a more direct correspondence between the IR and generated C code which I like.

hugomg:
I like the idea of setting all the upvalues in a batch. It would let us add a compile-time `assert()` that the number of upvalues is correct and that we are not skipping any.

The IR will never want to move or reorder the setupvalue instructions. Therefore, they don't necessarily have to be at the granularity of individual upvalues.

It would also resolve the issue with the repeated clcvalue, even if we compile with -O0.

srijan-paul:
> It would also resolve the issue with the repeated clcvalue, even if we compile with -O0.

This should be possible now.

pallene/coder.lua
```diff
@@ -1328,17 +1328,29 @@ end
...
-        local upvalue_dst = string.format("&(%s->upvalue[%s])", cclosure, C.integer(i))
+    assert(cmd.src_f._tag == "ir.Value.LocalVar")
+    local cclosure = string.format("clCvalue(&%s)", self:c_var(cmd.src_f.id))
+
+    local capture_upvalues = {}
+    for i, val in ipairs(cmd.srcs) do
+        local typ   = func.captured_vars[i].typ
+        local c_val = self:c_value(val)
+        local upvalue_dst = string.format("&(ccl->upvalue[%s])", C.integer(i))
         table.insert(capture_upvalues, set_stack_slot(typ, upvalue_dst, c_val))
```
hugomg, line 1340: I just noticed another thing: we're using `set_stack_slot` here even though it is actually a heap slot. We should have a comment explaining why it is OK.

The reason is that there are no GC calls between the `NewClosure` and the `SetUpvalue`. The newly created closure is guaranteed to be white and therefore the write barrier is not necessary.

pallene/ir.lua
```diff
@@ -166,8 +159,8 @@ local ir_cmd_constructors = {
     SetField   = {"loc", "rec_typ",        "src_rec", "field_name", "src_v"},
 
     -- Functions
-    NewClosure = {"loc", "dst", "f_id"},
-    SetUpvalue = {"loc", "dst", "src", "f_id"},
+    NewClosure = {"loc", "dst"  , "f_id"},
+    SetUpvalue = {"loc", "src_f", "srcs", "f_id"},
```
hugomg, line 163: Maybe it should call it SetUpvalues (plural).

hugomg (changes requested):
Nice! I think it is much cleaner than it was before.

hugomg (approved):
🚀

PR #469, opened 2021-08-20
Continue parsing after some syntax errors

hugomg:
If it is obvious how to recover from a syntax error, then continue parsing.
For example, consider the following program:

    function f()
        break
    end

Previously, it would stop and complain at the start, when it failed to parse the module declaration.
Now, it continues to parse and reports several errors.

    syntax error: bar.pln:1:1: must begin with a module declaration; local <modname> = {}
    syntax error: bar.pln:2:5: break statement outside of a loop
    syntax error: bar.pln:1:1: function 'f' was not forward declared
    syntax error: bar.pln:4:1: must end by returning the module table

At the moment, the parser only continues if it is obvious how to proceed. For more complicated syntax errors, we abort immediately because we would rather avoid producing spurious errors. We also show an message saying that the parsing was aborted, to say that there may also be other errors in the part of the program that we did not look at.

    syntax error: bar.pln:2:20: unexpected '%' while trying to parse a statement
    compilation aborted due to previous error

One thing that I didn't try to implement is to sort the errors so that they appear in the order of their position in the file. Is this something that we should try to fix?

### DIFF
--- a/pallene/driver.lua
+++ b/pallene/driver.lua
@@ -55,36 +55,38 @@ driver.list_of_compiler_passes = {"lexer", "ast", "checker", "assignment_convers
 function driver.compile_internal(filename, input, stop_after, _opt_level)
     stop_after = stop_after or "optimize"
 
-    local lexer = Lexer.new(filename, input)
-    if stop_after == "lexer" then
-        return lexer, {}
+    local errs
+
+    local function abort()
+        if type(errs) == "string" then errs = { errs } end
+        table.insert(errs, "compilation aborted due to previous error")
+        return false, errs
     end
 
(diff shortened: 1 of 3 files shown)


hugomg:
This PR doesn't do some of the fancier things that I have been thinking about recently. It doesn't try to recover from unexpected tokens, and it doesn't use indentation as a heuristic to improve the message for "missing end". For now, I'm not sure if I actually want to do those things in Pallene... But in any case, I suppose that it doesn't hurt to pick the low hanging fruit and let the parser recover from some errors when it's easy to do so.

srijan-paul (approved):

PR #471, opened 2021-08-21
Fix a bug with "break"

### LINKED ISSUE #470: Nested functions can have break statement outside a loop
The following program passes the parser and typechecker. The compiler crashes with a failed assertion in the uninitialized.lua

```lua
local m = {}
local function outer()
    while true do
        local function inner()
            break
        end
    end
end
return m

```

```
lua: ./pallene/uninitialized.lua:161: ./pallene/uninitialized.lua:71: assertion failed!
stack traceback:
```

The problem is that when the parser checks that break statements should be inside a loop, it should take nested functions into account.

hugomg:
Previously, if a lambda was inside a loop then the body of the lambda was allowed to have break statements.  (That would crash the compiler)

Fixes #470.

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -551,7 +551,7 @@ function Parser:FuncStat(is_local)
         self:region_end()
     end
 
-    local block = self:Block()
+    local block = self:FuncBody()
     local _     = self:e("end", start)
 
     for _, decl in ipairs(params) do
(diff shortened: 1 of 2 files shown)


srijan-paul:
Wow, this one was sneaky.

srijan-paul (approved):

PR #472, opened 2021-08-22
Improved error message when an "end" is missing

hugomg:
If an "end" token is missing, we now use an indentation-based heuristic to try to spot the actual culprit for the error. In addition to showing the location where the error was detected, we also show the location of the block opener that we think is to blame. For example,

```lua
local m = {}
local function foo()
    if true then
end
return m
```

Now shows this error:

    syntax error: foo.pln:6:1: expected 'end' before end of the file, to close the 'function' at line 2
    syntax error: foo.pln:3:5: ...possibly because this 'if' is missing an 'end' (mismatched indentation)
    compilation aborted due to previous error

We give a similar error message for mismatched parenthesis and curly braces, provided that the open and close brace are on different lines.

-----------------

🙋 I tried to make the error message fit in a single line, but I'm not sure if this wording is ideal... Any suggestions?

### DIFF
--- a/pallene/parser.lua
+++ b/pallene/parser.lua
@@ -40,10 +40,10 @@ function Parser:init(lexer)
     self.mismatched_indentation = {} -- list of tokens (2i+1 = open, 2i+2 = close)
     setmetatable(self.indent_of_token, { __mode = "k" })
 
-    self:_advance(); self:_advance()
+    self:advance(); self:advance()
 end
 
-function Parser:_pay_attention_to_suspicious_indentation(open_tok, close_tok)
+function Parser:pay_attention_to_suspicious_indentation(open_tok, close_tok)
     local d1 = assert(self.indent_of_token[open_tok])
     local d2 = assert(self.indent_of_token[close_tok])
     if d1 > d2 then
@@ -52,7 +52,7 @@ function Parser:_pay_attention_to_suspicious_indentation(open_tok, close_tok)
     end
 end
 
-function Parser:_advance()
+function Parser:advance()
     local tok, err
     repeat
         tok, err = self.lexer:next()
@@ -95,11 +95,10 @@ end
 -- If the name is not provided, match whatever token we just peek-ed.
 -- If the optional open_tok is provided then we are matching a closing token ({}, (), do end, etc).
 function Parser:e(name, open_tok)
-    name = name or self.next.name
     local tok = self:try(name)
     if tok then
         if open_tok then
-            self:_pay_attention_to_suspicious_indentation(open_tok, tok)
+            self:pay_attention_to_suspicious_indentation(open_tok, tok)
         end
         return tok
     else
@@ -112,7 +111,7 @@ function Parser:try(name)
     assert(name)
     assert(name ~= "EOF")
     if self:peek(name) then
-        return self:_advance()
+        return self:advance()
     else
         return false
     end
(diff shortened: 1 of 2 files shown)


srijan-paul:
Should we add tests for this in `parser_spec`?

hugomg:
Are you suggesting that the error message should point to the suspicious closing token, instead of pointing to the suspicious opening token?

I think Luacheck does that. Maybe we should copy what they do?

    stdin:3:1: expected 'end' (to close 'while' on line 2) near less indented 'end' (indentation-based guess)

hugomg:
After the last commit, the error message is more similar to the one that Luacheck uses. It points to the location of the missing 'end', and we also no longer show the original error (which would typically be an expected 'end' before end of file, to close the function).

    syntax error: foo.pln:4:1: expected 'end' to close 'if' at line 3, before this less indented 'end'
    compilation aborted due to previous error

One thing I did slightly different from Luacheck is that I don't explicitly say that it is an indentation-based guess. I think the current message might already suggest that?

pallene/parser.lua
```diff
@@ -24,13 +24,34 @@ function Parser:init(lexer)
...
+    self.curr_line   = 0
+    self.curr_indent = 0
+    self.indent_of_token = {}    -- { token => integer }
+    self.mismatched_indentation = {} -- list of tokens (2i+1 = open, 2i+2 = close)
+    setmetatable(self.indent_of_token, { __mode = "k" })
+
     self:_advance(); self:_advance()
 end
 
+function Parser:_pay_attention_to_suspicious_indentation(open_tok, close_tok)
```
srijan-paul, line 46: Why is this (and `_advance`) prefixed with an underscore?

pallene/parser.lua
```diff
@@ -1055,16 +1086,43 @@ function Parser:unexpected_token_error(non_terminal)
...
     if not open_tok or loc.line == open_tok.loc.line then
         self:syntax_error(loc, "expected %s before %s", what, where)
     else
         local owhat = self:describe_token_name(open_tok.name)
-        self:syntax_error(loc, "expected %s before %s, to close the %s at line %d",
-            what, where, owhat, open_tok.loc.line)
+        local oline = open_tok.loc.line
+        if is_stolen_delimiter then
+            self:syntax_error(loc,
+                "expected %s to close %s at line %d, before this less indented %s",
```
srijan-paul, line 1117: Is it easy to tell what "**this** less indented token" refers to when reading the error?
Ideally, we'd be highlighting the exact places on the terminal, but Pallene does not need that atm so I guess this the best we can do.

srijan-paul (approved):

srijan-paul:
And I was also wondering, would it be good to have tests for mismatched open-close pairs across multiple levels of indentation? Currently, we're only testing if it works for an immediately missing end token.

hugomg:
> And I was also wondering, would it be good to have tests for mismatched open-close pairs across multiple levels of indentation? Currently, we're only testing if it works for an immediately missing end token.

Sorry, I don't understand what you mean here. Could you please add an example?

srijan-paul:
> Sorry, I don't understand what you mean here. Could you please add an example?

Oh I meant:

```lua
function foo(a, b)
   if a > b then
       while a > b do
            a = a - b
       -- missing end
  end   
end
```

Or:

```lua
function foo(a, b)
   if a > b then
       while a > b do
            a = a - b
       end
   -- missing end
end
```

hugomg:
I think these two might already be working correctly? In the first example, it detects the error when it gets to the end of the file but then the heuristic correctly identifies that the 'while' is the culprit.

    syntax error: foo.pln:7:4: expected 'end' to close 'while' at line 4, before this less indented 'end'

PR #473, opened 2021-08-23
Toplevel statements can now end in semi-colons.

### LINKED ISSUE #446: Toplevel statement ending with ";" is currently forbidden
Is it intentional behavior for toplevel statements to not end with semi colons?

```
local m = {};
return m
```
```
test.pln:1:13: unexpected ';' while trying to parse a statement
```

I ask because it's also forbidden in the grammar in [the manual](https://github.com/pallene-lang/pallene/blob/master/doc/manual.md)

```
program ::= {toplevelrecord | topleveltypealias | toplevelvar | toplevelfunc}
```

srijan-paul:
Small patch to fix #446.
I think it's okay to allow repeated semi-colons (`;;;;`), since Lua does so too?

The  grammar for `statement` suggests that as well.

I'm not too sure where to add the tests for this though. 
The tests for the frontend generally exist to assert certain errors and I'm not sure if having something like this be tested in it's own `it` block in `execution_tests` is a good idea.

### DIFF
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -409,6 +409,13 @@ describe("Parser /", function()
             })
         end)
 
+        it("cannot be followed by a multiple semicolons", function()
+            assert_program_error([[
+                local m: module = {}
+                return m;;
+            ]], "the module return statement must be the last thing in the file")
+        end)
+
         it("must be the last statement in the block", function()
             assert_statements_error([[
                 return 10
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -227,6 +227,17 @@ describe("Parser /", function()
                 return i
             ]], "must return exactly the module variable 'm'")
         end)
+
+        it("can have semi-colons at the toplevel", function ()
+            assert_parses_successfuly([[
+                local m: module = {};
+                record Number
+                    num: integer;
+                end;
+                return m;
+            ]])
+        end)
+
     end)
 
     --
@@ -396,12 +407,6 @@ describe("Parser /", function()
             assert_statements_ast("return 10;", {
                 { _tag = "ast.Stat.Return", exps = { { _tag = "ast.Exp.Integer", value = 10 } } }
             })
-        end)
-
-        it("cannot be followed by a multiple semicolons", function()
-            assert_statements_error([[
-                return;;
-            ]], "expected 'end' before ';', to close the 'function' at line 2")
         end)
 
         it("must be the last statement in the block", function()
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -513,7 +513,7 @@ end
 Here is the complete syntax of Pallene in extended BNF.
 As usual, {A} means 0 or more As, and \[A\] means an optional A.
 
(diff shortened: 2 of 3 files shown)


hugomg:
Maybe a cleaner way to do it would be to use the same logic that Lua does. I think we got bit by this bug because we did things differently in Pallene.

Instead of teaching the toplevel about semicolons, we could tell the parser that the `;` is an empty statement. If we add a case for the `;` in the `Parser:Stat`, that should also automatically work for the toplevel. Currently, we handle the semicolons in the `StatList`. But this bug with the toplevel suggests that it might be better to deal with the `;` in the `Stat` instead.

As for the return statement, we could tell the parser to skip the optional semicolon. Currently it's using `peek` but it should be using `try` instead. (By the way, this optional semicolon after the return is a bit weird, but Lua allows it...)
```lua
    elseif self:peek("return") then
        local start = self:e()
        if self:peek(";") or self:block_follow() then
```

>  The tests for the frontend generally exist to assert certain errors and I'm not sure if having something like this be tested in it's own it block in execution_tests is a good idea.

This was tricky enough to become a bug, so it deserves some tests 🙂️. There are two options:

1. Test it in the parser_spec, by looking at the generated AST
2. Test it in the execution_tests, by seeing if the parser accepts it without errors

AST tests are more fragile, because they break when we change the AST. However, they can be useful for testing these corner cases of parsing, because the best way to be sure that the parser is behaving exactly as intended is by looking at the AST. (For example, it's more direct to test operator precedence by looking at the AST).

My suggestion for this case would be to put them in the parser_spec. There are already some test cases for semicolons after return statements;  we could add some for semicolons in the toplevel.

hugomg:
Alternatively, perhaps we could have the Toplevel call StatList. To summarize:

* We should try to handle toplevel semicolons using the function that we use for normal statements
* If we use `try(";")` when parsing the return statement, that should also handle the semicolon in the module return statement.

srijan-paul:
Toplevel statements can now end with multiple semi-colons, including the return statement.
I decided to make a tiny helper that repeatedly eats semi-colons since we'd still need them in places where we previously had `self:try(";")` (e.g -> after a module declaration).

`Parser.Toplevel` now delegates to `Parser.StatList`, which is now capable for parsing both toplevel and non-toplevel statements.

spec/parser_spec.lua
```diff
@@ -398,12 +409,6 @@ describe("Parser /", function()
             })
         end)
 
-        it("cannot be followed by a multiple semicolons", function()
-            assert_statements_error([[
-                return;;
-            ]], "expected 'end' before ';', to close the 'function' at line 2")
-        end)
```
hugomg, line 405: We should not have deleted this test case. This is intentional, because that is what Lua does. As I said before, semicolons after the return statement are an exception...

hugomg (changes requested):

spec/parser_spec.lua
srijan-paul: My bad, I thought you meant it the other way around.

hugomg (approved):
Nice!

PR #478, opened 2021-08-29
Re-introduce constant propagation pass.

### LINKED ISSUE #444: The constant propagation pass is out of place
Because we are doing the the constant propagation after the closure conversion, the constant variables are kept as upvalues even if the constant propagation gets rid of the variable.

We have some possibilities...

1. Do this analysis in the AST, before closure conversion
2. Keep the analysis in the IR, taking care to remove the unused upvalues
3. Maybe something else I didn't think?

srijan-paul:
This PR re-introduces the constant propagation pass and fixes #444 (hopefully).

What's different:
- Upvalues and local variables are now propagated (There are no globals).
- Non-toplevel variables that are constants can be propagated as well.

For example:
```lua
local function f()
	local x = 1
	local function g(): integer
		return x + 10
	end
end
```
The code for the function `g` compiles down to:

```c    
lua_Integer x1;
    
x1 = intop(+, 1, 10);
return x1;
```

Overview:
- For every `ir.Function` in the module, we do a little bit of book keeping. Namely,
    - Read/Write count of every local var ID
    - Constant initializer (if any) of a local variable.
    - If a certain upvalue is actually a constant in disguise or not.
    - Constant value of an upvalue (which is just the constant initializer of the local variable it captures).

The overall structure is more or less the same.

First, we inspect all the `ir.Cmd.Move` instructions that have their `src` set to a constant value. This tells us which local variables have constant initializers.
Then, we inspect the read/write counts for all local variables and upvalues.

A local variable is a constant if:
 - It has a constant initializer.
 - It has a write count of 1 (assigned to only once).

An upvalue that captures a local var from it's surrounding function is a constant if:
  - If the local var is a constant

An upvalue that captures another upvalue from it's surrounding function is a constant if:
- The other upvalue that it's capturing is a constant too.

Once we have all this information, we go over the bodies of all functions, and propagate the constants.
This is done by **mutating** any `ir.Cmd` that has a `dst`/`dsts` which contain local vars/ upvalues that can be propagated.

**NOTE**: At the moment, unused variables (with write count of 0) are not elimited, and the upvalues which get constant propagated are still captured (but not used) by the `SetUpvalues` instruction. This requires a small fix. I refrained from doing this because I wasn't sure if the current way of mutating existing instructions is preferable. Moreover, I wasn't sure if we still want to keep the overall structure of the constant propagation or make it a class (like the other compiler passes).

### DIFF
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -131,17 +131,21 @@ function constant_propagation.run(module)
         for cmd in ir.iter(func.body) do
             if cmd._tag == "ir.Cmd.SetUpvalues" then
                 local next_f   = assert(data_of_func[cmd.f_id])
+                local ir_func  = module.functions[cmd.f_id]
                 local new_u_id = next_f.new_upvalue_id
 
                 local new_srcs = {}
+                local new_captured_vars = {}
                 for u_id, value in ipairs(cmd.srcs) do
                     if not next_f.is_upvalue_constant[u_id] then
                         table.insert(new_srcs, value)
                         new_u_id[u_id] = #new_srcs
+                        table.insert(new_captured_vars, ir_func.captured_vars[u_id])
                     end
                 end
 
                 cmd.srcs = new_srcs
+                ir_func.captured_vars = new_captured_vars
             end
         end
     end
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -59,23 +59,16 @@ function constant_propagation.run(module)
 
     for f_id, func in ipairs(module.functions) do
         -- DFS traversal to find the ir.Cmd.Move instructions which have constant
-        -- values as their src. We skip initializers in loops to make sure we only count
-        -- the initializers which are guaranteed to be evaluated at runtime.
-        local f_data   = assert(data_of_func[f_id])
+        -- values as their src. We can look inside initializers in loops and if-statements because
+        -- the `uninitialized.lua` pass takes care of variables that are used before being initialized.
+        local f_data = assert(data_of_func[f_id])
 
-        local stack =  { func.body }
-        while #stack > 0 do
-            local cmd = table.remove(stack)
+        for cmd in ir.iter(func.body) do
             local tag = cmd._tag
             if     tag == "ir.Cmd.Move" then
                 local id = cmd.dst
                 if is_constant_value(cmd.src) then
                     f_data.locvar_constant_init[id] = cmd.src
-                end
-
-            elseif tag == "ir.Cmd.Seq" then
-                for i = #cmd.cmds, 1, -1 do
-                    table.insert(stack, cmd.cmds[i])
                 end
 
             elseif tag == "ir.Cmd.SetUpvalues" then
@@ -92,6 +85,10 @@ function constant_propagation.run(module)
                 end
 
             end
+        end
+
+        for loc_id = 1, #func.typ.arg_types do
+            f_data.locvar_constant_init[loc_id] = false
         end
     end
 
@@ -154,10 +151,9 @@ function constant_propagation.run(module)
     -- @param f_data  FuncData of the function whose IR contains `src_val`.
     -- @param func    corresponding ir.Function
     -- @param src_val The value that we may need to update.
-    local function updated_value(f_data, func, src_val)
+    local function updated_value(f_data, src_val)
         if src_val._tag == "ir.Value.LocalVar"
            and f_data.n_writes_of_locvar[src_val.id] == 1
-           and src_val.id > #func.typ.arg_types
            and f_data.locvar_constant_init[src_val.id]
         then
             return f_data.locvar_constant_init[src_val.id]
@@ -180,13 +176,13 @@ function constant_propagation.run(module)
         func.body = ir.map_cmd(func.body, function(cmd)
             local inputs = ir.get_value_field_names(cmd)
             for _, src_field in ipairs(inputs.src) do
-                cmd[src_field] = updated_value(f_data, func, cmd[src_field])
+                cmd[src_field] = updated_value(f_data, cmd[src_field])
             end
 
             for _, src_field in ipairs(inputs.srcs) do
                 local srcs = cmd[src_field]
                 for i, value in ipairs(srcs) do
-                    srcs[i] = updated_value(f_data, func, value)
+                    srcs[i] = updated_value(f_data, value)
                 end
             end
 
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -49,7 +49,7 @@ end
 -- Does not currently recognize non-trivial constant expressions as being constant.
 function constant_propagation.run(module)
 
-    -- 1) Find what toplevel variables are initialized to a constant.
+    -- 1) Find which variables are initialized to a constant.
 
     local data_of_func = {} -- list of FuncData
     for _, func in ipairs(module.functions) do
@@ -79,13 +79,22 @@ function constant_propagation.run(module)
                         next_f.constant_val_of_upvalue[u_id] = const_init
 
                     elseif value._tag == "ir.Value.Upvalue" then
+                        -- A `NewClosure` or `SetUpvalues` instruction can only reference values in outer scopes,
+                        -- which exist in surrounding functions that have a numerically lesser `f_id`.
+                        -- Due to this, we can reliable tie the constant initializer of an inner upvalue in a nested
+                        -- function to the constantant initializer of the outer upvalue that it captures.
                         local const_init = f_data.constant_val_of_upvalue[value.id]
                         next_f.constant_val_of_upvalue[u_id] = const_init
+
+                    else
+                        typedecl.tag_error(value._tag)
                     end
                 end
 
             end
         end
+
+
 
         for loc_id = 1, #func.typ.arg_types do
             f_data.locvar_constant_init[loc_id] = false
@@ -101,13 +110,16 @@ function constant_propagation.run(module)
         for cmd in ir.iter(func.body) do
             local tag = cmd._tag
             if tag == "ir.Cmd.SetUpvalues" then
+                local next_f = assert(data_of_func[cmd.f_id])
                 for u_id, value in ipairs(cmd.srcs) do
-                    local next_f = assert(data_of_func[cmd.f_id])
-
-                    if value._tag == "ir.Value.LocalVar" and n_writes[value.id] ~= 1 then
-                        next_f.constant_val_of_upvalue[u_id] = false
+                    if value._tag == "ir.Value.LocalVar" then
+                        if n_writes[value.id] ~= 1 then
+                            next_f.constant_val_of_upvalue[u_id] = false
+                        end
                     elseif value._tag == "ir.Value.Upvalue" then
                         next_f.constant_val_of_upvalue[u_id] = f_data.constant_val_of_upvalue[value.id]
+                    else
+                        typedecl.tag_error(value._tag)
                     end
                 end
 
@@ -118,6 +130,32 @@ function constant_propagation.run(module)
                 end
             end
         end
+
+        -- Because of the way the previous compiler passes work, it is guaranteed that an upvalue that has a
+        -- constant initializer always references a local variable with a write count of 1. In other words,
+        -- IR like this is currently not possible:
+        -- ```
+        -- x1 <- 10
+        -- loop {
+        --     x2 = NewClosure()
+        --     x2.upvalues <- x1
+        --     x1 <- 20
+        -- }
+        -- ```
+        -- Since x1 is a "mutable upvalue", the assignment_conversion pass turns it into a record type.
+        -- With this loop, we assert this assumption.
+        for cmd in ir.iter(func.body) do
+            local tag = cmd._tag
+            if tag == "ir.Cmd.SetUpvalues" then
+                local next_f = assert(data_of_func[cmd.f_id])
+                for u_id, value in ipairs(cmd.srcs) do
+                    if value._tag == "ir.Value.LocalVar" and next_f.constant_val_of_upvalue[u_id] then
+                        assert(n_writes[value.id] == 1)
+                    end
+                end
+            end
+        end
+
     end
 
     -- 3) Remove propagated upvalues from the capture list.
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -151,29 +151,43 @@ function constant_propagation.run(module)
 
     -- 4) Propagate the constants local variables and upvalues.
 
+    -- Returns a new `ir.Value` representing `src_val` after constant propagation.
+    -- @param f_data  FuncData of the function whose IR contains `src_val`.
+    -- @param func    corresponding ir.Function
+    -- @param src_val The value that we may need to update.
+    local function updated_value(f_data, func, src_val)
+        if src_val._tag == "ir.Value.LocalVar"
+           and f_data.n_writes_of_locvar[src_val.id] == 1
+           and src_val.id > #func.typ.arg_types
+           and f_data.locvar_constant_init[src_val.id] then
+
+            return f_data.locvar_constant_init[src_val.id]
+
+        elseif src_val._tag == "ir.Value.Upvalue" then
+            if f_data.is_upvalue_constant[src_val.id] then
+                return f_data.constant_val_of_upvalue[src_val.id]
+            else
+                local u_id = assert(f_data.new_upvalue_id[src_val.id])
+                return  ir.Value.Upvalue(u_id)
+            end
+        end
+
+        return src_val
+    end
+
     for f_id, func in ipairs(module.functions) do
         local f_data = data_of_func[f_id]
-        local n_writes = f_data.n_writes_of_locvar
-        local new_u_id = f_data.new_upvalue_id
 
         func.body = ir.map_cmd(func.body, function(cmd)
             local inputs = ir.get_value_field_names(cmd)
             for _, src_field in ipairs(inputs.src) do
-                local val = cmd[src_field]
-                if val._tag == "ir.Value.LocalVar"
-                   and n_writes[val.id] == 1
-                   and val.id > #func.typ.arg_types
-                   and f_data.locvar_constant_init[val.id] then
+                cmd[src_field] = updated_value(f_data, func, cmd[src_field])
+            end
 
-                    cmd[src_field] = f_data.locvar_constant_init[val.id]
-
-                elseif val._tag == "ir.Value.Upvalue" then
-                    if f_data.is_upvalue_constant[val.id] then
-                        cmd[src_field] = f_data.constant_val_of_upvalue[val.id]
-                    else
-                        local u_id = assert(new_u_id[val.id])
-                        cmd[src_field] = ir.Value.Upvalue(u_id)
-                    end
+            for _, src_field in ipairs(inputs.srcs) do
+                local srcs = cmd[src_field]
+                for i, value in ipairs(srcs) do
+                    srcs[i] = updated_value(f_data, func, value)
                 end
             end
 
--- a/pallene/constant_propagation.lua
+++ b/pallene/constant_propagation.lua
@@ -29,6 +29,7 @@ local function FuncData(func)
         is_upvalue_constant     = {}, -- { upv_id => boolean  }
         locvar_constant_init    = {}, -- { loc_id => ir.Value }
         constant_val_of_upvalue = {}, -- { upv_id => ir.Value }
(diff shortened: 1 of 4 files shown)


pallene/constant_propagation.lua
```diff
@@ -16,128 +16,140 @@ local function is_constant_value(v)
...
     end
 end
 
+-- Stores the constant initializers for local variables and upvalues of a function.
+local function FuncData(func)
+    local fdata = {
+        n_writes_of_locvar      = {}, -- { loc_id => integer  }
+        is_upvalue_constant     = {}, -- { upv_id => boolean  }
+        locvar_constant_init    = {}, -- { loc_id => ir.Value }
+        constant_val_of_upvalue = {}, -- { upv_id => ir.Value }
```
hugomg, line 31: Do we need the `is_upvalue_constant`, if we have `constant_val_of_upvalue`?

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,140 @@ local function is_constant_value(v)
...
     end
 
-    do
-        -- DFS traversal to find SetGlobal instructions with a constant initializer. We ignore the
-        -- instructions inside If, Loop, and For statements, since those might be skipped.
-        local stack = { module.functions[1].body }
+
+    for f_id, func in ipairs(module.functions) do
+        -- DFS traversal to find the ir.Cmd.Move instructions which have constant
+        -- values as their src.
```
hugomg, line 63: The old commend used to explain why we are ignoring If, Loop, and For statements. If we're still doing that, I the new comment should also say it.

hugomg (changes requested):
At a high level, I like how this code looks. It's well-organized and easy to read. However, I'm not sure if it is always correct...

> Non-toplevel variables that are constants can be propagated as well.
> 
> A local variable is a constant if: It has a constant initializer. It has a write count of 1 (assigned to only once).

The old code made some strong assumptions that I'm not sure apply to regular variables. For example, we don't look at initializers inside a for-loops and while-loops, because we know that those are not allowed at the toplevel. However, for regular local variables we need to pay attention to those things...

I think we have two options. The less ambitious option is to optimize only the toplevel variables, as the old code did. That would "unbreak" the code with the least amount of work.

The more ambitious option would be to turn this optimization into a full constant-folding optimization. However, if we choose this option, I suspect that we will need to do some proper dataflow analysis. That would require a bit of planning to sort out the details, because at the moment we don't have any such optimization merged in the master branch to take inspiration from (only in some experimental branches).

I'm inclined to do the less ambitious version first and then consider doing the more ambitious version. One of the main things the less ambitious version will need to do is take care of cleaning up the captured upvalues that are no longer captured. Even if we choose to implement a full constant propagation later, we'll still be able to reuse the old upvalue cleaning code. The effort of implementing the less ambitious version won't be wasted.

> At the moment, unused variables (with write count of 0) are not elimited

This would be a nice-to-have, because it will come in handy later. However, I don't consider this  an immediate necessity, because the current version doesn't delete the unused globals either.

That said, if we are going to remove unused upvalues, perhaps removing unused locals at the same time wouldn't be too much extra work?

> and the upvalues which get constant propagated are still captured (but not used) by the SetUpvalues instruction.

I think not removing the unused captures is more of a problem, because the affected closures may potentially be created several times during the life of the program. If possible I would like if we removed the unused captures.

> Moreover, I wasn't sure if we still want to keep the overall structure of the constant propagation or make it a class (like the other compiler passes

One of the reasons we use classes in some of the other passes is that they work nice with mutually recursive functions. When we use classes we don't have to forward-declare the functions.

If the optimization fits in a single function, then it doesn't necessarily have to be a class. Another example that does this is the gc.lua.

hugomg:
By the way, thanks for the clear explanation in the PR message. It was very helpful.

srijan-paul:
> However, for regular local variables we need to pay attention to those things...

Could you give an example? If a local variable was initialized that way, it wouldn't be constant propagated (as of this implementation), right?

```lua
local m = {}

local function f()
	local x: integer;
	if true then  x = 1 end
	local function g(): integer
		return x + 10
	end
end

return m
```
The `g`  just compiles down to:
```c
static lua_Integer function_03(
    lua_State *L,
    StackValue *base,
    Udata * restrict K,
    TValue * restrict U
) {
    
    lua_Integer x1;
    lua_Integer x2;
    
    x2 = crecord_01_prims(uvalue(&U[1]))->f01;
    x1 = intop(+, x2, 10);
    return x1;
}
```
Where `x` is captured as a regular upvalue instead of being constant propagated. It *could* be constant propagated, but that doesn't happen because we aren't performing data flow analysis. Another example:
```lua
local function f()
	local x: integer;
	local y = x + 10;
	x = 10;
end
```
Here, `x` has a constant initializer of `10` and a write count of 1, but still doesn't get constant propagated into `y`'s initializer.
```c
Udata *x1; /* x */
lua_Integer x2; /* y */
lua_Integer x3;
// ..
x3 = crecord_01_prims(x1)->f01;
x2 = intop(+, x3, 10);
```

hugomg:
That uninitialized thing certainly looks like a bug! It's behaving strangely too:

* If we remove the `if false` then the uninitialized complains
* If we use `return x` instead of the `local function g()` then the uninitialized complains (even if the `if false` is present)

Could you please create a separate issue for that?

hugomg:
> Could you give an example?

I was worried about having the wrong count of re-assignments, because we weren't counting inside loops. But I think I misunderstood.  When we count the number of assignments, we also look inside loops.

Now I have a different question... In the first step, what is the purpose of doing that DFS traversal that doesn't look inside loops? Wouldn't it be safe to look everywhere and also constant-propagate the local variables that are defined inside a loop? The situations where the optimization is invalid should be ruled out by the uninitialized.lua, right?

srijan-paul:
> Now I have a different question... In the first step, what is the purpose of doing that DFS traversal that doesn't look inside loops? 

My idea was to avoid cases like the one I mentioned above. But I agree that it might actually work if we looked inside loops as well.

> The situations where the optimization is invalid should be ruled out by the uninitialized.lua, right?

True, although currently that doesn't appear to be the case at the moment.
I think it might make sense to look inside conditionals and loops once we have that bug fixed (else we risk having UB in the tests).  What do you think?

hugomg:
That other bug can only happen when there are undefined variables. I think it's OK if we do this PR first. But you can also try to solve that other bug first if you prefer.

srijan-paul:
constant upvalues that can be propagated are no longer captured by the closures.
small example:

```lua
local function f(): () -> integer
	local x: integer = 1
	local y: integer = 2
	local function g(): integer
		y = y + 1
		return x + y + 10
	end
	return g
end
```

compiled C code:

```c
    lua_Integer x1; /* x */
    Udata *x2; /* y */
    TValue x3; /* g */

     // ...
     {
        CClosure *ccl = luaF_newCclosure(L, 3);
        ccl->f = function_03_lua;
        setuvalue(L, &ccl->upvalue[0], K);
        setclCvalue(L, &x3, ccl);
    }
    
    {
        CClosure* ccl = clCvalue(&x3); 
        setuvalue(L, &(ccl->upvalue[1]), x2); // <-- no longer captures x1 here.
    }
```

srijan-paul:
I just realized, we're still not changing the number of upvalues in the closure when it's allocated, sec.

srijan-paul:
Well, the game of life benchmark seems to be failing for some reason, this is strange because on my machine all the tests pass.

hugomg:
Strange... On my machine the tests are also passing.

srijan-paul:
> Strange... On my machine the tests are also passing.

I think I figured out the problem, let me see if my assumption is correct.

srijan-paul:
I attempted to fix what I thought was the problem (and added an explainer comment in `ir.Function`, `ir.lua`) but the tests still fail on CI (and pass on my machine) so I'm a bit confused.

hugomg:
I'll take a closer look at it later, with a fresh head.

hugomg:
I think I found a way to reproduce the bug locally. Before running the test suite, delete all the ".so" files in the benchmark directory.

I have been bitten by this issue before. The benchmark compilation tries to be clever and uses `make` to avoid recompiling if not necessary. But sometimes, it doesn't recompile when it ought to.

srijan-paul:
Aah.
I wonder if rebasing after merging #481 will fix it.

hugomg:
I don't think the bug will fix itself, unfortunately. However, we should now be able to reproduce the bug locally, and figure out what is causing it.

hugomg:
By the way, the conway benchmark is one of the ones that uses constant propagation the most. If there is a bug in the constant propagation, it makes sense that it would appear in the conway benchmark.

srijan-paul:
Oh, That makes sense. I'll try to look for the bug.

srijan-paul:
The bug has been fixed.
The problem was we were not visiting the `srcs` field `ir.Cmd.*`, only `src`s.
I'm glad the conway benchmark caught this oversight.

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,185 @@ local function is_constant_value(v)
...
+    -- @param f_data  FuncData of the function whose IR contains `src_val`.
+    -- @param func    corresponding ir.Function
+    -- @param src_val The value that we may need to update.
+    local function updated_value(f_data, func, src_val)
+        if src_val._tag == "ir.Value.LocalVar"
+           and f_data.n_writes_of_locvar[src_val.id] == 1
+           and src_val.id > #func.typ.arg_types
+           and f_data.locvar_constant_init[src_val.id] then
+
+            return f_data.locvar_constant_init[src_val.id]
```
hugomg, line 164: Style nitpick: when the condition has multiple lines, we like to put the `then` on it's own line:

```lua
if
  xxx
  and yyy
then
```

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,185 @@ local function is_constant_value(v)
...
+    -- 4) Propagate the constants local variables and upvalues.
+
+    -- Returns a new `ir.Value` representing `src_val` after constant propagation.
+    -- @param f_data  FuncData of the function whose IR contains `src_val`.
+    -- @param func    corresponding ir.Function
+    -- @param src_val The value that we may need to update.
+    local function updated_value(f_data, func, src_val)
+        if src_val._tag == "ir.Value.LocalVar"
+           and f_data.n_writes_of_locvar[src_val.id] == 1
+           and src_val.id > #func.typ.arg_types
```
hugomg, line 161: Perhaps we can avoid looking at the `src_val.id` if at the start of the analysis we say that the function arguments start having one assignment to a non-constant value.

If we think about this problem as a dataflow problem, the function arguments counts as a definition for the variable.

pallene/ir.lua
```diff
@@ -45,12 +44,17 @@ function ir.VarDecl(name, typ)
 end
 
 function ir.Function(loc, name, typ)
+    -- We need to store the number of upvalues separately (instead of just using #captured_vars) because
+    -- even though the captured_vars table is append-only, the actual number of upvalues may change
+    -- after the constant propagation pass. We cannot mutate the `captured_vars` table to reflect this change
+    -- because we still need type information of the propagated upvalues (which become constant values).
```
hugomg, line 50: This is awkward. Do you have an idea of what we would need to do to avoid this problem?

Not necessarily for this PR. I'm OK with keeping this workaround for now.

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,185 @@ local function is_constant_value(v)
...
 
-    do
-        -- DFS traversal to find SetGlobal instructions with a constant initializer. We ignore the
-        -- instructions inside If, Loop, and For statements, since those might be skipped.
-        local stack = { module.functions[1].body }
+
+    for f_id, func in ipairs(module.functions) do
+        -- DFS traversal to find the ir.Cmd.Move instructions which have constant
+        -- values as their src. We skip initializers in loops to make sure we only count
+        -- the initializers which are guaranteed to be evaluated at runtime.
```
hugomg, line 65: Actually, maybe it would be OK to also look inside the conditionals? Maybe it was a mistake to not look inside the conditionals before. I guess it should be OK provided that we do this after the uninitialized.lua?

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,185 @@ local function is_constant_value(v)
...
-            elseif tag == "ir.Cmd.SetGlobal" then
-                local id = cmd.global_id
-                n_writes[id] = n_writes[id] + 1
+            if tag == "ir.Cmd.SetUpvalues" then
+                for u_id, value in ipairs(cmd.srcs) do
+                    local next_f = assert(data_of_func[cmd.f_id])
+
+                    if value._tag == "ir.Value.LocalVar" then
+                        next_f.is_upvalue_constant[u_id] = n_writes[value.id] == 1
+                            and f_data.locvar_constant_init[value.id]
```
hugomg, line 114: This might benefit from parenthesis.

hugomg (changes requested):
I'm looking forward to merging this. Just some minor comments...

The big one, is that I suspect that maybe the right thing to do would be to also look inside the if statements?

There is also a question about if we really need the `is_upvalue_constant` or if the `constant_val_of_upvalue` is enough.

pallene/constant_propagation.lua
srijan-paul: I agree. But `uninitialized.lua` doesn't actually work the way we expect right now, right?

pallene/ir.lua
srijan-paul: Another idea I tried was to get rid of the entries in `captured_vars` that get constant propagated, thereby shortening it.
However, this way we lose not only the decl but also the type information for the upvalue, which is necessary in `coder.lua`.

So one way could to be to use a separate map for `typ_of_upvalue` , and then replace `captured_vars` with an updated version that has the correct number of upvalues in the constant propagation pass.

srijan-paul:
> There is also a question about if we really need the `is_upvalue_constant` or if the `constant_val_of_upvalue` is enough.

I feel like the way they're worded, they perform two distinct actions. The first tells us "is this upvalue a constant" and the second tells us "what is the constant initializer of this upvalue". An upvalue may not be a constant, but may still have a constant initializer. Though it's true that a single table can do the trick here, but I'm wondering what a better name for such a table would be 😅

pallene/ir.lua
hugomg: Could you give an example of how we might need that type information in the code? I though that if a variable or upvalue is constant propagated, then all it's uses should have been removed.

pallene/constant_propagation.lua
hugomg: Sorry, what do you mean here? Are you talking about that bug with captured & uninitialized values?

hugomg:
Maybe the constant_initializer is enough, without changing the name?

If n_writes is 1 and there is a constant initializer, then we know that we can constant propagate the variable.

pallene/constant_propagation.lua
srijan-paul: Yeah, that is the one. I was wondering if it can help to have a stable uinitialized variable detection pass before we rely on it.

pallene/ir.lua
srijan-paul: Turns out the reason was some of my experiments with the coder that I didn't commit xD
I git reset and readjust the captured_vars and now it works.

pallene/constant_propagation.lua
hugomg: If you want to be cautious, we can keep the constant_propagation disabled in the driver.lua, until we fix that other bug in the uninitialized.lua.

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,184 @@ local function is_constant_value(v)
...
-        -- instructions inside If, Loop, and For statements, since those might be skipped.
-        local stack = { module.functions[1].body }
+
+    for f_id, func in ipairs(module.functions) do
+        -- DFS traversal to find the ir.Cmd.Move instructions which have constant
+        -- values as their src. We skip initializers in loops to make sure we only count
+        -- the initializers which are guaranteed to be evaluated at runtime.
+        local f_data   = assert(data_of_func[f_id])
+
+        local stack =  { func.body }
```
hugomg, line 66: Now that we are no longer avoiding if statements, would it be possible to use an `ir.iter` here?

hugomg (changes requested):

pallene/constant_propagation.lua
srijan-paul: Ah, we can do that 👍

pallene/ir.lua
hugomg: Great! One less thing for us to worry about later.

hugomg (changes requested):
For this final review, I gave a close look at the entire file from top to bottom. I think I spotted a possible bug, plus two places that might make assumptions about the structure of the IR. It would be nice to assert these assumptions, because although the assumptions may be true today, it's possible that future optimization passes could break these assumptions.

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,180 @@ local function is_constant_value(v)
...
-                local id = cmd.global_id
-                n_reads[id] = n_reads[id] + 1
-            elseif tag == "ir.Cmd.SetGlobal" then
-                local id = cmd.global_id
-                n_writes[id] = n_writes[id] + 1
+            if tag == "ir.Cmd.SetUpvalues" then
+                for u_id, value in ipairs(cmd.srcs) do
+                    local next_f = assert(data_of_func[cmd.f_id])
+
+                    if value._tag == "ir.Value.LocalVar" and n_writes[value.id] ~= 1 then
```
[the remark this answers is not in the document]
srijan-paul: since x1 is being reassigned to, wouldn't that make it a mutable upvalue, and therefore not a constant (since the assignment conversion pass turns it into a record). I think our assumption was that all upvalues that are initialized to a constant in the IR are guaranteed to be `<const>` because assignment_conversion pass would box them otherwise.

pallene/constant_propagation.lua
```diff
@@ -16,128 +16,180 @@ local function is_constant_value(v)
...
+                        next_f.constant_val_of_upvalue[u_id] = false
+                    elseif value._tag == "ir.Value.Upvalue" then
+                        next_f.constant_val_of_upvalue[u_id] = f_data.constant_val_of_upvalue[value.id]
+                    end
```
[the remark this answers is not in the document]
srijan-paul: Good point, I do believe the compiler won't pass anything other than those two value types as upvalues.

pallene/constant_propagation.lua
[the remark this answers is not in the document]
hugomg: I think the problem is that we're computing the `n_writes` in the same loop that we are computing the `constant_val_of_upvalue`. When we  process the  `x2.upvalues <- x1`, we don't know the total `n_writes` for the `x1` yet.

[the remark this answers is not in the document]
srijan-paul: Could you perhaps give a Pallene code snippet as example?
I'm wondering if generating IR like above is even possible. Because if the Pallene source is something like this:

```lua
local x = 10
if x == 10 then
	local function f(): integer return x end
	x = 20
end
```

Then the `x` would become a record type, and so the n_writes wouldn't really matter since constant_val_of_upvalue would be false anyway. This makes me also realize that we are relying on assumptions across multiple modules that are not properly documented anywhere. I wonder what the appropriate place for that would be? Since `manual.md` is meant for programmers who will use Pallene and not developers who will look into it's source.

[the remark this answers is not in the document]
hugomg: You might be right, this might be another case where it is allowed by the IR but never happens in the IR that we generate.

> This makes me also realize that we are relying on assumptions across multiple modules that are not properly documented anywhere. I wonder what the appropriate place for that would be?

We don't necessarily want to codify these assumptions. While it might be true that the assumptions hold for the output of the `to_ir`, that may not necessarily be the case if we add more optimizations between the `to_ir` and the `constant_propagation`.

I think the ideal thing to do would be to make the optimization more general (accepting all representable IR programs)  or to modify the IR to make the undesirable programs impossible to represent.

A less ideal solution, which is less work,  would be to add some assertions to the optimization pass. That way, if the assumption gets broken later, at least we crash the compiler instead of producing the wrong output.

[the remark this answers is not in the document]
srijan-paul: One possibility is to move forward with this assumption by adding some `assert`s for now, since that makes a lot of things easier. We can then open an issue to make it a point that the optimizations don't work on all possible valid IR, just IR that is generated by the `to_ir` pass and verified by the `uninitialized.lua` pass.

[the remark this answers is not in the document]
hugomg: Agreed. But I wouldn't give a high priority for fixing that issue. There's a decent chance that at least some of the optimizations will have to be re-implemented in the future.

hugomg (approved):
Great!

PR #482, opened 2021-09-02
Recompile benchmarks when running test suite

hugomg:
Because the test suite did not recompile the benchmarks, sometimes the tests would pass on the local machine but not pass on the CI server. The immediate workaround is to force recompilation when running the test suite. In the future it would be nice if we replaced the benchlib by something less crazy.


hugomg:
Hopefully this workaround should do for now. I opened #483 to remember to make a proper fix later.

srijan-paul (approved):

Issue #490, opened 2021-10-05
math builtins


ewmailing:
### Discussed in https://github.com/pallene-lang/pallene/discussions/489

<div type='discussions-op-text'>

<sup>Originally posted by **ewmailing** October  4, 2021</sup>
Hi, I was trying to implement some code for my little experiment project that depend on math library functions, specifically: exp, log, pow.
I figured out how to hack these into Pallene using sqrt as the template. It's not a lot of code, but would you like me to submit a pull request?

I should note that my math.log is not fully correct because I could only figure out how to support one parameter (mapping to C's log.) But I figure at the moment, this incomplete implementation is still no worse than no implementation. For two parameters, I haven't figured out:
- How to tell the system that a second optional float parameter is supported (and how to refer to it)
- How to dynamically vary the output depending if the second parameter is 1, 2, 10, or something else (so the generated code is one of: log(x), log2(x), log10(x), or log(x)/log(base)

Similarly, math.abs also needs to handle floats and integers differently (call fabs or abs (or is it labs)). 

Thanks</div>

hugomg:
>  I needed a few more math functions, so I tried implementing a few more (ceil, floor, fmod, modf).
I got stuck on modf.

More functions would be nice :) However, I think  it would be easier to review if we put those on separate pull request(s).

> I'm getting an error that looks like:

I don't know what is the problem but usually a Lua stack trace like that points to a bug in the compiler itself. The sort of type errors we can get when programming in a dynamic language... If it is something simple like a typo then luacheck might catch it.

> Also, does the infrastructure exist to export the constants like math.pi, math.huge, etc.?

After a brief look, my guess is that we may not need to change the typechecker (checker.lua) but we may need to do some small changes to builtins.lua and to_ir.lua.

The builtins.lua has a table called `builtins.functions`, which sounds like it accepts only functions. However, I'm not sure if we actually require them to be functions. Maybe this could just as well have been called `builtins.values`? I'm not 100% sure.

In the to_ir.lua, one part that we will have to deal with is the ast.Exp.Var case in exp_to_value. Currently, returning builtins as a value is not yet implemented (#337). However, we could check if the type of the builtin is a function, or if it is a number. That way we can implement constants like `e` and `pi` while still leaving the builtin functions for later.

hugomg:
> Okay,, I figured out the first bug. It was my mistake. I missed a pair of parenthesis after T.Integer in builtins.lua.

[...]

PR #491, opened 2021-10-05
math builtins for abs, exp, low, pow

ewmailing:
This contains an implementation for math builtins for abs, exp, log, pow.

This is the issue number, which should refer to the original discussion.
https://github.com/pallene-lang/pallene/issues/490

### DIFF
--- a/pallene/coder.lua
+++ b/pallene/coder.lua
@@ -1437,6 +1437,42 @@ gen_cmd["BuiltinIoWrite"] = function(self, cmd, _func)
     return util.render([[ pallene_io_write(L, $v); ]], { v = v })
 end
 
+gen_cmd["BuiltinMathAbs"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    return util.render([[ $dst = l_mathop(fabs)($v); ]], { dst = dst, v = v })
+end
+
+gen_cmd["BuiltinMathExp"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    return util.render([[ $dst = l_mathop(exp)($v); ]], { dst = dst, v = v })
+end
+
+gen_cmd["BuiltinMathLog"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    local b = self:c_value(cmd.srcs[2])
+	local base_num = tonumber(b)
+    if base_num == 1.0 then
+        return util.render([[ $dst = l_mathop(log)($v); ]], { dst = dst, v = v })
+	elseif base_num == 2.0 then
+        return util.render([[ $dst = l_mathop(log2)($v); ]], { dst = dst, v = v })
+	elseif base_num == 10.0 then
+        return util.render([[ $dst = l_mathop(log10)($v); ]], { dst = dst, v = v })
+    else
+        return util.render([[ $dst = l_mathop(log)($v)/l_mathop(log)($b); ]], 
+            { dst = dst, v = v, b = b })
+    end    
+end
+
+gen_cmd["BuiltinMathPow"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local x = self:c_value(cmd.srcs[1])
+    local y = self:c_value(cmd.srcs[2])
+    return util.render([[ $dst = l_mathop(pow)($x, $y); ]], { dst = dst, x = x, y = y })
+end
+
 gen_cmd["BuiltinMathSqrt"] = function(self, cmd, _func)
     local dst = self:c_var(cmd.dsts[1])
     local v = self:c_value(cmd.srcs[1])
--- a/pallene/builtins.lua
+++ b/pallene/builtins.lua
@@ -23,6 +23,10 @@ builtins.modules = {
         write = T.Function({ T.String() }, {}),
     },
     math = {
+        abs  = T.Function({ T.Float() }, { T.Float() }),
+        exp  = T.Function({ T.Float() }, { T.Float() }),
+        log  = T.Function({ T.Float(), T.Float() }, { T.Float() }),
+        pow  = T.Function({ T.Float(), T.Float() }, { T.Float() }),
         sqrt = T.Function({ T.Float() }, { T.Float() }),
     },
     string = {
--- a/pallene/ir.lua
+++ b/pallene/ir.lua
@@ -167,6 +167,10 @@ local ir_cmd_constructors = {
 
     -- Builtin operations
     BuiltinIoWrite    = {"loc",         "srcs"},
+    BuiltinMathAbs    = {"loc", "dsts", "srcs"},
+    BuiltinMathExp    = {"loc", "dsts", "srcs"},
+    BuiltinMathLog    = {"loc", "dsts", "srcs"},
+    BuiltinMathPow    = {"loc", "dsts", "srcs"},
     BuiltinMathSqrt   = {"loc", "dsts", "srcs"},
     BuiltinStringChar = {"loc", "dsts", "srcs"},
     BuiltinStringSub  = {"loc", "dsts", "srcs"},
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -1030,6 +1030,18 @@ function ToIR:exp_to_assignment(cmds, dst, exp)
             if     bname == "io.write" then
                 assert(#xs == 1)
                 table.insert(cmds, ir.Cmd.BuiltinIoWrite(loc, xs))
+            elseif bname == "math.abs" then
+                assert(#xs == 1)
+                table.insert(cmds, ir.Cmd.BuiltinMathAbs(loc, dsts, xs))
+            elseif bname == "math.exp" then
+                assert(#xs == 1)
+                table.insert(cmds, ir.Cmd.BuiltinMathExp(loc, dsts, xs))
+            elseif bname == "math.log" then
+                assert(#xs == 2)
+                table.insert(cmds, ir.Cmd.BuiltinMathLog(loc, dsts, xs))
+            elseif bname == "math.pow" then
+                assert(#xs == 2)
+                table.insert(cmds, ir.Cmd.BuiltinMathPow(loc, dsts, xs))
             elseif bname == "math.sqrt" then
                 assert(#xs == 1)
                 table.insert(cmds, ir.Cmd.BuiltinMathSqrt(loc, dsts, xs))


pallene/coder.lua
```diff
@@ -1437,6 +1437,42 @@ gen_cmd["BuiltinIoWrite"] = function(self, cmd, _func)
...
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    return util.render([[ $dst = l_mathop(exp)($v); ]], { dst = dst, v = v })
+end
+
+gen_cmd["BuiltinMathLog"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    local b = self:c_value(cmd.srcs[2])
+	local base_num = tonumber(b)
```
srijan-paul, line 1456: We can use spaces to indent this instead of tabs, like the rest of the codebase.

hugomg: Aha! I know that linter check would pay off some day. :)

ewmailing: Oops. Sorry, I missed that one. I just fixed that line. How do I amend the PR with GitHub to include this change?

hugomg: I believe that converting the C string back to a number technically works, but it feels fragile because the c_value doesn't promise that the string it produces is something that can be passed to `tonumber`. For example, a future version of `c_value` could return a string that contains a `/**/` comment.

I think a better way to do this would be to check the `._tag` field of the `cmd.srcs[2]`. If it is an `ir.Value.Float` then we know it is a floating-point constant and we can look at the `.value` to find out what value it has.

pallene/coder.lua
```diff
@@ -1437,6 +1437,42 @@ gen_cmd["BuiltinIoWrite"] = function(self, cmd, _func)
...
+    if base_num == 1.0 then
```
hugomg, line 1457: Technically, using `1.0` to stand for `e` would be a break in compatibility. As a rule of thumb we try to ensure that when a Pallene program runs without errors, it produces the same result that Lua would.

However, it might be acceptable as a temporary workaround, until we implement proper optional arguments. If we choose to go forward with this workaround, I would recommend adding a code comment explaining it, and creating a separate issue in the issue tracker, so we do not forget to fix it later.

pallene/coder.lua
srijan-paul: I believe just pushing to this branch should automatically update the PR.

hugomg:
Thank you for the PR!

The code looks mostly alright, but I see two parts that may need attention.

First, we will want to add some test cases for these new builtin functions. Look for the math.sqrt tests inside the execution tests file and add some similar-looking tests for the new math functions.

For the log function, one thing that is different between our implementation and the Lua implementation is that if the second argument is not a compile-time constant then we will always use the last case, the one with the division. Even if at runtime, it turns out that the second argument is `2.0` or `10.0`.

Maybe this is ok. After all, if the second argument is not a compile-time constant then it's more likely that it's not going to be `2.0` or `10.0`. However, if it is not ok, I think the alternative would be to do the if statement in C, probably creating a helper function in the pallene_core.c.

hugomg:
Thinking a bit more, I think using the general division case is going to be a problem for compatibility. It we implement the following using log division then it is going to produce different results for Lua and for Pallene.

    local x = 3.0
    local y = 2.0
    return math.log(3.0, 2.0)
    
I still think that it is a good idea to special case the implementation if the second argument is a compile-time constant. However, if it is not a compile-time constant then for compatibility with Lua we should still have that if-else-if inside the generated C code.

pallene/coder.lua
ewmailing: I think 1.0 is what Lua does. Below is the code from lmathlib.c Notably, the first case uses lua_isoneornil(L, 2) to trigger the log() case. So, by my understanding, if the parameter is either 1.0 (explicit parameter) or nil (omitted optional parameter), log() is used.

static int math_log (lua_State *L) {
  lua_Number x = luaL_checknumber(L, 1);
  lua_Number res;
  if (lua_isnoneornil(L, 2))
    res = l_mathop(log)(x);
  else {
    lua_Number base = luaL_checknumber(L, 2);
#if !defined(LUA_USE_C89)
    if (base == l_mathop(2.0))
      res = l_mathop(log2)(x);
    else
#endif
    if (base == l_mathop(10.0))
      res = l_mathop(log10)(x);
    else
      res = l_mathop(log)(x)/l_mathop(log)(base);
  }
  lua_pushnumber(L, res);
  return 1;
}

ewmailing:
> Maybe this is ok. After all, if the second argument is not a compile-time constant then it's not likely that it's going to be `2.0` or `10.0`. However, if it is not ok, I think the alternative would be to do the if statement in C, probably creating a helper function in the pallene_core.c.

Since the compiler rules currently demand 2 float arguments for math.log() at compile time (as our temporary workaround), this won't be a problem until this is changed. 

But I don't know how to defensibly defend against this while still being correct, since it is not defined how this optional parameter mechanism will work and interact with this section (or at least not clear to me).

pallene/coder.lua
hugomg: It's a subtle difference of only one character... Lua is using "none or nil", not "one or nil". It checks if the second argument is absent, or if it is nil. That is, `math.log(x)` or `math.log(x, nil)`. While in regular Lua these two are alsmost indistinguishable, inside the C API  there is a distinction between missing argument and nil argument.

```
> math.log(3.0, 1.0)
inf
```

hugomg:
> But I don't know how to defensibly defend against this while still being correct

Forget about the things I said before, I just remembered a simpler way...

What we normally do in these cases is put the if statements in C instead of in Lua. Look at the pallene_core.h for inspiration. For example, the `pallene_shiftL` function that implements the left shift.

Because that is a `static inline` function, the C compiler is able to inline and optimize the code when the second argument is a compile-time constant. For example, if the `y ` is a compile-time constant, then after inlining the C compiler can get rid of all the if statements and keep only  `intop(<<, x, y)`.

We could do something similar for the sqrt function. Create a static inline function in the pallene_core.h, and use this function to implement the math.sqrt builtin

```c
static inline
lua_Number pallene_sqrt(lua_Number x, lua_Number base)
{
    if (base == l_mathop(1.0)) { // temporary workaround for lack of optional args
        return l_mathop(log)(x);
#if !defined(LUA_USE_C89)
    } else if (base == l_mathop(2.0)) {
        return l_mathop(log2)(x);
#endif
    } else if (base == l_mathop(10.0)) {
        return l_mathop(log10)(x);
    } else {
        return l_mathop(log)(x)/l_mathop(log)(base);
    }
}
```

ewmailing:
> > But I don't know how to defensibly defend against this while still being correct
> 
> Forget about the things I said before, I just remembered a simpler way...
> 
> What we normally do in these cases is put the if statements in C (the generated code) instead of in Lua (the compiler). Look at the pallene_core.h for inspiration. For example, the `pallene_shiftL` function that implements the left shift.
> 
> Because that is a `static inline` function, the C compiler is able to inline and optimize the code when the second argument is a compile-time constant. For example, if the `y ` is a compile-time constant, then after inlining the C compiler can get rid of all the if statements and keep only `intop(<<, x, y)`.
> 
> We could do something similar for the sqrt function. Create a static inline function in the pallene_core.h, and use this function to implement the math.sqrt builtin
> 
> ```c
> static inline
> lua_Number pallene_sqrt(lua_Number x, lua_Number base)
> {
>     if (base == l_mathop(1.0)) { // temporary workaround for lack of optional args
>         return l_mathop(log)(x);
> #if !defined(LUA_USE_C89)
>     } else if (base == l_mathop(2.0)) {
>         return l_mathop(log2)(x);
> #endif
>     } else if (base == l_mathop(10.0)) {
>         return l_mathop(log10)(x);
>     } else {
>         return l_mathop(log)(x)/l_mathop(log)(base);
>     }
> }
> ```

Can this be used to handle the optional parameter? Perhaps in combination with using Any you mentioned somewhere else, if need be?
I now have a lot of reservations about using 1.0 in the second parameter for log() since I was wrong about Lua doing this.

hugomg:
Perhaps a less ugly workaround would be to introduce a separate function for the natural logarithm. What do you think about  `math.ln`?

ewmailing:
> Perhaps a less ugly workaround would be to introduce a separate function for the natural logarithm. What do you think about `math.ln`?

I think that is reasonable. However, I think we need to do something to protect the --emit-lua flag so if/when used, the pure Lua code will do the right thing. I haven't looked into any of this code, but I'm thinking that either the pallene compiler should either replace instances of math.ln with math.log, or the generated Lua output injects some the following piece of code somewhere in the right place. 

```
math.ln = math.log
```

hugomg:
We can see that you're already getting a good idea of how the compiler works :)

Since the math.ln is a temporary workaround, I vote for whatever is the simplest to implement. Probably adding that `math.ln = math.log` assignment at the start of the generated lua file.

ewmailing:
I just committed and pushed all the changes we discussed into my repo.

As per the discussion, this changes the log implementation to use a
static inline function in C implementation for the two parameter
variant, and introduces a new math.ln built-in. Execution tests have
been added for all the new math functions. 

But there is one remaining problem with this implementation in that it breaks the Lua translator test.

As discussed, the easy solution of inserting math.ln = math.log at the
top of the generated Lua files was implemented here. To avoid breaking
line counts, the command is injected at the very beginning of the file
without a newline. Unfortunately, the Lua translator test has correctly
identified that this injection was not in the original code. I'm
committing this so this can be reviewed by others so we can determine
what to do next.

pallene/coder.lua
```diff
@@ -1449,21 +1449,19 @@ gen_cmd["BuiltinMathExp"] = function(self, cmd, _func)
     return util.render([[ $dst = l_mathop(exp)($v); ]], { dst = dst, v = v })
 end
 
+-- We are introducing math.ln as a workaround for 1 param math.log
+gen_cmd["BuiltinMathLn"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+        return util.render([[ $dst = l_mathop(log)($v); ]], { dst = dst, v = v })
```
hugomg, line 1456: The indentation is strange here

spec/execution_tests.lua
```diff
@@ -1425,6 +1425,166 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
...
+                assert(1.0 == test.exponential_value(0.0))
+                assert(2.718 == tonumber(string.format("%.3f", 
+                    test.exponential_value(1.0))))
+            ]])
+        end)
+
+        it("works on negative numbers", function()
+            run_test([[
+                assert(0.368 == tonumber(string.format("%.3f", 
+                    test.exponential_value(-1.0))))
```
hugomg, line 1477: Maybe this fits in a single line? We use a 100-column lines instead of 80-columns. A good way to get this working is to install an editorconfig plugin, see the editorconfig file for details.

pallene/translator.lua
```diff
@@ -61,6 +61,18 @@ function Translator:erase_region(start_index, stop_index)
     self.last_index = stop_index + 1
 end
 
+-- This is a workaround to handle math.log built-in optimizations.
+-- Because Pallene currently cannot handle optional parameters,
+-- we've decided to introduce math.ln as a replacement for single param log(x).
+-- But for --emit-lua, we must do something to make the code work in pure Lua.
+-- For now, the easiest thing to do is inject math.ln = math.log at the top.
+-- A smarter routine would replace math.ln with math.log.
```
hugomg, line 69: I love what this comment is saying, but I would expect that the translation.lua might not be the first place someone would look for it. What do you think about moving it to the  coder.lua or builtins.lua?

spec/execution_tests.lua
```diff
@@ -1425,6 +1425,166 @@ function execution_tests.run(compile_file, backend, _ENV, only_compile)
         end)
     end)
 
+    -- Note: Implementation currenly only supports float and uses fabs().
+    -- Test avoids differences that may be caused by needing integers.
```
hugomg, line 1429: Maybe it would be nice if we said in the manual.md that the math builtins only accept floating point numbers. And also mention in the manual.md that the math builtins exist :)

hugomg (commented):
Looks excellent! I think this needs only very minor revisions before we can merge it.

spec/execution_tests.lua
srijan-paul: This reminds me that for every update to the upstream repo, we'll have to change docs accordingly as well.

hugomg: That certainly is a downside of keeping things in a separate repository. But while we're still experimenting with restructuring the docs, I think there are advangages to keeping it separate. Once we stabilize, we can consider if we want to move the md files back to the main repo, or if we want to continue keeping them separate.

ewmailing: I made the changes you requested and just pushed them. The elephant in the room is still the translator test mismatch failure.

hugomg: The pull-request comments might not be the best place to discuss this. This comment thread is attached to a line of code and if that line of code gets edited, then the commends under it may end up hidden. I'm going to continue the discussion back on #490.

PR #493, opened 2021-10-13
Math builtins for ceil, floor, fmod, modf

ewmailing:
These are more math library builtins, as originally discussed in https://github.com/pallene-lang/pallene/issues/490


doc/manual.md
```diff
@@ -490,6 +494,11 @@ All these functions currently require float arguments. The compiler will reject
 At this time, math.log requires two parameters, and the compiler will reject one parameter calls.
 So as a workaround, use math.ln in place of the one parameter math.log.
 
+math.modf returns two values, an integer and a float. One minor difference from Lua is that Lua 
+may return a float for the first value if some reason it cannot return an integer, such as if
+the result is NaN (which only exists in floating point). In Pallene, the first value will always be
+converted to an integer.
```
hugomg, line 500: I believe that this difference also applies to other functions that use `pushnumint` internally, including `ceil` and `floor`.

ewmailing: You are right. I missed that. I just pushed a change to make ceil and floor return integers. I just did a direct cast to LUA_INTEGER instead of making a set of new inline functions just to call lua_numbertointeger. I think that will be sufficient in these cases since we can't really fall back to float anyway if it doesn't fit.

ewmailing: One more push to include ceil/float in the manual discussion about the integer conversion.

hugomg:
I know that github suggests using a merge to update the branch on top of master, but on this repo we prefer to use rebase. It keps the history less cluttered. `git rebase -i master`.

hugomg:
Having corner cases where Pallene math functions returns slightly differently than Lua makes me a bit nervous. The rule of thumb that we like to follow is that if Pallene behaves differently, the only difference is that Pallene may raise an error where Lua would not.

Alternatively, if we want to create functions that behave differently then we could name them differently. For example, `math.floorf` could be a `float->float` function that just wraps the C floor function, without any of the integer stuff.

ewmailing:
I don't know what to do with the rebase command. When I ran it, it opened a text editor with the following. Can you tell me what to do here?
```
noop                                                                            
                                                                                
# Rebase 33d223e..33d223e onto 33d223e (1 command)                              
#                                                                               
# Commands:                                                                     
# p, pick <commit> = use commit                                                 
# r, reword <commit> = use commit, but edit the commit message                  
# e, edit <commit> = use commit, but stop for amending                          
# s, squash <commit> = use commit, but meld into previous commit                
# f, fixup <commit> = like "squash", but discard this commit's log message      
# x, exec <command> = run command (the rest of the line) using shell            
# b, break = stop here (continue rebase later with 'git rebase --continue')     
# d, drop <commit> = remove commit                                              
# l, label <label> = label current HEAD with a name                             
# t, reset <label> = reset HEAD to a label                                      
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]                    
# .       create a merge commit using the original merge commit's               
# .       message (or the oneline, if no original merge commit was              
# .       specified). Use -c <commit> to reword the commit message.             
#                                                                               
# These lines can be re-ordered; they are executed from top to bottom.          
#                                                                               
# If you remove a line here THAT COMMIT WILL BE LOST.                           
#                                                                               
# However, if you remove everything, the rebase will be aborted.  

```

hugomg:
The rebase might be doing that because you already did that merge commit. Probably will require calling the rebase with different arguments. I'll take a look into this tomorrow.

hugomg:
Ah, and one last thing before I leave for today. 

I just realized that our test suite isn't testing that the math functions behave the same as they do in Lua. We could do that by calling the Lua math functions instead of hard coding the results:

    assert(math.sqrt(4.0) == test.sqrt(4.0))
    
For extra points, we could automatically test on a large list of "tricky" inputs, like we do with arithmetic operators in the coder_test_operators. (not necessarily this PR though)

ewmailing:
> Having corner cases where Pallene math functions returns slightly differently than Lua makes me a bit nervous. The rule of thumb that we like to follow is that if Pallene behaves differently, the only difference is that Pallene may raise an error where Lua would not.

So when I look at the Lua reference manual, it seems ambiguous about whether it returns float or int for these functions, as well as omits the details about its NaN handling or other exceptional cases. My understanding of how the Lua authors write the manual/spec is that the details and omissions are deliberate to allow other potential Lua implementations to be created that conform to the reference they documented, instead of requiring developers to conform to the PUC-Rio implementation behaviors. One benefit is this allows alternative implementations flexibility and optimization avenues at their disposal.

So for these 3 cases where returning a float is only for exceptional circumstances in the implementation and not the reference manual/spec, I don't feel like we are violating anything by always returning an integer.

ewmailing:
> Ah, and one last thing before I leave for today.
> 
> I just realized that our test suite isn't testing that the math functions behave the same as they do in Lua. We could do that by calling the Lua math functions instead of hard coding the results:
> 
> ```
> assert(math.sqrt(4.0) == test.sqrt(4.0))
> ```
> 
> For extra points, we could automatically test on a large list of "tricky" inputs, like we do with arithmetic operators in the coder_test_operators. (not necessarily this PR though)

Coincidentally, for the mathconstants branch I made (which I intend to submit a PR for after this one is wrapped up), the test functions directly test equality against Lua's.

pallene/coder.lua
```diff
@@ -1443,6 +1443,25 @@ gen_cmd["BuiltinMathAbs"] = function(self, cmd, _func)
     return util.render([[ $dst = l_mathop(fabs)($v); ]], { dst = dst, v = v })
 end
 
+gen_cmd["BuiltinMathCeil"] = function(self, cmd, _func)
+    local dst = self:c_var(cmd.dsts[1])
+    local v = self:c_value(cmd.srcs[1])
+    return util.render([[ $dst = (LUA_INTEGER)l_mathop(ceil)($v); ]], { dst = dst, v = v })
```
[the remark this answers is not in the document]
ewmailing: > I believe that casting an float to an integer is undefined behavior if the float is outside of the integer range. We might want to avoid that, because the C compiler can do weird things if it identifies undefined behavior. For example, that could happen if the argument to math.ceil is a suitable floating-point constant...

So for the modf implementation, I created an inline function which ultimately goes through the lua_numbertointeger macro. The side-effect of that is that since I initialized the temporary return value to 0, if lua_numbertointeger cannot convert, the value returned to the user is 0. This at least side-steps undefined behavior in that we can dictate we will return 0 (or whatever integer value we want) for this case. I can make that change to ceil & floor if you wish. 

> One way to avoid this would be to copy what Lua does to the letter (except by possibly raising an error if there is overflow). Alternatively, we could create math functions with diffferent names, without trying to convert to integer.

As I mentioned in my other response, I feel like the Lua reference gives us leeway here, as they do not explicitly dictate what we should return here. I personally don't think it is worth creating different variants of this function right now, just to handle these exceptional cases, especially when the spec doesn't even suggest it.

Raising an error is an interesting idea, but I feel it might be too aggressive since Lua just lets it slide through. I think raising an error here when Lua doesn't may just be obnoxious.

ewmailing:
Yes, I don't really have any experience/understanding with rebasing. Most of the projects I've participated in used Mercurial or didn't do rebasing in Git.

So I was working in my master branch, so I think I've broken a bunch of things for myself. I managed to rebase my local copy against the official repo master I think. However, I do not think I can push it back to my Github since they now conflict. So I created a new branch called mathbuiltins and pushed it into my Github so you can see it.
https://github.com/ewmailing/pallene/tree/mathbuiltins

I'm not sure what to do next. If I did the rebase correctly, I presume this is where the PR request needs to come from now, although I don't know what the process to do that is. Additionally, I also presume my master branch on Github is now dead in the water as it is unmergable with anything now. Is there a way to reset this one on Github?

Finally, we also still need to decide on a course of action for what the behavior of the exceptional cases for ceil, floor, modf. Based on your emphasis for safety from undefined behavior, I now feel like throwing a Lua error (or whatever your mechanism is...I haven't figured out what Pallene's error system is) is the best choice here, alternatively followed closely by simply returning 0 if the float->int conversion can't be done.

hugomg:
>  I now feel like throwing a Lua error (or whatever your mechanism is... I haven't figured out what Pallene's error system is)

Right now, we don't expose `error` and `pcall` to Pallenem so you're right that it is not clear what Pallene's error system is. That said, some internal Pallene subroutines do raise Lua errors. For example, whenever Pallene performs a downcast, if the type tag doesn't match then it raises an error. You can find some examples of this in the pallene_core.c. Look for the places that call luaL_error.

In any case, I have opened a separate issue for `error` and `pcall`: #496. Thanks for reminding me of that.

ewmailing:
Okay, I started from a fresh clone and manually copied all the old files into the new repo. I pushed a new branch called mathbuiltins2 into my Github repo and made a new PR here:
https://github.com/pallene-lang/pallene/pull/497

PR #497, opened 2021-10-19
Math builtins for ceil, floor, fmod, modf.

ewmailing:
Discussion can be found at
https://github.com/pallene-lang/pallene/pull/493


vm/src/pallene_core.h
```diff
@@ -215,6 +219,34 @@ lua_Integer pallene_shiftR(lua_Integer x, lua_Integer y)
     }
 }
 
+static inline
+lua_Number pallene_ceil(lua_State *L, const char* file, int line, lua_Number n)
```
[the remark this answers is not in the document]
ewmailing: Okay, done.

vm/src/pallene_core.h
```diff
@@ -232,6 +264,16 @@ lua_Number pallene_math_log(lua_Integer x, lua_Integer base)
     }
 }
 
+static inline
+lua_Number pallene_math_modf(lua_State *L, const char* file, int line, lua_Number n, lua_Integer* out)
```
hugomg, line 268: Should the order of the return values be (integer, float) instead of (float, integer)?

hugomg: Ah, now I see. In the coder.lua the return value is assigned to the dst2 so in the end the returned values are in the correct order.

Nevertheless, I think it might be less confusing if we use the same calling convention that se use for multiple-return Pallene functions. Return the first result (the integer) as a normal return value and return the second result (the float) via an out parameter.

ewmailing: Okay, I pushed the change. Btw, I wasn't aware there were other multiple return functions implemented. I didn't see any in coder.lua, so I made everything up as I went along.

hugomg (approved):
Great! Thanks, again!

PR #498, opened 2021-10-20
Math builtins for constants: huge, mininteger, maxinteger, pi

ewmailing:
Original discussion is at:
https://github.com/pallene-lang/pallene/issues/490

### DIFF
--- a/pallene/to_ir.lua
+++ b/pallene/to_ir.lua
@@ -863,25 +863,13 @@ function ToIR:exp_to_value(cmds, exp, _recursive)
             elseif def._tag == "checker.Def.Function" then
                 decl = def.func
             elseif def._tag == "checker.Def.Builtin" then
-                if var._type._tag == "types.T.Float" then
-                    local bname = def.id
-                    if bname == "math.pi" then
-                        return ir.Value.Float(math.pi)
-                    elseif bname == "math.huge" then
-                        return ir.Value.Float(math.huge)
-                    else
-                        typedecl.tag_error(bname)
-                    end
-                elseif var._type._tag == "types.T.Integer" then
-                    local bname = def.id
-                    if bname == "math.maxinteger" then
-                        return ir.Value.Integer(math.maxinteger)
-                    elseif bname == "math.mininteger" then
-                        return ir.Value.Integer(math.mininteger)
-                    else
-                        typedecl.tag_error(bname)
-                    end
+                local bname  = def.id
+                if     bname == "math.pi"   then return ir.Value.Float(math.pi)
+                elseif bname == "math.huge" then return ir.Value.Float(math.huge)
+                elseif bname == "math.maxinteger" then return ir.Value.Integer(math.maxinteger)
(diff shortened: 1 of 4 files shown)


pallene/to_ir.lua
```diff
@@ -863,7 +863,27 @@ function ToIR:exp_to_value(cmds, exp, _recursive)
...
+                    if bname == "math.maxinteger" then
+                        return ir.Value.Integer(math.maxinteger)
+                    elseif bname == "math.mininteger" then
+                        return ir.Value.Integer(math.mininteger)
+                    else
+                        typedecl.tag_error(bname)
+                    end
+                else
+                    error("not implemented")
+                end
```
hugomg, line 886: Would this also work if we had a single level of if-else-lf? In theory, if we know the `bname` then we should know the type.

```lua
elseif def._tag == "checker.Def.Builtin" then
    local bname  = def.id
    if     bname == "math.pi"   then return ir.Value.Float(math.pi)
    elseif bname == "math.huge" then return ir.Value.Float(math.huge)
    elseif bname == "math.maxinteger" then return ir.Value.Integer(math.maxinteger)
    elseif bname == "math.mininteger" then return ir.Value.Integer(math.mininteger)
    else
        error("not implemented")
    end

hugomg (commented):
This was simpler to implement then I expected it would be. It is always a nice surprise when that happens.

pallene/to_ir.lua
ewmailing: Okay, I flattened the if-else and committed. 

My original implementation was very nasty because I was trying to inject literal strings in the code generation so I could use the C math.h M_PI and other constants and wrap in arbitrary additional macros. I actually wrote/hacked a new ir.Value.Literal type so I could support this.

Then I realized I could just use ir.Value.Float on the Lua math.* values as long as Pallene was writing enough digits to file, which it appears to be. So this created the check for the Float type you see in the if-else case.

Then I forgot that there were both float and integer constants, which led to the Integer type check. 

Then I deleted all the really nasty Literal code. But that's how the old if-else structure got created.

The only edge case I can think of with this flattened case is if the user tries to assign over one of these constants. But I don't know if the old if-else would handle that any better. I hadn't really thought about it.

hugomg: That string thing was actually considered doing at some point, to keep the literals in the generated code similar to what the programmer wrote. It is strange when 0.1 renders as 0.10000000003 or something like that... But in the end we settled for a slightly more inteligent routine that prints just enough digits. Also, Lua supports hexadecimal floating point constants, which is something that we can't pass verbatim to C (some compilers don't accept that).

As for assigning to the math constants, I believe assigning to the math constants should be forbidden in Pallene. Do you know if we already have test cases for that?

As for what happens if Lua reassigns the math constants, we have decided that we won't worry about that kind of thing. Pallene assumes that the standard library was not messed with. If the Lua programmer does naughty things to the standard library, we trust that Santa is watching and will send them coal for christmas.  🎅

PR #518, opened 2022-04-21
Conform to latest Luacheck

hugomg:
Our CI started failing because some recent version of Luacheck (don't know which one) changed how it deals with unused variables. This is what I noticed:

1. By default, Luacheck now considers that a '_' prefix marks variables which are unused. We no longer need to handle this special case in our luacheckrc.
2. There is a new warning category (214) that complains about using a variable that was marked as being unused. There were two places in our code that were affected by this. The first was a variable called "_recursive", which I renamed without the underscore. The second was _ENV, which I told luacheck to ignore in the configuration file.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -19,10 +19,9 @@
 -- [3] https://github.com/mpeterv/luacheck#editor-support
 
 ignore = {
-    "211/_.*",  -- Unused local variable, when name starts with "_"
-
-    "212/_.*",  -- Unused argument, when name starts with "_"
     "212/self", -- Unused argument "self"
+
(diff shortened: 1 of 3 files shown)


hugomg:
ps.:  If it has been a while since the last time you updated Luacheck in your local machine, it might be a good idea to re-install / update it. I only noticed this problem when I pushed this to the CI, because in my local machine the lint was passing.

agladysh:
_ENV seems like a bug in luacheck. Did you report it?

hugomg:
It hadn't cross my mind, but now that you mention it I definitely should. Thanks for reminding me of that!

PR #529, opened 2022-04-28
Use l_likely, l_noret

### LINKED ISSUE #525: Replace PALLENE_LIKELY and PALLENE_NORETURN by standard Lua macros?
Looks like Lua has recently introduced some macros for this... Namely `l_likely` (from luaconf.h) and `l_noret` (from llimits.h).

hugomg:
We no longer need to define our own version of these macros. Fixes #525.

### DIFF
--- a/vm/src/pallene_core.h
+++ b/vm/src/pallene_core.h
@@ -29,38 +29,26 @@
 
 #include <math.h>
 
-#define PALLENE_NORETURN __attribute__((noreturn))
-#define PALLENE_UNREACHABLE __builtin_unreachable()
-
-#define PALLENE_LIKELY(x)   __builtin_expect(!!(x), 1)
-#define PALLENE_UNLIKELY(x) __builtin_expect(!!(x), 0)
-
 const char *pallene_tag_name(int raw_tag);
 
-void pallene_runtime_tag_check_error(
+l_noret pallene_runtime_tag_check_error(
     lua_State *L, const char* file, int line, int expected_tag, int received_tag,
-    const char *description_fmt, ...)
-    PALLENE_NORETURN;
+    const char *description_fmt, ...);
 
-void pallene_runtime_arity_error(
-    lua_State *L, const char *name, int expected, int received)
-    PALLENE_NORETURN;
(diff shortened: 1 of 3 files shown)


vm/src/pallene_core.h
```diff
@@ -29,38 +29,26 @@
...
-#define PALLENE_NORETURN __attribute__((noreturn))
-#define PALLENE_UNREACHABLE __builtin_unreachable()
-
-#define PALLENE_LIKELY(x)   __builtin_expect(!!(x), 1)
-#define PALLENE_UNLIKELY(x) __builtin_expect(!!(x), 0)
-
 const char *pallene_tag_name(int raw_tag);
 
-void pallene_runtime_tag_check_error(
+l_noret pallene_runtime_tag_check_error(
```
srijan-paul, line 34: I'm late. But curious as to why we need `l_noret` at all.
What's wrong with just `void`?

hugomg: The `__attribute__((noreturn))` marks function that never return to the caller, because they raise an exception or abort the execution of the program. It is an optimization hint to the C compiler. One example where this helps is array access in a loop. When we compile down to C, a loop like this

```
for i = 1, N do
    local x = xs[i]
end
```

might compile down to this pseudocode

```
for (i = 1; i <= N; i++){
    arraypart = xs->arraypart
    x = arraypart[i]
}
```

The compiler can deduce that the `xs->arraypart` is loop invariant and lifts it outside the loop:

```
arraypart = xs->arraypart
for (i = 1; i <= N; i++){ 
    x = arraypart[i]
}
```

The problem is that this only works if there are no function calls inside the loop. If the original program is

```
for i = 1, N do
    local x = xs[i]
    f()
}
```

the C compiler assumes the worst case scenarion, that the f() function can change the contents of the `xs` table. It stops being able to lift the `->arraypart` out of the loop. Functions marked with noreturn don't have this problem though, because the C compiler knows doesn't have to worry about what happens after those functions return. We know that the function doesn't return.

srijan-paul: Aha! Thanks for the explanation. I did not know that :D

Issue #516, opened 2022-04-21
Cross-platform replacement for c_compiler.lua?


hugomg:
If we ever want to make Pallene more cross-platform, the c_compiler.lua part will be one of the sticking points. Currently, windows support is completely missing and the macOS support is not actively tested. (None of the devs still use macOS and it is not tested in our continuous integration). This is all very unfortunate, and it is a problem area that is out of our area of expertise. We should consider delegating this functionality to someone else. Perhaps Luarocks? It knows how to compile C code in a cross-platform way...

Mehgugs:
You could use zig's c compiler frontend to do the cross compilation. You can invoke `zig cc` with the same options as clang/gcc and compile C code on any platform zig supports (targeting any platform zig supports). 

I personally think using luarocks is not the solution because the onus is on the user to teach luarocks how to compile C modules on their system, for windows users this often requires a lot of extra work.

hugomg:
Could you talk more about why Luarocks would be more difficult on Windows? I thought Luarocks also knew how to compile on Windows?

Also, another reason we were thinking about Luarocks is that it knows where to find the Lua interpreters and header files. Even if we used zig cc, maybe we'd still need some help from Luaricks to sort out the compilation flags.

Mehgugs:
Well given that pallene wants you to build a customized version of lua from source why not just have it be built as part of the set up?
Then you can make sure all the headers (plus luacore.h) are known to the pallene system.

hugomg:
Sorry, I'm not sure I understand 100%. Currently we have to install the custom Lua by hand and the Pallene compiler using Luarocks. Are you suggesting that the whole thing (custom Lua + pallene compiler) should be packaged as a Luarocks rock? Or that we should ditch Luarocks and install the compiler with a custom installation script?  Or sometjing else?

Mehgugs:
Okay so what I'm saying is:

zig cc is a good compiler frontend for cross platform c compilation. It is also very easy to use on windows. 

I would like this project to be usable without luarocks since I – and many other prospective users I imagine – already have lua environments set up. Zig's c compiler seems like the best way to achieve this.

hugomg:
I see. I suppose we can consider Zig if at some point we decide that we should bundle a C compiler into the project.

In the mean time, one escape valve that already exists is that you can compile Pallene to C using --emit-c. This is 100% portable and then you can compile the C code however you want.

srijan-paul:
[...]

Issue #18, opened 2018-10-02
We should check at module-load time if the runtime library version is correct


hugomg:
When a Pallene library is linked with a version of the runtime library that is not the same version that it was compiled for, it is very easy to get segmentation faults at execution time. We should protect against this by checking the library versions at module-load time.

To do this we would need some way to represent the library version at runtime. Would adding a `RUNTIME_VERSION` in `tcore.h` be good enough?

gligneul:
`RUNTIME_VERSION` should be enough. However, we should only worry about this when we release a version. This is a minor issues for devs.

ahmedkrmn:
@gligneul Do you suggest working on this now or leaving it for later since you've said it's a minor issu?

hugomg:
Now that I think about it, I'm not sure if adding a RUNTIME_VERSION will help with development, since it is very easy to firget to increment the version when making other changes to the library.

It might be more useful once we release a stable version of Pallene, but we can think about the proper approach when we get to that point.

For now, I feel like we can leave this open, but I'm removing the "good first issue" tag.

gligneul:
@hugomg we don't need to update the variable every time we change the file, just when we release a new Pallene versions. What would be a proper approach?

hugomg:
My original motivation for opening the issue was that, during development, sometimes I'd forget to recompile the library after making changes to it.

But yes, from the point of view of an user using a release of Pallene, it eould only need yo be updated when a new release is made.

gligneul:
I'm marking this as a discussion since it is not clear what should be done. I'm thinking about creating two new issues, one to solve the release problem, which should be simple, and another one to solve the dev problem, which I don't have a solution in mind yet. What do you think?

PR #534, opened 2022-05-14
Rework the installation process

### LINKED ISSUE #16: Allow the compiler to be called from any directory

hugomg:
This pull request allows the Pallene compiler to be called from any directory, not just the root of the repository. This fixes #16. Hooray! To accomplish this, I did the following things:

The first thing I did was move the custom Lua interpreter to [a separate repository](https://www.github.com/pallene-lang/lua-internals). This custom Lua only exposes the Lua internals and can be used for both Pallene and LuaAOT. This custom Lua only needs to be updated sporadically, whenever a new version of Lua comes out.

The second thing I did was to move all the things in the pallene_core.c into the pallene_core.h. Effectively, this statically links the pallene standard library into the generated modules. Maybe it is a bit bloated, but it has the advantage that when we build pallenec itself, we don't need to compile any C code.

The third thing I did was to copy paste the contents of the pallene_core.h into the generated C code, instead of using an #include directive. (Btw, I also renamed it from pallene_core.h to pallenelib.lua) This produces a larger C file but has the advantage that we don't need to worry about passing any -I flags to the compiler.

Some limitations that we can fix later:

1. The custom Lua must be installed to /usr or /usr/local. This way the C compiler can find the header files. To support other Lua installation locations we will need to add a way to customize the C compiler flags. Or find a way to ask Luarocks to compile the files for us ([#516](https://github.com/pallene-lang/pallene/issues/516)).
2. We still need to add some code to check if the Lua release matches the one that Pallene expects ([#18](https://github.com/pallene-lang/pallene/issues/18)).
3. If the Pallene library grows, we may want to create a reusable shared library for the parts that aren't inline functions. (We could also find ways to reduce the size of the library. For example, by reducing the number of builtins that we implement)
4. The test suite is running substantially slower. I don't know the cause yet. Maybe it is slower compilation of the generated file?


README.md
```diff
@@ -14,97 +14,85 @@ Compared to LuaJIT, Pallene aims to offer more predictable run-time performance.
...
 
-### Compiling the custom interpreter
+Remember that in order for the local rocks tree to work, you must to set some environment variables
 
-Pallene must be run against a custom-built version of the Lua interpreter.
-This custom version of Lua 5.4 doesn't have to be the same one that you will use to run the compiler itself,
-or to install the Luarocks packages.
+```sh
+# Put this in your .bashrc
+eval "$(luarocks path)"
```
srijan-paul, line 68: (Alternatively)
```sh
echo "eval\"$(luarocks path)\"" >> ~/.bashrc
```

pallene/pallenelib.lua
```diff
@@ -0,0 +1,623 @@
+-- /*
```
srijan-paul, line 1: returning C code from a Lua code is a bit weird when viewing in a text editor 😅 
How would you feel if this was a regular old C file and we read it using `io` before returning the code?

srijan-paul (changes requested):

README.md
hugomg: I think I prefer telling people to do this by hand, because with the `>>` it is easy to run it more than once and end up with multiple redundant lines inside the bashrc. In any case, I tweaked the wording to make it just a little bit clearer.

pallene/pallenelib.lua
hugomg: Yeah, it is certainly awkward. I have to manually switch the syntax highlighting color in my editor. The main problem is where to install the file, and how to find the path to the file after it is installed. I believe that with Luarocks you need to use a third-party library: https://github.com/hishamhm/datafile. I didn't want to add a dependency just so we could use this one file, but maybe it's clearer than the current hack... What do you think?

srijan-paul: I'm also in favor of having inline C code over adding a dependency.
But I also wonder why we can't just do `io.open("./pallene.h", "r")` to read the header.

README.md
srijan-paul: Gotcha

srijan-paul (approved):

pallene/pallenelib.lua
hugomg: One of the things I want to achieve with this PR is to no longer care about the current working directory (issue #16). From now on, using relative paths is no longer allowed.

hugomg:
Cool. The next step I'm planning after this PR is #516 and it will be very helpful to be able to test that on a Mac.

PR #536, opened 2022-05-19
pallenec: Remove --dump flag

hugomg:
This commit removes the rarely-used --dump pallenec command flag. If you ain't gonna need it, better delete it...  If a Pallene developer still wants to see this info, they can add some temporary print statements to the driver.lua.

One thing we gain by removing this code is that it gets rid of our dependency on the "inspect" library. Another is that the we get rid of the "list_of_compiler_passes" array, which was always out of date anyway.


srijan-paul:
Personally I liked having the `--dump` flag :p
But I guess we can do without it.

srijan-paul (approved):

PR #539, opened 2022-05-21
In --print-ir, explicitly show CallStatic and CallDyn

hugomg:
Previously, --print-ir didn't distinguish between CallStatic and CallDyn

    x2 <- f(x1)

Now it shows which one it is.

    x2 <- CallStatic f(x1)
    x2 <- CallDyn g(x1)


hugomg:
This was prompted by the need to debug what was happening in #414.

PR #540, opened 2022-05-21
Take advantage of Rockspec 3.0 features

### LINKED ISSUE #418: Ensure rockspec is up to date with the Pallene source code tree
We often forget to update the list of Lua files in the rockspec when we create a new file. We should either:

1. Rewrite the rockspec so that we don't need to list every single file
2. Add a linter check that catches if a file is not in the rockspec

I think that Luarocks 3.0 might have a way to do option (1), which is something that wasn't possible back when we first wrote the Rockspec file. However, I'm not 100% sure.

hugomg:
A couple of changes to our rockspec:

1. Move Lua files to a src/ directory. In the 3.0 rockspec format, this spares us from having to manually list every module.
2. Move pallenec to the bin/ directory. This should put a stop to the bad habit of calling ./pallenec directly.
2. Tell Luarocks about our test suite. This enables `luarocks test`.

Fixes #418.


hugomg:
I had hesitated about adding a `src/` directory, because it means that we need to call `luarocks make` every time we edit a Lua file. We no longer can take advantage of the `pallene` directory being in the default LUA_PATH, if we invoke pallenec from the root of the repository.

However, now that we promote using `luarocks make` to build Pallene, we can't use that trick anyway. Luarocks by default gives a greater priority to the rocks tree than to the current working directory, so if you ever install Pallene with `luarocks make` then require will use that instead of the current working directory. Which gets us to this PR. Let's embrace `luarocks make` and benefit from not having to manually list all the modules in the rockspec...

PR #541, opened 2022-05-21
Check that the installed Lua version is the correct one.

### LINKED ISSUE #18: We should check at module-load time if the runtime library version is correct

hugomg:
This adds a compile-time check to see that the installed version of Lua matches the one that is supported by the current version of Pallene. It also adds a run-time check that raises an error in case you update your Lua but do not recompile your Pallene module. Fixes #18.

ATTENTION: I had to add a luaL_checkcoreversion function to the lua-internals. This means that you must update your lua-internals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ on:
         branches: [ master ]
 
 env:
-    LUA_VERSION: 5.4.4
+    LUA_VERSION: 5.4.4-2
     LUAROCKS_VERSION: 3.9.0
 
 jobs:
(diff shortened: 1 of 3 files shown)


hugomg:
One thing that is a bit awkward here is that we have to do simultaneous PRs in two different repositories, this one and the lua-internals one. But it should be an uncommon thing so I think it's OK.

srijan-paul (approved):

PR #543, opened 2022-05-24
Lint the pallenec

hugomg:
We had left the pallenec out of the list of Lua files in the run-lint, which allowed it to accumulate a bunch of Lint errors. To stop this from happening again, I moved the bulk of the pallenec into a module inside the src/ directory, where the linter can see it. The actual bin/pallenec now is a small stub that doesn't need to be linted.

I considered just telling the linter to add the pallenec to the list but that was a bit awkward. Because the pallenec doesn't have a ".lua" extension, it would not be matched by the globs.


hugomg:
The combined diff might be a bit hard to read, because we are moving the pallenec file and also making changes to it. It is easier to see what is going on by looking at the individual commits.

srijan-paul (approved):
Kool refactor!

Issue #551, opened 2022-09-21
Reproducible case that can trigger Lua internal assertion: "results from function overflow current stack size" (LUAI_ASSERT)


ewmailing:
This bug is discussed in https://github.com/pallene-lang/pallene/discussions/547
I have a reproducible case where something in Pallene seems to trigger Lua's internal LUAI_ASSERT.

This bug is a nightmare to isolate.
I have eliminated 99% of the program and hope this reproducible test case is now simple enough to use for diagonsing the problem.

Make sure to enable the LUAI_ASSERT if you want to trigger it.

- The original program used records, but to create this stripped down case, I made every thing use tables. Since I still hit the assertion, the problem is not related to records.
- The original program did tons of work to generate a data structure, which is fed into this final routine for processing, which is were the assertion failure happens. I have deleted all that code so there is no program, but I have serialized a dataset that would lead to the assertion failure. This new test case reloads all the data into a table so the problematic function can be called. My most trivial datasets were not able to trigger the assertion, so there is some particular amount of data needed, or a specific type to hit enough of the conditional cases in my function to trigger the assertion.
	- the serialized data is stored in debug_serialize_backTestPlot.lua (as Lua syntax)
- You need to build all the .pln files (ta, options, datetul, stocks_stackoverflow2)
	- ta and options modules are not actually used since I removed all the other code, but I didn't feeling like re-writing all the interfaces to finish removing them from the code.
- Removing too much code can actually cause the assertion to not trigger.
- Adding or moving code can also cause the assertion to happen somewhere else or not at all.
- The line it crashes at is a call into dateutil.IntegerToDate, but don't be fooled. This is called earlier, and was called a lot in my program. I also tried calling it a whole bunch of extra times earlier in this function, and it didn't do anything. So I don't think the routine itself is the problem. 
- Sometimes I was able to cause the assertion to happen earlier by declaring more local variables a little earlier and then moving the dateutil.DateToInteger call earlier right after those, e.g.
local tmp_arr1:{string}
local tmp_arr2:{string}
	I have removed a lot of code in the functions I provided you, so it may be a little harder to recreate this situation.

- The functions you care about are in stocks_stackoverflow2.pln.
	- GenerateBackTestPlotForChartJS
	- GenerateCodeForChartJSSeries

- These functions used to also call helpers in BackTesterChartJS_stackoverflow.lua, but I think I removed/disabled all the calls in this example.
[...]

PR #557, opened 2022-10-18
Fix a bug where the stack would not grow when it should

hugomg:
We were using the luaD_checkstack macro incorrectly. That macro asks for the total height of the Lua stack, all the way to the Lua entry point. The result of the bug was that we would not grow the Lua stack when we should, which could surface as a segmentation fault.

While debugging this, I found that address sanitizer was very helpful. I have also added some documentation describing how to run Pallene with address sanitizer.


ewmailing:
I just tested this, but unfortunately it does not fix #551. All 3 test cases I submitted in #551 still fail at LUAI_ASSERT. And my real program continues to fail at LUAI_ASSERT. I did remember to delete all my old .so files and rebuild all the .pln files.

I did execute run-tests, and it did pass.  (Total: 39 warnings / 0 errors in 49 files). I speculate your new test case is still missing something that the my submitted test cases are doing.

hugomg:
Oh well, maybe we still have more work to do then...?

But just do be sure: did you get to test my new test case with the LUAI_ASSERT? I didn't test with that setting enabled so I don't know what happens there. Also, have you tried testing those examples with LUAI_ASSERT disabled to see if they still segfault?

ewmailing:
> But just do be sure: did you get to test my new test case with the LUAI_ASSERT? 

Yes, I built with that setting enabled and executed run_tests. All the tests appeared to complete, as I wrote earlier.

> Also, have you tried testing those examples with LUAI_ASSERT disabled to see if they still segfault?

I tried running the 3 test examples I submitted with LUAI_ASSERT disabled and they did not segfault. I also built using the address sanitizer, but I did not see anything appear from that.

Because my original segfaults only happen sometimes, I actually suspect your fix may actually make it harder to reproduce seg faults since you have probably fixed one of the corruption sources. But the LUAI_ASSERT has been a more reliable indicator that something is wrong.

Also, I am trying to run my real program again, with the address sanitizer, but the flag setting is so slow, I might have to give up. At full speed, my runs can take hours before I hit a seg fault (if I even hit it). But with address sanitizer, I may be running a 1/10000 the speed so I may never get to see anything.

hugomg:
Address sanitizer shouldn't slow things down by that much. Maybe you need to add -O2 to the cflags?

I'll try running the tests with the assertions enabled to figure out what's up with those.

ewmailing:
It looks like my projected runtime estimates in my program were thrown off a lot because of really slow startup speeds being impacted by address sanitizer. Speed picked up, but looked liked it was still in the 10x-100x slower ballpark. I think my use of Lua Lanes is disproportionately punished because I suspect each new Lane creation has to allocate/copy a bunch of memory to the new lane and bottlenecked the whole program so I wasn't getting much parallelism, which is a large multiplier factor in my program's runtime.

I disabled address sanitizer (and disabled LUAI_ASSERT), and tried doing a roughly 7 hour run of my program. It managed to complete without a segfault. Usually I don't make all the way through, so your fix might have improved this aspect, although since the crash is pretty random, it is hard to say for sure. Sometimes the crash happens almost immediately, and sometimes it never shows up.

hugomg:
On second thought, I think my original solution was not correct. Could you please try the one that I just pushed? Hopefully it should fix both the segfaults and the assertions now :)

ewmailing:
I think this new solution is really close... 2 of the 3 reproducible test cases I submitted now pass without triggering the LUAI_ASSERT. And also, my main program doesn't seem to be triggering the LUAI_ASSERT any more.

But the second reproducible test case I submitted (stackoverflow2) triggers the LUAI_ASSERT message:
```
lua: lapi.c:1015: lua_callk: Assertion `((nresults) == (-1) || (L->ci->top - L->top >= (nresults) - (nargs))) && "results from function overflow current stack size"' failed.
Aborted (core dumped)
```

I admit, I'm a little surprised that the second test case is the only one that triggers the problem, because if I recall, it was derived from the first test case by trying to further simplify the first test case.

hugomg:
Debugging it a bit more, I think the stackoverflow2 bug might be a bit different. Before it was an off-by-many error but now it looks like an off-by-one error. But now I wonder if the fault is in the assertion itself, not in our code?

The assertion is checking if there is enough stack space to write the return values of the function: `L->ci->top - L->top >= nresults - nargs`. But shouldn't it be `nresults - (nargs+1)`? The lua_call API pops (nargs+1) items from the stack and then pushes (nresults). Currently we're hitting the assertion exactly when we hit that limit.

hugomg:
Anyway, at least we have a minimal example now:

*Pallene*

```lua
local m = {}

typealias func = (integer) -> (nil,nil,nil)

function m.f(i: integer, g:func): string
    if i == 0 then
        return ""
    else
        local x = tostring(i)
        local a,b,c = g(1)
        local z = m.f(i-1, g)
        return x.." "..z
    end
end

return m
```

*Lua*

```lua
local foo = require "foo"

local function g()
    return nil,nil,nil
end

print(fat.f(64, g))
```

ewmailing:
> Debugging it a bit more, I think the stackoverflow2 bug might be a bit different. Before it was an off-by-many error but now it looks like an off-by-one error. But now I wonder if the fault is in the assertion itself, not in our code?
> 
> The assertion is checking if there is enough stack space to write the return values of the function: `L->ci->top - L->top >= nresults - nargs`. But shouldn't it be `nresults - (nargs+1)`? The lua_call API pops (nargs+1) items from the stack and then pushes (nresults). Currently we're hitting the assertion exactly when we hit that boundary.

I am skeptical this would be in a bug in Lua because this one seems bad enough that it should have been caught in the wild by now. But I guess you can always ask the Lua team :)


So I am not that intimate with the Lua internals. But glancing at the implementation for lua_call, I don't see where it pops (nargs+1) and pushes nresults.

The closest related lines that I see are:

1. In lua_callk where there is a (nargs+1) used:
```
  func = L->top - (nargs+1);
```
But my interpretation of that this is just to get the stack position of the Lua function to call, not manipulating the nargs or nresults.

2. Near the end of lua_callk, there is a macro call to adjustresults which looks like how the results stack gets manipulated.
```
#define adjustresults(L,nres) \
    { if ((nres) <= LUA_MULTRET && L->ci->top < L->top) L->ci->top = L->top; }
```
It's not clear to me how top and ci->top correspond to nargs, but I don't see a (nargs+1) here.

hugomg:
Well, I think you're right :) Earlier today I couldn't find where that `+1` was hiding but talking to a skeptical mind does help show the light... Anyway, I now believe the `+1` we're looking for is hidden inside the luaD_poscall in the ldo.c, in the part where it calls the "moveresults".

Basically, to call a Lua function you first push the function closure to the stack and all the arguments:

    f a1 a2 a3 a4

Then the function runs and pushes the return values to the stack. During the computation it may delete some of the input items from the stack and may push some temporary variables (v1, v2, etc)

    f a1 a2 v1 v2 v3 r1 r2 r3

Finally, when we're ready to return to the caller, the poscall clears the stack and moves the return vals to the bottom

    r1 r2 r3

The overall effect of a function call on the stack is that it removes (narg+1) items and then adds (nret) items . However, there is an intermediate step where you need at least (nret+1). Even if in step 2 the function tries to use as little stack space as possible, by popping all the arguments and using no temporary variables, it will still need at least (nret+1) because of the function closure.

    f r1 r2 r3

----

tl;dr: Previously Pallene reserved only (nret) stack slots for a function call. I have just pushed another commit to change this to (nret+1) stack slots. Hopefully that should finally fix the bug!

ewmailing:
Okay, I think I understand that. All the arguments get removed and the return values get moved into place. But the function must also get removed and the arguments will need to shift over one slot further. And due to Lua implementation details, the function needs to remain on the stack longer than anticipated, so you can't just remove all the arguments and function at the same time.

Anyway, I retested all 3 test programs, plus my real program, and I no longer trip the LUAI_ASSERT. I think this bug is fixed.

Thank you so much for your efforts on this.

P.S. Trivia: Re-disabling LUAI_ASSERT made one of my test runs go from 25 minutes to 22 minutes ~= 12% decrease.

bjornbm:
Wow, well done guys! 👍

PR #559, opened 2023-01-29
Improve wording

bjornbm:
I'm not sure about the second edit. Maybe you want it to say “anys”, but it looks weird to me?


hugomg (approved):
Your gut feeling was right. That line is much better without the "anys". Thanks!  :)

PR #560, opened 2023-02-07
More flexible consistency rule

hugomg:
This commit loosens the type consistency restriction for type casts. Previously we used the rule from Gradual Typing, which required that both types be equal up to 'any'. In the new rule we only check if the type tag matches. Casting to and from "any" is still allowed. Furthermore, the new rule also allows casts from one array type to another, even if the array contents are different. We also between any pair of function type no matter the type of argument or arity.

The motivation for the new rule is that Pallene type checks only check the type tag anyway. The type casts that used to be forbidden but are now allowed are type casts that theoretically might run without error at runtime. Now we only forbid type casts that are guaranteed to fail, such as int=>string or array=>function.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -757,28 +757,6 @@ describe("Binary operator", function()
 
 end)
 
-describe("Cast operator", function()
-
-    local function test(typ1, typ2, expected_error)
-        local description = string.format(
-            "'as' cannot convert incompatible types (%s, %s)", typ1, typ2)
-
-        local code = util.render([[
-            function m.fn(x: $typ1)
-                local _ = x as $typ2
-            end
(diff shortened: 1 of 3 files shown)


srijan-paul:
We should still be able to cast from `int` to `float`, though. No?

hugomg:
The current version of the `as` operator already forbids conversions between ints and floats.  The way to convert from int to float is to multiply by `1.0`. And to convert float to int you can use the `math.floor` and friends.

I think there used to be a time a while ago when we used `as` for int<->float conversions, but I don't remember when that changed.

srijan-paul:
Hm, I see.
I think this looks good, in that case.

hugomg:
Alright, thanks!

PR #566, opened 2023-03-19
Add a type-checking command line flag

hugomg:
Previously we used --stop-after, but we removed that... A flag to stop the compiler after the type-checking step might be useful for text editor and linter support.


hugomg:
ps.: I still need to add tests before this can be merged.

bjornbm:
In reply to https://github.com/pallene-lang/pallene/issues/277#issuecomment-1475333198

I don't think this adds much for the purposes of https://github.com/pallene-lang/pallene/issues/277#issuecomment-1475304943. It is nice perhaps to not generate the .lua files, but they aren't/weren't causing much trouble either. For sure this is cleaner if one does not intend to run additional Lua tools over them.

bjornbm:
Actually, it does add significant value if the checking is more rigorous that that of `--emit-lua`, but I would think that `--emit-lua` should do full checking (it currently does not as noted in https://github.com/pallene-lang/pallene/issues/567#issuecomment-1479756211).

PR #568, opened 2023-03-23
Mark Pallene code blocks as lua for highlighting

bjornbm:
Marking the Pallene code blocks as `lua` code provides better-than-nothing syntax highlighting on Github (and other tools). Obviously Pallene specifics (`typealias`, `record`, types…) do not get highlighted as keywords.

I didn't put `lua` on every codeblock. For example, the `record` block would look strange with only `end` highlighted.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -15,7 +15,7 @@ The exception are the times when Pallene produces an error, either a compile-tim
 Here is an example Pallene subroutine for summing the elements in an array of floating-point numbers.
 Note that type annotations are required for function argument and return types but for local variable declarations Pallene can often infer the types.
 
-```
+```lua
 local m = {}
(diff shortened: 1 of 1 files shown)


hugomg:
Could you please post a screenshot of how the record block looks? If it's not too yoo bad I would be inclined to use lua blocks for everything.

bjornbm:
I created pull request #569 will all Pallene code block marked `lua`, so you can chose whether to merge that or this one. You can compare the rendering at:

* #568: https://github.com/bjornbm/pallene/blob/patch-3/doc/manual.md
* #569: https://github.com/bjornbm/pallene/blob/patch-4/doc/manual.md

But for your convenience and as requested, from #569:

image

hugomg:
Closed in favor of #569

Issue #567, opened 2023-03-22
Add flag --emit-lua-preserve-columns to benefit from Lua diagnostics tools


bjornbm:
Opening an issue based on discussion at: https://github.com/pallene-lang/pallene/issues/277#issuecomment-1475304943

For “cheap” editor support, I am considering running `--emit-lua` and then, for example, run [Luacheck](https://github.com/mpeterv/luacheck) on the .lua file(s) and use this to drive the editor (for example, [neo]vim quickfix functionality). @sumneko's [Lua language server](https://github.com/LuaLS/lua-language-server) can be jury-rigged for Pallene in a similar fashion (roundtripping via `--emit-lua`).

The diagnostics (and editor highlighting/cursor positioning) could be more accurate if the columns were preserved in the Lua output. Basically the point of this issue is to add a flag to revert the change in #277. The proposed name is `--emit-lua-preserve-columns`. It is long, but better descriptive than short, I guess. Other suggestions are welcome.

@hugomg said “maybe it's useful if there are extra error messages you can get from luacheck but not from pallenec.” As a motivating example:

```lua
local m = {}

local function f(x : any)
   local y : integer
   local x : integer = y
end

return m
```

Luacheck output:

```
Checking lck.lua                                  6 warnings

    lck.lua:1:1: setting undefined field ln of global math
    lck.lua:3:16: unused function f
    lck.lua:3:18: unused argument x
    lck.lua:4:10: variable y is never set
    lck.lua:5:10: unused variable x
    lck.lua:5:10: variable x was previously defined as an argument on line 3

Total: 6 warnings / 0 errors in 1 file
```

In this simple case the column numbers were OK. The point was to demonstrate that Luacheck adds value on top of `pallenec`. (The first warning should to be ignored with Luacheck ignore pattern.)

A PR will follow (although feel free if anyone is less time constrained).

Note that I'm not using any of this in my editor yet. I might try to add it in @JLPLabs' [pallene-vim](https://github.com/JLPLabs/pallene-vim).

bjornbm:
I also realize now that `pallenec --emit-lua` does not perform full checking. In particular, those of `unitialized.lua`. For example, if does not catch the use of unassigned `y` and the absence of a return in the above example. I think this warrants a new issue?

hugomg:
I agree that this deserves a separate issue. I also expect that it shouldn't be too difficult to implement. Add a couple of lines to the compile_pln_to_lua function in driver.lua

```lua
[...]

PR #571, opened 2023-04-05
Include variable verifications in checking

bjornbm:
Applies to `--only-check and `--emit-lua` flags.

Fixes the observation in https://github.com/pallene-lang/pallene/issues/567#issuecomment-1479756211, and also improves on #566 which also didn't do full type checking.

There are some repeated compilation steps performed now by `--emit-lua`. A refactor allowing to resume compilation midway by passing the intermediate `prog_ast` could avoid this, but I'm not convinced the complexity would be worth it just for the sake of --emit-lua.

### DIFF
--- a/spec/translator_spec.lua
+++ b/spec/translator_spec.lua
@@ -483,6 +483,7 @@ local m: module = {}
 local xs: {any} = {10, "hello", 3.14}
 
 local function f(x: any, y: any): any
+    return nil as nil
 end
 
 return m
@@ -492,6 +493,7 @@ local m = {}
 local xs = {10, "hello", 3.14}
 
 local function f(x, y)
+    return nil
 end
 
 return m
--- a/spec/translator_spec.lua
+++ b/spec/translator_spec.lua
@@ -483,7 +483,7 @@ local m: module = {}
 local xs: {any} = {10, "hello", 3.14}
 
 local function f(x: any, y: any): any
-    return nil as nil
+    return nil
 end
 
 return m
@@ -496,6 +496,24 @@ local function f(x, y)
     return nil
 end
 
+return m
+]])
+    end)
+
+   it("Remove casts in return statement", function ()
+        assert_translation(
+[[
+local m: module = {}
+local function f(): (any, any)
+    return nil as nil, 1 as integer
+end
+return m
+]],
+[[
+local m = {}
+local function f()
+    return nil, 1
+end
 return m
 ]])
     end)
--- a/src/pallene/parser.lua
+++ b/src/pallene/parser.lua
@@ -285,7 +285,7 @@ function Parser:Program()
                 exp.var._tag == "ast.Var.Name" and
                 exp.var.name == modname)
             then
-                -- The checker also needs to check that this name has not been shadowed
+                -- The type checker also needs to check that this name has not been shadowed
                 self:syntax_error(exp.loc,
                     "must return exactly the module variable '%s'", modname)
             end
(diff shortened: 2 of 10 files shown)


spec/translator_spec.lua
```diff
@@ -483,6 +483,7 @@ local m: module = {}
 local xs: {any} = {10, "hello", 3.14}
 
 local function f(x: any, y: any): any
+    return nil as nil
```
hugomg, line 486: Could this be just `return nil`, without the `as`?

hugomg (changes requested):
Thanks!

We should probably also fix the root cause of this: the `checker.lua` module and the `"checker"` step in the compilation pipeline have an outdated name. They are no longer the only place doing compile-time checking. What do you think about renaming them to or `typechecker` or `typer`?

bjornbm:
I think that is a good idea. I was thinking the same, that the names are confusing.

I will admit that I didn't look in detail, but my impression is that `checker` does syntax checking while `uninitialized` does the type checking. So `syntaxchecker` and `typechecker` could be better names. But there may be more going on in either or both.

spec/translator_spec.lua
bjornbm: It could be, but I added `as nil` to increase the testing of this test case. As the removal of `as …` syntax was not tested before.

Of course, I believe I could remove this line and the function return type to be closer to the original test case, which would also remove the trailing white space but arguably isn't testing much (such as the type at EOL corner case).

hugomg: A test for EOL stuff sounds like a good idea but I feel it deserves a separate testcase (`it` block with appropriate name)

bjornbm: Added testcase in c43d3d1d18f5867d2a2d8566c77aa24e7e8a0210.

bjornbm:
I think I have addressed your requests/suggestions in this pull request, @hugomg.

src/pallene/parser.lua
```diff
@@ -285,7 +285,7 @@ function Parser:Program()
                 exp.var._tag == "ast.Var.Name" and
                 exp.var.name == modname)
             then
-                -- The checker also needs to check that this name has not been shadowed
+                -- The type checker also needs to check that this name has not been shadowed
```
bjornbm, line 288: I wasn't sure about the meaning of this comment, and if it relates to typechecker or if something else was meant by “checker”?

src/pallene/typechecker.lua
```diff
@@ -974,7 +974,7 @@ function Checker:check_exp_synthesize(exp)
         -- We check the child expression with verify instead of synthesize because Pallene cases
         -- also act as type annotations for things like empty array literals: ({} as {value}).
         -- However, this means that the call to verify almost always inserts a redundant cast node.
-        -- To keep the --dump-checker output clean, we get rid of it.  By the way, the Pallene to
+        -- To keep the --dump-typechecker output clean, we get rid of it.  By the way, the Pallene to
```
bjornbm, line 977: I didn't see if/where this flag was implemented. There was another reference to a “--dump compiler option” in a comment in driver.lua (line 51). But I don't see any --dump flags implemented in pallenec.lua. Were they replaced by `--emit-c` and `--print-ir`?

src/pallene/parser.lua
hugomg: Good point. If we want to be picky, Scope Checking would be separate idea from Type Checking. But if you ask me it's fine to bundle it with the type checker (in our current implementation they are one and the same)

src/pallene/typechecker.lua
hugomg: Nice catch. The dump flag was removed in #536. I'll fix that comment later.

hugomg (approved):
Excellent! Thank you.

PR #575, opened 2023-05-19
Nullary tagged union constructors no longer need parens

hugomg:
For example, now we can write types.T.Any instead of types.T.Any(). It should have been like this from the beginning.

### DIFF
--- a/spec/types_spec.lua
+++ b/spec/types_spec.lua
@@ -8,41 +8,41 @@ local types = require "pallene.types"
 describe("Pallene types", function()
 
     it("pretty-prints types", function()
-        assert.same("{ integer }", types.tostring(types.T.Array(types.T.Integer())))
+        assert.same("{ integer }", types.tostring(types.T.Array(types.T.Integer)))
         assert.same("{ x: float, y: float }", types.tostring(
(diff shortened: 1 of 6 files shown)


hugomg:
I'm trying to bisect this segfault and I believe it is unrelated to this PR. It seems to also happen on older commits. That said, to be safe we should probably track down that other bug first...

hugomg:
Turns out the segfault really was because of this PR. In our gc.lua we use Ir.Cmd objects as table keys, assuming that they are all separate objects with an unique address. Having many command share the same command object caused collision in the table keys, which messed up the liveliness analysis and resulted in an use-after-free segfault.

```lua
live_gc_vars[cmd] = {}
```

The underlying issue here is that our gc.lua is trying to come up with a flat "address" for a nested IR. Either we should embrace nestedness and have a nested `live_gc_vars`, our we should switch to flat IR. Today I'm slightly tending towards the later: it's "by the book" and would be compatible with `goto`.

If we don't want to wait too long for that, a possible workaround would be to require parenthesis for ir.Cmd (e.g. `ir.Cmd.Break()`) but not require parenthesis for other things (e.g. `types.T.Integer`) .

gabrielsferre (approved):

Issue #580, opened 2023-05-31
Assertion failure on method call


bjornbm:
Test case:
```
local m = {}
local function f()
    a:x()
end
return m
```

`pallenec` output:
```
lua: src/pallene/typechecker.lua:748: assertion failed!
stack traceback:
	[C]: in function 'assert'
	src/pallene/typechecker.lua:748: in method 'check_fun_call'
	src/pallene/typechecker.lua:532: in method 'check_stat'
	src/pallene/typechecker.lua:356: in local 'body'
	src/pallene/symtab.lua:27: in function 'pallene.symtab.with_block'
	src/pallene/typechecker.lua:354: in method 'check_stat'
	src/pallene/typechecker.lua:1095: in local 'body'
	src/pallene/symtab.lua:27: in function 'pallene.symtab.with_block'
	src/pallene/typechecker.lua:1090: in method 'check_exp_verify'
	src/pallene/typechecker.lua:604: in method 'check_stat'
	src/pallene/typechecker.lua:262: in function <src/pallene/typechecker.lua:226>
	(...tail calls...)
	[C]: in function 'xpcall'
	src/pallene/trycatch.lua:64: in function 'pallene.trycatch.pcall'
	src/pallene/typechecker.lua:56: in function 'pallene.typechecker.check'
	src/pallene/driver.lua:75: in function 'pallene.driver.compile_internal'
	src/pallene/driver.lua:108: in local 'f'
	src/pallene/driver.lua:215: in function 'pallene.driver.compile'
	src/pallene/pallenec.lua:47: in upvalue 'compile'
	src/pallene/pallenec.lua:72: in function 'pallene.pallenec.main'
	bin/pallenec:3: in main chunk
	[C]: in ?
```

The test case obviously should not compile, but I believe the error message from `pallenec` is not a good one.

I understand it is somewhat a known problem (method syntax not yet implemented?), but this issue is to say that I think method syntax should result in a better error message than a failed assertion. Presumably with a not-yet-implemented message?

What follows is more FYI than any solution…

Changing `typechecker.lua` line 748 to:
```
    assert(exp._tag == "ast.Exp.CallFunc" or exp._tag == "ast.Exp.CallMethod")
```

Results in a good error message (but unrelated to the method):

```
type error: /Users/bjorn/repos/lua/test2.pln:3:5: variable 'a' is not declared
```

 with a proper method use it breaks down at a later point. Specifically at line 755 with something like:

```
[...]

PR #582, opened 2023-06-04
Improved error messages for unimplemented method syntax

bjornbm:
@hugomg, please check if this is what you had in mind for #580. I'll look into test cases once you confirm, unless you consider this non-functionality does not need tests.


hugomg (approved):
Looks great! I think this is fine without tests.

PR #583, opened 2023-06-10
Remove half-baked support for method calls

### LINKED ISSUE #581: Confusing error message referring to table as function
```lua
local m = {}
function a:f() end
return m
```

Error message:

```
syntax error: /Users/bjorn/repos/lua/test3.pln:2:1: function 'a' was not forward declared
compilation aborted due to previous error           ^^^^^^^^^^^^
```

In this case, `a` is not a function so the error message is a bit misleading.

Maybe this will “fix itself” once method syntax is implemented, and it is not worth improving the error message in the meantime?

hugomg:
The current support for ":" syntax is well-intended but buggy (See #580 and #581). Let's get rid of it for now. When it comes the time to implement method calls for real, it should not be difficult to add this stuff back in.

Closes #581

### DIFF
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -744,25 +744,6 @@ describe("Parser /", function()
         end)
     end)
 
-    describe("Method calls without parenthesis", function()
-
-        it("for string literals", function()
-            assert_expression_ast([[ o:m "asd" ]], {
-                _tag = "ast.Exp.CallMethod",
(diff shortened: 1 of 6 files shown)


hugomg:
@bjornbm please let me know if this helps fix those issues you found.

hugomg:
Great idea! Let's go with the shortest version.

Issue #151, opened 2019-12-22
Improve record tag error messages to include the record name


gligneul:
The issue can be found in the following test case in `coder_spec.lua`:
```
        it("check record tags", function()
            -- TODO: change this message to mention the relevant record types
            -- instead of only saying "userdata"
            run_test([[
                local prim = test.make_prim(123)
                local ok, err = pcall(test.get_x, prim)
                assert(not ok)
                assert(string.find(err, "expected userdata but found userdata",
                        nil, true))
            ]])
        end)
```

Akylzhan:
I can work on this issue, as I understood I need to replace the message with ```string.format("expected userdata but found $s", type(prim))``` ?

gligneul:
It is not that simple because `type(prim)` will return only the string `"userdata"`. You have to retrieve the record type at runtime. Given these definitions in Pallene:
```
record circle float x; float y; float radius end
record square float x; float y; float length end
function area(s: square): float
  return s.length ** 2
end
```
If I call the function `area` passing a circle or a string, I would like to see the following error messages:
```
m.area("foo") --> "expected record square but found string"
m.area(circle_obj) --> "expected record square but found record circle"
```

hugomg:
One direction we could go would be to store the name of the record type in the __name field of its metatable, as is done by luaL_newmetatable.

We'd also have to change pallene_tag_name and/or pallene_runtime_tag_check_error, to correctly work with these record type names. 

We might want to check if it would make sense to use luaT_objtypename to get the type name. I wasn't aware of this function when I originally wrote pallene_type_name.

PR #595, opened 2024-04-05
Record tag error message includes record name instead of only "userdata"

Vipul-Cariappa:
Fixes: #151

Code:

```lua
-- main.lua

Mod = require "Shapes"

C1 = Mod.Circle_New(3.0)
R1 = Mod.Rectangle_New(3.0, 4.0)

print(Mod.Circle_Area(R1))
print(Mod.Circle_Area(1)) 
```

```lua
-- Shapes.pln

local Shapes: module = {}

record Circle
    radius: float
end

record Rectangle
    length: float
    width: float
end

function Shapes.Circle_New(r: float): Circle
    local c: Circle = {radius = r}
    return c
end

function Shapes.Circle_Area(c: Circle): float
    return c.radius * c.radius * 3.1415
end

function Shapes.Rectangle_New(l: float, w: float): Rectangle
    local c: Rectangle = {length = l, width = w}
    return c
end

function Shapes.Rectangle_Area(c: Rectangle): float
    return c.length * c.width
end

return Shapes
```

Old behaviour
```bash
❯ lua main.lua                                  
lua: main.lua:8: file Shapes.pln: line 17: wrong type for argument 'c', expected userdata but found userdata
stack traceback:
        [C]: in function 'Shapes.Circle_Area'
        main.lua:8: in main chunk
        [C]: in ?

❯ lua main.lua
lua: main.lua:9: file Shapes.pln: line 17: wrong type for argument 'c', expected userdata but found integer
stack traceback:
        [C]: in function 'Shapes.Circle_Area'
        main.lua:9: in main chunk
        [C]: in ?
```

New behaviour
```bash
❯ lua main.lua
lua: main.lua:8: file Shapes.pln: line 17: wrong type for argument 'c', expected Circle but found Rectangle
stack traceback:
        [C]: in function 'Shapes.Circle_Area'
        main.lua:8: in main chunk
        [C]: in ?

❯ lua main.lua
lua: main.lua:9: file Shapes.pln: line 17: wrong type for argument 'c', expected Circle but found integer
stack traceback:
        [C]: in function 'Shapes.Circle_Area'
        main.lua:9: in main chunk
        [C]: in ?
```

### DIFF
--- a/src/pallene/coder.lua
+++ b/src/pallene/coder.lua
@@ -245,17 +245,23 @@ function Coder:get_stack_slot(typ, dst, slot, loc, description_fmt, ...)
     else
         assert(not typ.is_upvalue_box)
         local extra_args = table.pack(...)
+        local expected_type = pallene_type_tag(typ)
+        if expected_type == "LUA_TUSERDATA" then
+            expected_type = C.string(typ.name)
+        else
+            expected_type = "pallene_tag_name(" .. expected_type .. ")"
+        end
         check_tag = util.render([[
             if (l_unlikely(!$test)) {
                 pallene_runtime_tag_check_error(L,
-                    $file, $line, $expected_tag, rawtt($slot),
+                    $file, $line, $expected_type, pallene_type_name(L, $slot),
                     ${description_fmt}${opt_comma}${extra_args});
             }
         ]], {
             test = self:test_tag(typ, slot),
             file = C.string(loc and loc.file_name or "<anonymous>"),
             line = C.integer(loc and loc.line or 0),
-            expected_tag = pallene_type_tag(typ),
+            expected_type = expected_type,
             slot = slot,
             description_fmt = C.string(description_fmt),
             opt_comma = (#extra_args == 0 and "" or ", "),
@@ -1771,11 +1777,13 @@ function Coder:generate_luaopen_function()
         if     tag == "coder.Constant.Metatable" then
             is_upvalue_box = upv.typ.is_upvalue_box
             if not is_upvalue_box then
-                table.insert(init_constants, [[
-                    lua_newtable(L);
+                table.insert(init_constants, util.render([[
+                    luaL_newmetatable(L, $type_name);
                     lua_pushstring(L, "__metatable");
                     lua_pushboolean(L, 0);
-                    lua_settable(L, -3); ]])
+                    lua_settable(L, -3); ]], {
+                        type_name = C.string(upv.typ.name)
+                    }))
             end
         elseif tag == "coder.Constant.String" then
             table.insert(init_constants, util.render([[
--- a/src/pallene/coder.lua
+++ b/src/pallene/coder.lua
@@ -254,7 +254,7 @@ function Coder:get_stack_slot(typ, dst, slot, loc, description_fmt, ...)
         check_tag = util.render([[
             if (l_unlikely(!$test)) {
                 pallene_runtime_tag_check_error(L,
-                    $file, $line, $expected_type, pallene_type_name(L, $slot),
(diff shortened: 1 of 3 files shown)


src/pallene/coder.lua
```diff
@@ -245,17 +245,23 @@ function Coder:get_stack_slot(typ, dst, slot, loc, description_fmt, ...)
     else
         assert(not typ.is_upvalue_box)
         local extra_args = table.pack(...)
+        local expected_type = pallene_type_tag(typ)
+        if expected_type == "LUA_TUSERDATA" then
+            expected_type = C.string(typ.name)
+        else
+            expected_type = "pallene_tag_name(" .. expected_type .. ")"
```
hugomg, line 252: Do we still need `pallene_tag_name`, now that we have `pallene_type_name`?

src/pallene/pallenelib.lua
```diff
@@ -48,6 +48,7 @@ return [==[
 
 /* Type tags */
 static const char *pallene_tag_name(int raw_tag);
+static const char *pallene_type_name(lua_State *L, const TValue *raw_type);
```
hugomg, line 51: The `TValue*` points to a Lua value, not to a raw type. Please name this parameter more similarly to how we name the other TValue parameters in the neighboring functions.

src/pallene/coder.lua
```diff
@@ -245,17 +245,23 @@ function Coder:get_stack_slot(typ, dst, slot, loc, description_fmt, ...)
...
+        end
         check_tag = util.render([[
             if (l_unlikely(!$test)) {
                 pallene_runtime_tag_check_error(L,
-                    $file, $line, $expected_tag, rawtt($slot),
+                    $file, $line, $expected_type, pallene_type_name(L, $slot),
```
hugomg, line 257: Perhaps it would be more efficient to pass the `$slot` directly and let the `pallene_runtime_tag_check_error` compute the type name. That way, we don't need to generate code for the type name function call at run-time tag check.

hugomg (changes requested):
Thanks, Vipul! Looks like the PR is not passing the test suite. Can you please look into that?

src/pallene/coder.lua
Vipul-Cariappa: We would require both `pallene_tag_name` and `pallene_type_name`. They are different in two ways.
1. Difference in the function signature `const char *pallene_tag_name(int raw_tag)` and `const char *pallene_type_name(lua_State *L, const TValue *v)`
2. `expected_type` is known at the compile time (and uses `pallene_tag_name`) i.e. the argument's expected type. Whereas the argument `const TValue *v` passed to `pallene_type_name` cannot be known at compile time. `v` is the type of the actual argument passed at runtime.

hugomg: 1. IIRC, there aren't any other places that call `pallene_tag_name`. The tag name function was there only for the error messages and `pallene_type_name` now does that job. Notice how both functions are currently quite similar.

2. If your goal was that the expected_type be known at compile time, then calling `pallene_tag_name` is not the way to do that. That C function call will only happen at run-time.

3. One last thing: In the coder.lua, please use different variable names for the type tag returned by `pallene_type_tag` and for the type name that you compute inside the if statement.

Vipul-Cariappa: Please have a look at the recent changes. Hope the latest commit resolves the issues/suggestions you mentioned.

hugomg:
Hmm, why is our CI failing? 🤔
The tests are passing on my machine. Is that also the case for you, Vipul?

hugomg (approved):
Looks like it's passing now, go figure...
Thanks again!

PR #597, opened 2024-05-28
Improved stack trace in Pallene using Pallene Tracer

### LINKED ISSUE #565: Full pallene stack traces
Currently, when a Pallene function calls another Pallene function they share the same Lua stack frams. As a result, if there's a run-time type error, it will show as a single "[C]" function call. 

I imagine three ways that maybe could improve this. The first  would be to make each Pallene function call have it's own Lua stack frame. That way, the default Lua stack trace might be able to show more info.

Another option would be to encode the extra call stack info in the code we generate for function calls. Simillarly to the line number and file name parameter we pass to most internal error subroutines.

A third option would be to rely on C debuggers. Add #line directives everywhere and write a stack take handler that uses gdb to build the stack trace. This would let people avoid the overhead if they compile without -g. Also, it would get rid of the annoying file name and line number function arguments that we mentioned previously. That said relying on an external program such as gdb would add a run-time dependency, which might be undesirable...

singul4ri7y:
Initial commits on Pallene Tracer implementation in Pallene compiler for improved stack-traces using a separate debug traceback function `pallene_tracer_debug_traceback`.

The Pallene Tracer initial design specification can be found here: [The Pallene Tracer](https://docs.google.com/document/d/1JNMswjvCXlri4nYjLI5f1w3HrkHa7xSKyXSYxNG1MWs/edit?usp=sharing)

Resolves #565


src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,46 @@ return [==[
...
+#define PALLENE_GLOBAL_SETLINE(L, line) pallene_tracer_global_setline(L, line)
+#define PALLENE_SETLINE(line)           pallene_tracer_setline(&_frame, line)
+#define PALLENE_FRAMEEXIT(L, ...)       pallene_tracer_frameexit(L);                            \
+                                        return __VA_ARGS__
+
+/* Pallene Tracer related data-structures. */
+typedef struct pt_frame {
+    /* Max name length is 127. */
+    char fn_name[128];
+    char mod_name[128];
```
hugomg, line 66: 256 bytes is a lot of overhead. Would it be possible to allocate these names statically? That would also get rid of the length limit.

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,46 @@ return [==[
...
+
+    /* Line number. */
+    int line;
+
+    /* The frame signature. */
+    ptrdiff_t frame_sig;
+
+    struct pt_frame *next;
+    struct pt_frame *prev;
```
hugomg, line 75: Would a singly-linked list suffice?

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+        head = frame;
+        tail = frame;
+
+        goto out;
+    }
+
+    /* The frame signature. */
+    /* If we don't have any frame signature, that denotes the call was made from Pallene
+       environment. */
+    frame -> frame_sig = sig == (ptrdiff_t) NULL ? tail -> frame_sig : sig;
```
hugomg, line 165: I found this hard to read without parenthesis. Maybe use an if statement instead?

hugomg (commented):
I still need to looked at the debug_traceback implementation, but might only be able to do so tomorrow.

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+
+            if(!(gstack = lua_getstack(L, level, &ar)))
+                continue;
+
+            level++;
+
+            /* We need more info for a good traceback entry. */
+            lua_getinfo(L, "Slnt", &ar);
+
+            /* We have got a C frame. Time to make a context switch. */
```
hugomg, line 276: As pointed out by Lourival,  `lua_iscfunction` might be more robust than looking at the ar.what string.

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+    lua_getglobal(L, "__pallene_tracer_stack_tail");
+    pt_frame_t *frame = (pt_frame_t *) lua_topointer(L, -1);
+    lua_pop(L, 1);
+
+    /* For context switch from Pallene to Lua or vice versa. */
+    /* We use the respective call-stack depending on the context.
+     * In Lua context we use the Lua call-stack and in Pallene context
+     * we use our self-maintained call stack. */
+    /* 1: Lua, 0: Pallene */
+    int context = 1;
```
hugomg, line 242: Since we're using if-else, I feel this variable wants to be named as a boolean (e.g. is_lua)

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+
+    /* The frame signature. */
+    ptrdiff_t frame_sig = frame -> frame_sig;
+
+    while(true) {
+        /* If we are done iterating through the Lua stack and
+           there is no more frame left in Pallene Tracer stack,
+           we are done with the traceback. */
+        if(!gstack && frame == NULL)
+            break;
```
hugomg, line 262: while(gstack || frame)

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+
+        if(context) {
+            /* Get lua call stack information. */
+            lua_Debug ar;
+
+            if(!(gstack = lua_getstack(L, level, &ar)))
+                continue;
```
hugomg, line 269: Is there ever a situation where we run out of Lua stack before we run out of Pallene stack? I believe that might not be possible.

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+            /* If the frame signature does not match, then it's just a normal
+               C function  */
+            if(frame == NULL || frame -> frame_sig != frame_sig) {
+                /* If we switched from Lua and the frame signature is
+                   not known, then function is just a C function oblivious to Pallene and Lua. */
+                if(prev_context == 1)
+                    fprintf(stderr, "    C Function: 0x%lx\n", (void *) frame_sig);
+
+                context = 1;
+                goto pallene_stack_done;
```
hugomg, line 299: I don't like this goto... Looks like it's there because of the prev_context. Perhaps when we get rid of frame_sig, the prev_context will also go away at the same time?

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+    head = (pt_frame_t *) lua_topointer(L, -1);
+    lua_getglobal(L, "__pallene_tracer_stack_tail");
+    tail = (pt_frame_t *) lua_topointer(L, -1);
+    lua_pop(L, 2);
+
+    /* If there is no frame in the stack. */
+    if(head == NULL) {
+        /* Just to be safe. */
+        frame -> prev = NULL;
+        frame -> next = NULL;
```
hugomg, line 151: Code style: please use `frame->prev` throughout, without the spaces.

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
 static lua_Integer pallene_math_modf(lua_State *L, const char* file, int line, lua_Number n, lua_Number* out);
 
 /* Other builtins */
-static TString* pallene_string_char(lua_State *L, const char* file, int line, lua_Integer c);
-static TString* pallene_string_sub(lua_State *L, TString *str, lua_Integer start, lua_Integer end);
+static TString *pallene_string_char(lua_State *L, const char* file, int line, lua_Integer c);
```
srijan-paul, line 130: Were these changes intentional?

src/pallene/coder.lua
```diff
@@ -1821,6 +1842,10 @@ function Coder:generate_luaopen_function()
             #error "Lua version must be exactly 5.4.6"
             #endif
 
+            /* Initialize Pallene Tracer. */
+            /**/
+            pallene_tracer_init(L);
+            /**/
```
srijan-paul, line 1848:
```suggestion
            pallene_tracer_init(L);
```

src/pallene/pallenelib.lua
srijan-paul: Yeah, I think we can allocate them statically and store pointers here.

src/pallene/pallenelib.lua
```diff
@@ -90,12 +127,209 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+    lua_getglobal(L, "__pallene_tracer_stack_tail");
+    tail = (pt_frame_t *) lua_topointer(L, -1);
+    lua_pop(L, 2);
+
+    /* We are popping the very last frame. */
+    if(tail -> prev == NULL) {
+        tail = NULL;
+        head = NULL;
+
+        goto out;
```
srijan-paul, line 207: I think it'll be easier to follow if we avoid this `goto`.
Maybe take a cursory look at all the `goto`s we have right now and reconsider some of them?

srijan-paul (changes requested):
Honestly, the overall idea looks pretty good to me.
Some minor feedback.

src/pallene/pallenelib.lua
singul4ri7y: Yeah, it's doable.

src/pallene/pallenelib.lua
singul4ri7y: We would implement a stack-trace inspector later on if possible. May need it then.

src/pallene/pallenelib.lua
singul4ri7y: I find ternary cleaner and it's a little faster than if. I will try and make it more readable :).

src/pallene/pallenelib.lua
singul4ri7y: `context` is fine. I mentioned the the word/name in the design spec as well. It also sound cool ;)

src/pallene/pallenelib.lua
srijan-paul: Performance won't be an issue!
Compilers generate identical assembly for both with `-O2`.

image

src/pallene/pallenelib.lua
singul4ri7y: No we do not run out of Pallene frames before the Lua frames. We need the `gstack` variable for the loop :).

src/pallene/pallenelib.lua
singul4ri7y: We won't/can't get rid of the frame signature for now. Otherwise we won't be able to figure out when to switch to Lua stack especially if there is a C call that is not of Pallenes or is not using our tracer.

But I will try to get rid of the nasty `goto` :)

src/pallene/pallenelib.lua
singul4ri7y: Probably not. I used some Neovim formatter on the code, maybe those lines also got affected. Shouldn't be an issue, should it?

src/pallene/pallenelib.lua
singul4ri7y: `goto`s are extensively used as finalizers. Also used in Linux kernel as well :).

srijan-paul: Sure, I don't mind them as long as we are consistent with their use.
If you feel this usage is consistent with the code style of the rest of `pallenelib`, we can keep them.

src/pallene/coder.lua
singul4ri7y: This is to add newlines before and after `pallene_tracer_init`, so that it looks separated instead of looking blended between.

singul4ri7y:
I have committed another change which fixes/enhances to the code in matter.

One problem I think we got, is the Pallene tracer line number is not quite correct. Any insights on this? @hugomg @srijan-paul

src/pallene/pallenelib.lua
hugomg: If we need it later, we can add it then. I'd rather keep it simple for now...

src/pallene/pallenelib.lua
hugomg: I'm confused about the won't/can't. I have the impression that we can, by giving a pt_frame to the Lua entry points, but maybe you just want to do it on a future PR?

src/pallene/pallenelib.lua
singul4ri7y: Okay then. Small change, I would do it in the later commits.

src/pallene/pallenelib.lua
singul4ri7y: We could certainly have a separate type of pt_frame for Lua entry points. But the `frame signature` in our case is important when we are trying to figure out whether we have a C sitting in the middle which Pallene tracer have no idea about. The only way we can figure this out is in my opinion is if we have some sort of signature or essence of the Previous Lua entry point, which is common in both Lua and Pallene stack. Hope that make sense.

Alternatively, you can elaborate your idea even more, that way I will be able to figure out should the frame signature be needed.

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,73 @@ return [==[
...
+} pt_fn_details_t;
+
+/* A single frame representation. */
+typedef struct pt_frame {
+    frame_type_t type;
+    int line;
+
+    union {
+            const pt_fn_details_t *details;
+            const void *frame_sig;
```
hugomg, line 99: Can frame_sig be a lua_Cfunction instead of void*?

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,73 @@ return [==[
...
 #include <math.h>
+#include <stdbool.h>
 
 #define PALLENE_UNREACHABLE __builtin_unreachable()
 
+/* Part of Pallene Tracer. */
+#define PALLENE_C_FRAMEENTER(L, name)   pt_fn_details_t _details = {                             \
+                                            .fn_name  = name,                                    \
+                                            .mod_name = PALLENE_SOURCE_FILE                      \
+                                        };                                                       \
```
hugomg, line 54: Make sure that the _details are statically allocated. `static pt_fn_details_t _details` ought to do the trick.

hugomg, line 54: This code is indented too much to the right; it's hard to read in the PR window.

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,73 @@ return [==[
...
+
+#define PALLENE_LUA_FRAMEENTER(L, sig)  pt_frame_t _frame = {                                    \
+                                            .type = PALLENE_TRACER_FRAME_TYPE_LUA,               \
+                                            .shared = {                                          \
+                                                .frame_sig = sig                                 \
+                                            }                                                    \
+                                        };                                                       \
+                                        pallene_tracer_frameenter(L, &_frame)
+
+#define PALLENE_GLOBAL_SETLINE(L, line) pallene_tracer_global_setline(L, line)
```
hugomg, line 71: I worry that a GLOBAL_SETLINE function might not be compatible with alternative ways to reference the stack. Perhaps it would be more appropriate to call PALLENE_SETLINE before calling the various "pallene_error" functions? (As opposed to what we do now, where the pallene_error calls global_setline)

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,73 @@ return [==[
...
+#define PALLENE_SETLINE(line)           pallene_tracer_setline(&_frame, line)
+#define PALLENE_FRAMEEXIT(L, ...)       pallene_tracer_frameexit(L);                             \
+                                        return __VA_ARGS__
```
hugomg, line 74: Is there a particular reason why the `return` needs to be inside the macro?

src/pallene/pallenelib.lua
```diff
@@ -43,9 +43,73 @@ return [==[
...
+typedef enum frame_type {
+    PALLENE_TRACER_FRAME_TYPE_C,
+    PALLENE_TRACER_FRAME_TYPE_LUA
+} frame_type_t;
+
+/* Details of a single function such as what is the name
+   and where it is from. */
+typedef struct pt_fn_details {
+    const char *const fn_name;
+    const char *const mod_name;
```
hugomg, line 89: Does the second `const` matter?

src/pallene/pallenelib.lua
```diff
@@ -90,12 +154,267 @@ static lua_Number  pallene_math_log(lua_Integer x, lua_Integer base);
...
+    lua_pushlightuserdata(L, stack);
+    lua_setglobal(L, "__pallene_tracer_stack");
+}
+
+static int pallene_tracer_debug_traceback(lua_State *L) {
+    const char *message = lua_tostring(L, 1);
+    fprintf(stderr, "Runtime error: %s\nStack traceback: \n", message);
+
+    /* Lua: 1, Pallene: 0 */
+    int context = 1;
```
hugomg, line 295: I wonder if we can structure our traversal as two nested loops and get rid of the context flag and some of the gotos.

    for each frame in the Lua stack
        if this frame is linked to the Pallene stack
            for each frame in the Pallene stack block
                 print C frame
        else
            print Lua frame

src/pallene/pallenelib.lua
singul4ri7y: Yes, switched to `lua_CFunction`.

src/pallene/pallenelib.lua
singul4ri7y: Removed `PALLENE_GLOBAL_SETLINE` and did workaround by calling `PALLENE_SETLINE` before calling any "pallene_error".

src/pallene/pallenelib.lua
singul4ri7y: It just made writing code a little bit cleaner. Instead of writing `PALLENE_FRAMEEXIT` and the `return` statement stacked up like sandwich, only writing `PALLENE_FRAMEEXIT(L, return_value)` would do the trick.

But removed it since it was causing problems with the Test Suite.

src/pallene/pallenelib.lua
singul4ri7y: Doable.

src/pallene/pallenelib.lua
singul4ri7y: Good programming practice. If something shouldn't be changed, make it constant in every possible way.
So, no oopsie-doopsie change :)

The second `const` is to make the pointer property constant, which is more important in this case.

src/pallene/pallenelib.lua
singul4ri7y: Moved it in the left, should be Okay now.

singul4ri7y: But why tho?

hugomg: Should be more efficient. Only one copy of the _details struct, instead of making a new one for every stack frame.

hugomg (approved):

PR #605, opened 2024-06-18
Update ast.lua(added Switch statement and Try-Catch block)

nnlgsakib:


hugomg:
Sorry, but I don't understand what's the purpose of this... Modifying the AST without modifying the parser and code generator to support the new feature?

Also, Pallene usually tries to stick to existing Lua syntax when possible.

PR #606, opened 2024-06-19
Migrate to Heap-allocated Call Stack and Performance Improvements

### LINKED ISSUE #602: Stacktrace: can we avoid xpcall?
Should we replace the default traceback in our patched Lua version?

### LINKED ISSUE #604: Stacktrace: Reset Pallene stack when done printing tracebacks to avoid stack corruption after `pcall`
As the situation stands, we do not clear our Pallene stack after printing the traceback. It is Okay in general cases when the interpreter exits, we don't have to care.

But the situation is different, when we use `pcall` to catch the error. The interpreter continues and the stacks tend to stay where they are and gets corrupted when we try to reuse the stack for later calls.

We need to find a way to reset the stack after we are done doing the traceback.

singul4ri7y:
Change log: 
1. Migrated to Heap-allocated stack where the allocation is completely handled by Pallene (using `malloc` and `free`, because we cannot use regular userdata here).
2. Added `pallene-debug` script which will allow one to debug Pallene modules without making any changes to the Lua script.
3. Added finalizers and other optimizations to yield maximum performance which fixes a critical bug.

Resolves #604


src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env lua
+
+if #arg == 0 then
+    io.stderr:write("pallene-debug: need a file to debug. Abort!\n")
+    os.exit(1)
```
hugomg, line 5: Consider using argparse. Gives us --help for free, standard "usage" message, etc. See `pallenec` main script.

src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
...
+if not content then
+    io.stderr:write("pallene-debug: could not read file '"..filename.."'. Abort!\n")
+    os.exit(1)
+end
+-- We no longer need the file.
+file:close()
+
+-- Silently run the code, loading the Pallene/Pallene Tracer compatible
+-- modules so that we can figure out whether we have Pallene Tracer or not.
+local _, _ = pcall(load(content))
```
hugomg, line 33: Can we use `loadfile` instead?

src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
...
+end
+
+-- Inject the wrapper code.
+content = [[
+function __pallene_debug_wrapper()
+]]..content..[[
+end
+
+xpcall(__pallene_debug_wrapper, pallene_tracer_debug_traceback)
+]]
```
hugomg, line 48: Why not run `xpcall` in pallene-debug itself, passing the function we got from load/loadfile?

src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
...
+
+load(content, filename)()
```
hugomg, line 50: I suspect that we should pass `table.unpack(arg)` as arguments, in case the script we're loading uses `...`.

hugomg (changes requested):

src/bin/pallene-debug
singul4ri7y: Would you please elaborate?

hugomg: In addition to the global `arg` table, Lua scripts can also access argv via `...` vararg parameters.

src/bin/pallene-debug
singul4ri7y: Done.

src/bin/pallene-debug
singul4ri7y: Sure. Done.

src/bin/pallene-debug
singul4ri7y: Muscle memory was trained seeing xpcall inside of all the scripts :). Moved it outside tho.

src/bin/pallene-debug
singul4ri7y: Done!

src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
...
+    os.exit(1)
+end
+
+-- Add a wrapper function.
+local function wrapper()
+    fn(table.unpack(opts.args))
+end
+
+-- Moment of truth.
+xpcall(wrapper, pallene_tracer_debug_traceback)
```
hugomg, line 50: I think we do not need a wrrapper.

    xpcall(fn, traceback, table.unpack(opts.args))

src/bin/pallene-debug
```diff
@@ -0,0 +1,50 @@
...
+-- Supress I/O.
+local original_print = print
+local original_io_write = io.write
+
+print    = function() end
+io.write = function() end
+
+-- Run the script initially so that we can figure out whether
+-- we have Pallene modules that is compiled with `--use-traceback` flag.
+local _, _ = pcall(fn)
```
hugomg, line 32: Running the script twice sounds sketchy. It can have other side effects other than printing to screen, and it will take twice as long.

I suppose it would make more sense for the pallene-debug script to supply the traceback function, instead of getting it from a global variable in the loaded script.

Maybe what we should do is create a standalone C module that exports the Pallene traceback and install it together with pallenec. Then, in this script we can require the traceback function.

hugomg (changes requested):
I have some bigger questions about the pallne-debug.

The C bits look good at a first glance but I still need to take a closer look at them.

src/bin/pallene-debug
singul4ri7y: Say the Lua script didn't use any Pallene module with traceback. What then? If we supply our own traceback function it might segfault because there is no stack to begin with. That would be risky. We need to make sure somehow that a Pallene module with traceback exists.

hugomg: Could the traceback function check the Lua registry to see if the Pallene stack exists?

singul4ri7y: It could.

Maybe it's a good idea to detach the traceback fn once and for all. Because we are using it explicitly anyway. But when calling with `xpcall`, can the traceback fn return anything? Maybe we can introduce another fn for that?

hugomg: I don't understand the "when calling with xpcall" question.

singul4ri7y: The traceback fn should return anything to denote there is no stack to work on. This way we would be able to warn the users to use the `--use-traceback` flag. It would be better than not printing any traceback when there is no Pallene stack at all.

My question was if we pass our traceback to `xpcall` as second argument can we fetch what it returned? Or we have to introduce another fn for it?

singul4ri7y: On the contrary, I have a grosshack (sorta) prepared. Though we have to run Lua scripts twice, it won't be slow. Also it shouldn't have side effects.

Creating a separate module, exporting traceback fn is way to hassle for a single debug script.

hugomg (approved):

PR #607, opened 2024-06-29
Basic-blocks-based intermediate language and flow analysis.

gabrielsferre:
Changing the previous intermediate language, which was a tree, to a new one made out of a graph of basic blocks. This change is being made so it's easier to implement future optimizations for the generated code using standard flow analysis techniques. Here, we're already using flow analysis on the blocks to identify uninitialized variables.


src/pallene/coder.lua
```diff
@@ -1648,130 +1670,91 @@ end
...
 gen_cmd["CheckGC"] = function(self, cmd, func)
     return util.render([[ luaC_condGC(L, ${update_stack_top}, (void)0); ]], {
         update_stack_top = self:update_stack_top(func, cmd) })
 end
 
+function Coder:generate_blocks(func)
+    local out = ""
+    for block_id,block in ipairs(func.blocks) do
+        -- putting "(void)0" at the label so the C compiler doesn't complain about dangling labels
+        local content = self:c_label(block_id) .. ":(void)0;\n"
```
hugomg, line 1746: It looks like we're adding (void)0 to all labels. Is there a way to only add them when necessary?

src/pallene/coder.lua
```diff
@@ -1648,130 +1670,91 @@ end
...
+        for _,cmd in ipairs(block.cmds) do
+            if cmd._tag ~= "ir.Cmd.Jmp" or cmd.target ~= block_id + 1 then
+                local cmd_str = self:generate_cmd(func, cmd) .. "\n"
+                content = content .. cmd_str .. "\n"
```
hugomg, line 1750: Repeat concatenation is O(n^2). Please use table.concat.

src/pallene/ir.lua
```diff
@@ -51,7 +51,8 @@ function ir.Function(loc, name, typ)
         captured_vars = {},   -- list of ir.VarDecl
         f_id_of_upvalue = {}, -- { u_id => integer }
         f_id_of_local = {},   -- { v_id => integer }
-        body = false,         -- ir.Cmd
+        blocks = {},          -- { ir.BasicBlock }
+        ret_vars = {},        -- { v_id }, list of return variables
```
hugomg, line 55: I think the argument vars are specified implicitly, but here we're specifying the return vars explicitly. Is this intentional? It is better to be implicit or explicit?

src/pallene/ir.lua
```diff
@@ -186,15 +194,14 @@ local ir_cmd_constructors = {
...
-    Return  = {"loc", "srcs"},
-    Break   = {},
-    Loop    = {"body"},
+    IterFor    = {"loc", "dst_i", "dst_cond", "dst_iter", "dst_count",
+                  "src_start", "src_limit", "src_step"},
 
-    If      = {"loc", "src_condition", "then_", "else_"},
-    For     = {"loc", "dst", "src_start", "src_limit", "src_step", "body"},
+    Jmp        = {"target"},
+    JmpIfFalse = {"loc", "src_cond", "target"},
```
hugomg, line 204: Why does one have a loc and the other not?

src/pallene/coder.lua
```diff
@@ -1648,130 +1670,91 @@ end
...
         start = self:c_value(cmd.src_start),
         limit = self:c_value(cmd.src_limit),
         step  = self:c_value(cmd.src_step),
-        body  = self:generate_cmd(func, cmd.body)
     }))
 end
 
+gen_cmd["Jmp"] = function(self, cmd, _func)
+    local jmp = "goto " .. self:c_label(cmd.target) .. ";"
+    return jmp
```
hugomg, line 1726: Can return the string directly, without `local jmp`

src/pallene/ir.lua
```diff
@@ -258,136 +265,117 @@ function ir.get_dsts(cmd)
...
-            cmd.else_ = go(cmd.else_)
-        elseif tag == "ir.Cmd.Loop" then
-            cmd.body = go(cmd.body)
-        elseif tag == "ir.Cmd.For" then
-            cmd.body = go(cmd.body)
-        else
-            -- no child nodes
+            if jmp_false then
+                depth_search(jmp_false.target)
+            end
```
hugomg, line 325: These get_jump funcitons feel strange. Can we avoid them by giving a uniform interface to the jump instructions? E.g. `block[#block].target`

src/pallene/ir.lua
```diff
@@ -258,136 +265,117 @@ function ir.get_dsts(cmd)
...
+    for cmd in ir.iter(block_list) do
+        table.insert(res, cmd)
+    end
+    return res
+end
 
-    else
-        return cmd
+-- Remove jumps that are never taken
+function ir.clean(func)
```
hugomg, line 363: Future work: implement a better clean function such as the one described in Cooper&Torczon book. (please open an issue so we don't forget)

src/pallene/to_ir.lua
```diff
@@ -365,7 +499,13 @@ function ToIR:convert_stat(cmds, stat)
...
             e1.exp.var._def.id == "ipairs")
 
-
+        bb:enter_loop()
+        local loop_begin
+        local loop_test_jmpIfFalse
+        local loop_test_jmp
+        local after_loop_test
+        local loop_end_jmp
+        local after_loop
```
hugomg, line 508: não entendi direito todas essas variáveis. Acho que precisa de um nome mais óbvio e/ou padronizado e/ou comentários

src/pallene/uninitialized.lua
```diff
@@ -3,149 +3,102 @@
 -- Please refer to the LICENSE and AUTHORS files for details
 -- SPDX-License-Identifier: MIT
 
+-- In this module we detect when variables are used before being initialized and when control flows
+-- to the end of a non-void function without returning. The analysis is fundamentally a dataflow
+-- one, but we don't check for convergence (this set of properties seem to converge in one pass) and
+-- we don't merge the properties of one basic block with one that it jumps to if the "jumped to"
+-- block has a lower id, that is, we don't consider backwards jumps (that's how we deal with loops
+-- for now).  Make sure that you call ir.clean first, so that it does the right thing in the
+-- presence of `while true` loops.
```
hugomg, line 12: this comment needs to be updated (e.g. now we do check for convergence and the dataflow is as explicit as it gets)

src/pallene/uninitialized.lua
```diff
@@ -156,31 +109,50 @@ function uninitialized.verify_variables(module)
...
+        local entry_uninit = uninit_sets[1]
+        for v_i = nargs+1, nvars do
+            entry_uninit[v_i] = true
+        end
+
+        -- solve flow equations
+        flow_analysis(func.blocks, uninit_sets, kill_sets)
+
+        -- check for errors
+        local check = coroutine.wrap(function()
```
hugomg, line 130: We used to need the coroutine cleverness because the analysis was a recursive traversal. Is this still needed?

src/pallene/uninitialized.lua
```diff
@@ -3,149 +3,102 @@
...
-                    uninit[val.id] = nil
-                end
-                check_use(val.id)
-            end
+    repeat
+        local changed = false
+        for _,block_i in ipairs(block_order) do
+            changed = block_analysis(block_i) or changed
         end
+    until not changed
```
hugomg, line 82: Future work: consider a worklist approach. (we should discuss this next time we talk). Please open an issue so we don't forget.

hugomg (changes requested):

src/pallene/ir.lua
```diff
@@ -258,136 +262,95 @@ function ir.get_dsts(cmd)
...
+    }
+end
 
-    local function go(cmd)
-        coroutine.yield(cmd)
+function ir.BasicBlock()
+    return {
+        cmds = {},           -- list of ir.Cmd
+        next = false,        -- index of next basic block if no conditional jump is taken
+        jmp_false = false,   -- JmpIfFalse
```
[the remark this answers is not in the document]
gabrielsferre: Já tirei o jmp_false e o next

src/pallene/ir.lua
```diff
@@ -258,136 +262,95 @@ function ir.get_dsts(cmd)
...
-        else
-            -- no child nodes
         end
+    end
+    return coroutine.wrap(function() go() end)
+end
 
-        -- Transform parent node
-        return f(cmd) or cmd
+function ir.flatten_cmd(block_list)
```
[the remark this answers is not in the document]
gabrielsferre: It can be removed now. I don't think the gc needs it, it uses the Cmd object itself as the table's key, not the index on flatten_cmd.

src/pallene/coder.lua
```diff
@@ -725,7 +747,7 @@ function Coder:init_upvalues()
 
     -- String Literals
     for _, func in ipairs(self.module.functions) do
-        for cmd in ir.iter(func.body) do
+        for cmd in ir.iter(func.blocks) do
```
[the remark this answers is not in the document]
gabrielsferre: I just copied what was already there. I'll try to get rid of the ir.iter and ir.flatten_cmd functions.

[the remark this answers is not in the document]
hugomg: In theory the iter is collapsing two for loops into one but I think writing the two for loops could be more obvious

src/pallene/ir.lua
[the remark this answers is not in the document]
hugomg: I am also looking forward to getting rid of commands as table keys :) see https://github.com/pallene-lang/pallene/pull/575

In a basic block world, the GC analysis should work with basic blocks indices aa the key.

src/pallene/coder.lua
```diff
@@ -1648,130 +1670,91 @@ end
...
-                ${then_}
-            }
-        ]]
-    elseif (not A) and B then
-        tmpl = [[
-            if (!$condition) {
-                ${else_}
-            }
-        ]]
+gen_cmd["InitFor"] = function(self, cmd, func)
```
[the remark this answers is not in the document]
gabrielsferre: I'll change it

src/pallene/coder.lua
```diff
@@ -1648,130 +1670,91 @@ end
...
+        start = self:c_value(cmd.src_start),
+        limit = self:c_value(cmd.src_limit),
+        step  = self:c_value(cmd.src_step),
     }))
 end
 
-gen_cmd["For"] = function(self, cmd, func)
-    local typ = func.vars[cmd.dst].typ
+gen_cmd["IterFor"] = function(self, cmd, func)
+    local typ = func.vars[cmd.dst_i].typ
```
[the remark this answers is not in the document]
gabrielsferre: Makes sense. I'll switch it to ForStep

src/pallene/coder.lua
gabrielsferre: I'll investigate this dangling label error further. Maybe if we only insert it at the last block it will work still.

src/pallene/coder.lua
gabrielsferre: Ok

src/pallene/ir.lua
gabrielsferre: It's intentional. The argument vars are supposed to always be the first on the vars list. We could establish further that the next vars are the return vars, but I find it easier to just have a list with the return vars ids. It's more straight forward to access them and we don't have to fear breaking things if changing the order in which variables are inserted on the function.

src/pallene/ir.lua
gabrielsferre: Jmp used to have a loc too, but stripping it off made the i.r. generation simpler. It was easier to add an indirect jmp without having to reference any a.s.t. node in particular. JmpIfFalse needs a loc because of "src_cond". If there's an error related to the source (e.g. it not being initialized), we need it's loc.

src/pallene/ir.lua
gabrielsferre: This would be a problem for the last block, it has no commands. The cleaner solution I could find for avoiding  the get_jump was to not have jump instructions and let the jumps be fields on the BasicBlock struct. In that case we would have to do this thing of putting the boolean source of the conditional jump as a special command at the end of the block, but then it would be a more uniform format because this special command will appear if and only if the block has a conditional jump.

src/pallene/ir.lua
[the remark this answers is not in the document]
gabrielsferre: I'm working on changing the gc to use flow analysis. Should we wait to make this change of "get rid of cmds as keys" for another PR?

src/pallene/coder.lua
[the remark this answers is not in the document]
gabrielsferre: Me too, I'll change that.

src/pallene/ir.lua
gabrielsferre: #608

src/pallene/to_ir.lua
gabrielsferre: Vou tentar pensar em nomes melhores, mas acho que vou acabar botando uns comentários mesmo.

src/pallene/uninitialized.lua
gabrielsferre: You're right. I'll update it.

src/pallene/uninitialized.lua
gabrielsferre: No, I'll change it to make a simple loop.

src/pallene/uninitialized.lua
gabrielsferre: #609

src/pallene/ir.lua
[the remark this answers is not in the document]
hugomg: It's ok to leave it for next PR (but in that case, please open an issue so we do not forget)

src/pallene/coder.lua
```diff
@@ -1648,130 +1673,94 @@ end
...
+        for _,cmd in ipairs(block.cmds) do
+            if cmd._tag ~= "ir.Cmd.Jmp" or cmd.target ~= block_id + 1 then
+                local cmd_str = self:generate_cmd(func, cmd) .. "\n"
+                table.insert(out, cmd_str)
+            end
+        end
+    end
+    return table.concat(out)
+end
+
```
hugomg, line 1763: The `B:(void)0` looks weirdly indented in the generated C code, because our indenter does not recognize it as a label. What about add a line break before the return, and include a return statement for the void case too?

Non-void:

```
B3:
  return x1;
```

void:

```
B3:
  return;
```

hugomg (changes requested):
One last thing...

src/pallene/coder.lua
gabrielsferre: done!

src/pallene/ir.lua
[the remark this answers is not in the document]
gabrielsferre: #610

hugomg (approved):

PR #611, opened 2024-07-09
coder.lua: omit "goto n+1" from JumpIf

hugomg:
De-clutter the generated C code a little bit, by not printing jumps when they jump right into the next block. Minor style tweaks as well.

### DIFF
--- a/src/pallene/coder.lua
+++ b/src/pallene/coder.lua
@@ -513,12 +513,15 @@ function Coder:pallene_entry_point_definition(f_id)
         table.insert(returns,
             util.render([[ *$reti = $v; ]], { reti = self:c_ret_var(i), v = var }))
     end
+
     local ret_mult = table.concat(returns, "\n")
-    local var1 = ""
+
+    local ret_stat
     if #func.ret_vars > 0 then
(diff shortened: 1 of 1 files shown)


src/pallene/coder.lua
```diff
@@ -513,12 +513,15 @@ function Coder:pallene_entry_point_definition(f_id)
...
     local ret_mult = table.concat(returns, "\n")
-    local var1 = ""
+
+    local ret_stat
     if #func.ret_vars > 0 then
-        var1 = " " .. self:c_var(func.ret_vars[1])
+        ret_stat = "return " .. self:c_var(func.ret_vars[1]) .. ";"
+    else
+        ret_stat = "return;"
     end
```
srijan-paul, line 524: The `ret_stat` change doesn't actually change the generated C code, right?
(Or am I understanding it wrong?)

srijan-paul (approved):

src/pallene/coder.lua
hugomg: Yes, this part does not change the generated C code.
(It's just that I don't like if without else)

PR #614, opened 2024-07-09
Adaptation of `pallene-tracer` directly in Pallene

singul4ri7y:
Resovles #602


run-tests
```diff
@@ -19,7 +19,7 @@ FLAGS=(--verbose --no-keep-going)
 
 # To speed things up, we tell the C compiler to skip optimizations. (It's OK, the CI still uses -O2)
 # Also, add some compiler flags to verify standard compliance.
-export CFLAGS='-O0 -std=c99 -Wall -Werror -Wundef -Wpedantic -Wno-unused'
+export CFLAGS='-O0 -std=c99 -Wall -Werror -Wundef -Wno-unused'
```
hugomg, line 22: Are you sure we want to get rid of -Wpedantic? If I remember correctly, it's not just for C89. In this case it would know about c99 and warn about things that are not c99.

src/pallene/c_compiler.lua
```diff
@@ -18,6 +18,7 @@ local c_compiler = {}
 
 local CC       = os.getenv("CC")       or "cc"
 local CFLAGS   = os.getenv("CFLAGS")   or "-O2"
+local SHLIBLOC = os.getenv("SHLIBLOC") or "/usr/local/lib"
```
hugomg, line 21: Pallene is installed via Luarocks, which checks that ptracer is installed. If Luarocks only checks default installation locations, then there would be nothing to gain by adding this configuration flag. And if we do need a configure flag, we should follow whatever scheme Luarocks uses (presumably it would use a ptracer-specific variables instead of a global sharedlib variable).

I think that for now we can just leave this out and assume people will install to default locations. We can add configuration later if it comes up.

hugomg (changes requested):
We also need to edit the README to tell our users how to install the ptracer dependency.

run-tests
hugomg: From the GCC manual:

> -std=

> Determine the language standard.   This option is currently only supported when compiling C or C++.
> The compiler can accept several base standards, such as  c90  or  c++98,  and  GNU  dialects  of  those
           standards,  such  as  gnu90  or  gnu++98.   When a base standard is specified, the compiler accepts all
           programs following that standard plus those using GNU  extensions  that  do  not  contradict  it.   For
           example,  -std=c90  turns  off  certain features of GCC that are incompatible with ISO C90, such as the
           "asm" and "typeof" keywords, but not other GNU extensions that do not have a meaning in ISO  C90,  such
           as  omitting  the middle term of a "?:" expression. On the other hand, when a GNU dialect of a standard
           is specified, all features supported by the compiler are enabled, even when those features  change  the
           meaning  of  the  base  standard.   As  a result, some strict-conforming programs may be rejected.  **The
           particular standard is used by -Wpedantic to identify which features  are  GNU  extensions  given  that
           version  of  the  standard.  For example -std=gnu90 -Wpedantic warns about C++ style // comments, while
           -std=gnu99 -Wpedantic does not.**

singul4ri7y: **From Wikipedia:**

ANSI C, ISO C, and Standard C are successive standards for the [C programming language](https://en.m.wikipedia.org/wiki/C_(programming_language)) published by the [American National Standards Institute](https://en.m.wikipedia.org/wiki/American_National_Standards_Institute) (ANSI) and [ISO/IEC JTC 1/SC 22](https://en.m.wikipedia.org/wiki/ISO/IEC_JTC_1/SC_22)/WG 14 of the [International Organization for Standardization](https://en.m.wikipedia.org/wiki/International_Organization_for_Standardization) (ISO) and the [International Electrotechnical Commission](https://en.m.wikipedia.org/wiki/International_Electrotechnical_Commission) (IEC). Historically, the names referred specifically to the original and best-supported version of the standard (known as C89 or C90). Software developers writing in C are encouraged to conform to the standards, as doing so helps [portability](https://en.m.wikipedia.org/wiki/Porting) between compilers.

**This is from GCC flags manual:**

-Wpedantic
-pedantic
Issue all the warnings demanded by strict ISO C and ISO C++; diagnose all programs that use forbidden extensions, and some other programs that do not follow ISO C and ISO C++. This follows the version of the ISO C or C++ standard specified by any -std option used.

Actually to be honest I am confused right now. From my experience, while using `-Wpedantic` with `-std=c11` I was getting warnings which was of `C89` standard as far as I can remember.

singul4ri7y: I am bringing back `-Wpedantic` flag.

Here is where I was wrong: ISO C used to be historically and strictly referred to `C89`. But now it is not. ISO C standard in this case is the standard you specify using `-std=`.

src/pallene/pallenelib.lua
```diff
@@ -46,6 +46,10 @@ return [==[
 #include <stdbool.h>
 #include <stdlib.h>
 
+/* Pallene Tracer for tracebacks (dynamically linked). */
+/* Look at https://github.com/pallene-lang/pallene-tracer for more info. */
```
srijan-paul, line 50: This repo is still empty, by the way.
Is that intentional?

srijan-paul (changes requested):
LGTM!
I just want to install it locally once.
Can the pallene tracer repo be updated before we merge this?

src/pallene/pallenelib.lua
singul4ri7y: Switch to the `dev` branch. And also please review the PR which intends to merge the `dev` branch to `main`.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,13 @@ jobs:
                   make
                   sudo make install
 
+            - name: Install Pallene Tracer
+              run: |
+                  git clone https://github.com/pallene-lang/pallene-tracer
```
hugomg, line 82: We should create version tags in that repo, and import a particular version. Otherwise, Pallene builds might break if there new commits over there.

src/pallene/c_compiler.lua
```diff
@@ -61,6 +61,8 @@ function c_compiler.compile_o_to_so(in_filename, out_filename)
         CFLAGS_SHARED,
         "-o", util.shell_quote(out_filename),
         util.shell_quote(in_filename),
+        "-lptracer",
+        "-Wl,-rpath=/usr/local/lib",
```
hugomg, line 65: Do we really need rpath? I'm worried about hardcoding user/local/lib

hugomg (changes requested):

src/pallene/c_compiler.lua
singul4ri7y: Yes.

.github/workflows/ci.yml
singul4ri7y: Let's do that when Pallene Tracer is stabilized.

src/pallene/c_compiler.lua
hugomg: Why yes? In my system, "/usr/local/lib" is part of the default search path. And presumably the user should have installed ptracer somewhere in their default search path. Also, that directory is configurable in the ptracer makefile, so we should not be hardcoding it here.

singul4ri7y: Removed the hardcoded part. I don't know whats up with the distro you are using, from my experience, you need some -rpath to defined. And here, my experience includes, Manjaro, Debian 10 and later and ArchLinux (which I use BTW). It does not work for every distro, I can assure you that much.

hugomg: Rpath should only be necessary if you're installing the shared library in a non-standard location, and don't want to use LD_LIBRARY_PATH.

You can check the search path in your distro by running  `ld --verbose | grep SEARCH_DIR | sed -e 's/; /\n/g'`. On mine, it includes /usr/local/lib

singul4ri7y: It includes `/usr/local/lib` in my case as well, but still fails to link `libptracer.so` :)

hugomg: Did you try removing the entire "-Wl,-rpath=/usr/local/lib" line?

singul4ri7y: Yes

singul4ri7y: For your convenience, this get this error while running the test suite after removing the line.

```
Failure → ./spec/execution_tests.lua @ 103
#c_backend / Exported functions / exports public functions
./spec/execution_tests.lua:89: command failed: lua '__test_c__script.lua' > '__test_c__output.txt'

stack traceback:
	./spec/execution_tests.lua:89: in upvalue 'run_test'
	./spec/execution_tests.lua:104: in function <./spec/execution_tests.lua:103>

lua: error loading module '__test_c__' from file './__test_c__.so':
	libptracer.so: cannot open shared object file: No such file or directory
stack traceback:
	[C]: in ?
	[C]: in function 'require'
	__test_c__script.lua:1: in main chunk
	[C]: in ?
●✱
1 success / 0 failures / 1 error / 0 pending : 0.211186 seconds
```

src/pallene/pallenec.lua
```diff
@@ -34,7 +34,8 @@ do
     )
 
     -- No Pallene tracebacks
-    p:flag("--use-traceback", "Use function traceback for debugging")
+    p:flag("--use-traceback",    "Use function traceback for debugging using Pallene Tracer")
```
srijan-paul, line 37: We should probably use user-friendly messages and omit mentions of implementation details (e.g: use of Pallene tracer).

```suggestion
    p:flag("--use-traceback",    "Show a call-stack trace on error")
```

srijan-paul (dismissed):
looking good.
Just a few minor changes.

src/pallene/pallenec.lua
singul4ri7y: I think it's fine. In Pallene we are just accumulating the traces, which is then shown by `pallene-tracer` script or `xpcall` with `pallene_tracer_debug_traceback`. Mention of Pallene Tracer might spark interest in peoples mind :)

hugomg (approved):
I still have reservations about the rpath bit. But we can address that in a future PR.

singul4ri7y:
@hugomg Agree on the rpath part. But if distros like Arch Linux don't resolve the /usr/local/lib directory by default, it's hard to say most distros will, especially distros like Manjaro, EndevourOS or SteamOS for that matter (Linux gaming is a thing btw), derived directly from ArchLinux. I would like to keep it for now, until we find a better alternative, which we should.

Issue #610, opened 2024-07-08
Stop using commands as keys for tables on the garbage collector


gabrielsferre:
We do liveliness analysis to check at each command of the intermediate representation which variables should be saved at the Lua stack before control transfers. Currently, we keep track of live variables per command using a table whose keys are the commands themselves, that is, the ir.Cmd objects: https://github.com/pallene-lang/pallene/blob/57c183789b2b1e0a657a96ce4383c6e6bc5b9e7b/src/pallene/gc.lua#L69-#L88

This approach for indexing the table comes with some aliasing problems (q.v. #575), so we want to replace the table's keys with something else.

hugomg:
In theory, we should be able to index by basic block, like the other analysis.

hugomg:
Done in #615.

@gabrielsferre protip: if you say "closes #615" in the commit message, then github closes the issue automatically

Issue #609, opened 2024-07-04
Use a work list when doing flow analysis


gabrielsferre:
When solving data-flow equations for flow analysis, we should use a *work list* approach instead of using a flag like we're doing now at __src/pallene/uninitialized.lua__: https://github.com/pallene-lang/pallene/blob/dceb5593a8001c4511a7ea81b97f176fe33b9057/src/pallene/uninitialized.lua#L82

PR #615, opened 2024-07-21
Flow analysis on garbage collector and more

gabrielsferre:
* Using flow analysis for checking liveness of variables at the garbage collector.
* Using work-list for checking flow analysis convergence (both in gc.lua and  uninitialized.lua). See #609.
* Not using commands as keys for tables anymore. See #610.

### DIFF
--- a/src/pallene/gc.lua
+++ b/src/pallene/gc.lua
@@ -35,6 +35,15 @@ local tagged_union = require "pallene.tagged_union"
 
 local gc = {}
 
+local function FlowState()
+    return {
+        input  = {},  -- {var_id -> bool?} live variables at block start
+        output = {},  -- {var_id -> bool?} live variables at block end
+        kill   = {},  -- {var_id -> bool?} variables that are killed inside block
+        gen    = {},  -- {var_id -> bool?} variables that become live inside block
+    }
+end
+
 function gc.cmd_uses_gc(tag)
     assert(tagged_union.typename(tag) == "ir.Cmd")
     return tag == "ir.Cmd.CallStatic" or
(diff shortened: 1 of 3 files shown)


src/pallene/gc.lua
```diff
@@ -28,77 +29,234 @@ local types = require "pallene.types"
 --   2) Identify functions that don't call the GC (directly or indirectly) and don't treat calls to
 --      them as potential GC sites. (Function inlining might mitigate this for small functions)
 --
---   3) Use a flow-based liveliness analysis to precisely identify the commands that a variable
---      appears live at, instead of approximating with first definition and last use.
---
---   4) Use SSA form or some form of reaching definitions analysis so that we we only need to mirror
+--   3) Use SSA form or some form of reaching definitions analysis so that we we only need to mirror
```
hugomg, line 32: Perhaps investigating the items on this list could be an interesting experiment for the project... We could do a subsequent PR for that and then compare how the benchmarks fare, before and after.

src/pallene/gc.lua
```diff
@@ -28,77 +29,234 @@ local types = require "pallene.types"
...
 
-function gc.compute_stack_slots(func)
+local function FlowState()
+    return {
+        input  = {},  -- {var_id -> bool?} live variables at block start
+        output = {},  -- {var_id -> bool?} live variables at block end
+        kill   = {},  -- {var_id -> bool?} variables that are killed inside block
+        gen    = {},  -- {var_id -> bool?} variables that become live inside block
+    }
+end
```
hugomg, line 45: Another question for a future PR:
1. How similar is our analysis code to the uninitialized analysis
2. Is it worth abstracting some of it to a separate flow analysis file?

src/pallene/gc.lua
```diff
@@ -28,77 +29,234 @@ local types = require "pallene.types"
...
+            in_changed = in_changed or (previous_val ~= new_val)
         end
-        for _, v_id in ipairs(ir.get_dsts(cmd)) do
-            local typ = func.vars[v_id].typ
-            if types.is_gc(typ) and not first_definition[v_id] then
-                first_definition[v_id] = i
-                table.insert(defined_variables, v_id)
+
+        for v, g in pairs(gen) do
+            assert(g ~= true or g ~= kill[v], "gen and kill can't both be true")
```
hugomg, line 82: Could it be `assert(not (g and kill[v]))`?
Checking against `==true` is too specific.

src/pallene/gc.lua
```diff
@@ -28,77 +29,234 @@ local types = require "pallene.types"
...
-                if a and b and a < i and i <= b then
-                    table.insert(live_gc_vars[cmd], v_id)
-                end
+    local function merge_live(input, output)
+        for v, _ in pairs(input) do
+            output[v] = true
+        end
+    end
+
+    local function empty_set(S)
```
hugomg, line 105: The name empty_set makes me think of function that returns a new empty_set. Perhaps rename to clear_set?

src/pallene/gc.lua
```diff
@@ -28,77 +29,234 @@ local types = require "pallene.types"
...
-    for v_id = 1, #func.vars do
-        variable_is_live_at_gc[v_id] = false
+    repeat
+        local found_dirty_block = false
+        for _,block_i in ipairs(block_order) do
+            if dirty_flag[block_i] then
+                found_dirty_block = true
+                -- CAREFUL: we have to clean the dirty flag BEFORE updating the block or else we
+                -- will do the wrong thing for auto-referencing blocks
+                dirty_flag[block_i] = nil
```
hugomg, line 149: Use `=false` here to avoid putting holes in the table. Holes can't be iterated with ipairs, and can cause Lua to reallocate the table.

Use `nil` for sparse sets only.

hugomg (changes requested):

src/pallene/gc.lua
gabrielsferre: The code is very similar. I was thinking of having a separate flow analysis file at some point, but wanted to do a couple more cases to help me abstract away the pattern.

src/pallene/gc.lua
gabrielsferre: Yeah, it's the same thing. I'll change it to `assert(not (g and kill[v]))`

src/pallene/gc.lua
gabrielsferre: Ok. Will call it clear_set.

src/pallene/gc.lua
gabrielsferre: Didn't know about the reallocation thing. I'll fix this.

src/pallene/gc.lua
gabrielsferre: Yes! We could start with them.

hugomg (approved):

PR #618, opened 2024-07-24
Improving detection of garbage collection bugs

gabrielsferre:
Adding another mode for the _run-tests_ script where it uses a macro that forces Lua to always run the GC, improving our chances of finding GC related bugs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,18 @@ Flag                  | Effect
 ./run-tests -k        | Run all tests even if some tests are failing
 ./run-tests -o gtest  | The gtest output format might be easier to read if you are using print statements for debugging.
 
-For convenience, when the test script is run without any parameters, it also runs the linter at the end.
+For convenience, when the test script is run without any busted parameters, it also runs the linter at the end.
+
+#### Testing garbage collection
+
+Doing proper GC testing takes longer, so we have a special mode for doing that.
+For testing garbage collection more carefully, set the `EXTRACFLAGS`
+environment variable to "-DHARDMEMTESTS" for the `run-tests` script. For
+example:
+
+```sh
+$ EXTRACFLAGS="-DHARDMEMTESTS" ./run-tests
+```
 
 ### Running the benchmarks suite
 
--- a/run-tests
+++ b/run-tests
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-# HOW TO USE: This is a wrapper around busted. The command-line arguments are the same.
+# HOW TO USE: This is a wrapper around busted. Pass "gc" as the first argument for testing the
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -47,7 +47,19 @@ Flag                  | Effect
...
+
+Doing proper GC testing takes longer, so we have a special mode for doing that.
+For testing garbage collection more carefully, pass "gc" as the first parameter
+of the script. The rest of the parameters will do the same as before:
+
+```sh
+$ ./run-tests gc                       # Run all tests but with proper GC testing
+$ ./run-tests gc spec/parser_spec.lua  # Run just one of the test suite files
+                                       # but with proper GC testing
+```
```
hugomg, line 62: What if instead of a magic positional argument, we used an environment variables? We could make a new one just for GC

    HARDMEMTESTS=1 ./runtests

or we could fit it with CFLAGS. (This would also work for other things that play with cflags, such as -fsanitize)

    EXTRACFLAGS=-DHARDMEMTESTS ./runtests

I'm not a fan of the magic positional argument, because it would make more sense to be a --flag instead. And now that I think about it, I also don't like a --flag because we pass all flags through to busted. It would be strange to have only a single flag that's different.

run-tests
```diff
@@ -21,6 +27,11 @@ FLAGS=(--verbose --no-keep-going)
 # Also, add some compiler flags to verify standard compliance.
 export CFLAGS='-O0 -std=c99 -Wall -Werror -Wundef -Wpedantic -Wno-unused'
 
+# When testing the garbage collector, we use -O2 for a greater chance of finding bugs.
+# The HARDMEMTESTS macro makes lualib do garbage collection whenever it can.
+[ -n "$test_gc" ] &&
+    CFLAGS='-DHARDMEMTESTS -O2 -std=c99 -Wall -Werror -Wundef -Wpedantic -Wno-unused'
```
hugomg, line 33: Instead of copy-pasting the whole CFLAGS line, can we add just the -DHARDMEMTESTS to the rest of the line?

Is -O0 too slow? It would be easier to debug with gdb if it comes to that.

hugomg (changes requested):
Looks good, but I'm not sold on the current user interface...

Please rebase this on top of master, before we merge it.

CONTRIBUTING.md
gabrielsferre: I like the EXTRACFLAGS idea, it's even more general. Will do that.

run-tests
hugomg: I don't remember that. I think I might have suggested that out of the concern that it would make the tests run faster.

If -O0 is fast enough, let's stick with -O0

hugomg (approved):

Issue #620, opened 2024-08-07
An `install.sh` to make life easier


singul4ri7y:
As the situation stands, we need to install Lua internals, separate version of Luarocks for that Lua and Pallene Tracer before we manually install Pallene in system. That is too much of a hassle to do by hand. A script to automate it would be nice to have.

bjornbm:
Maybe consider leveraging https://github.com/oploadk/localua which already supports installing Pallene?

hugomg:
Interesting idea! I had forgotten about that script.

Do you know how that installer deals with the C bits of Pallene? Pallene needs to install C header files and shared libraries; if those aren't in a standard directory such as /usr/local then one may need to pass extra flags to the C compiler and linker...

bjornbm:
See https://blog.separateconcerns.com/2022-10-10-localua-pallene.html

> The Pallene compiler does not have an easy way to specify a local Lua directory and I decided against hacking it so you will have to pass CFLAGS as an environment variable for it to work.

hugomg:
We should also pay attention to the installation instructions for the CI, inside .github/workflows. It would be lovely if we could keep those build instructions in sync with the install.sh and the README. Perhaps by having the CI workflow call the install.sh?

PR #623, opened 2024-08-10
Update to Lua 5.4.7

hugomg:
The only bits that broke were ralated to table handling. We should investigate the feasibility of using more stable APIs for those.


hugomg:
Another hiccup to keep in mind. Since ptracer is statically-linked, you will have to reinstall both lua-internals and ptracer. (This might be easier with the #620 install script...)

singul4ri7y (approved):

PR #625, opened 2024-08-17
Added `install.sh` script

### LINKED ISSUE #620: An `install.sh` to make life easier

singul4ri7y:
Changelog: 
 - Added `install.sh`.

> One script to rule them all, one script to find them, one script to bring them all and in the ease-of-use bind them.

 - Update the `ci.yml` script accordingly. If this PR gets merged, we may not need to put our hands on this script ever again (hopefully), Yay!

Resolves #620

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,13 +20,8 @@ mkdir -p $CLONE_DIR
 # Current working directory
 CURR_DIR=$(pwd)
 
-# Whether to install Pallene locally
-SUPERUSER=sudo
-LOCAL_FLAG=
-if [ "${PALLENE_LOCAL:-0}" -eq 1 ]; then
-    SUPERUSER=
-    LOCAL_FLAG=--local
-fi
+# Where to install/locate Lua
+LUA_PREFIX=${LUA_PREFIX:-/usr/local}
 
 # Function to display usage
 usage() {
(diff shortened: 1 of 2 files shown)


singul4ri7y:
Though this script is not as advanced as `localua.sh`, but it is simple and should suffice for most of the use cases. We may add improvements to it in the future.

install.sh
```diff
@@ -0,0 +1,136 @@
...
+        install_luarocks
+        ;;
+    ptracer)
+        install_ptracer
+        ;;
+    pallene)
+        install_pallene
+        ;;
+    --help)
+        usage 0
```
hugomg, line 127: Put the `exit 0`, outside the usage, instead of inside it.

install.sh
```diff
@@ -0,0 +1,136 @@
...
+    sudo make install
+    cd $CURR_DIR
+}
+
+# Install Luarocks
+install_luarocks() {
+    cd $CLONE_DIR
+    wget -O - https://luarocks.org/releases/luarocks-$LUAROCKS_VERSION.tar.gz | tar xzf -
+    cd luarocks-$LUAROCKS_VERSION
+    ./configure --with-lua=/usr/local
```
hugomg, line 69: Not a big fan of hardcoded paths in an installation script.

install.sh
```diff
@@ -0,0 +1,136 @@
...
+# Current working directory
+CURR_DIR=$(pwd)
+
+# Whether to install Pallene locally
+SUPERUSER=sudo
+LOCAL_FLAG=
+if [ "${PALLENE_LOCAL:-0}" -eq 1 ]; then
+    SUPERUSER=
+    LOCAL_FLAG=--local
+fi
```
hugomg, line 29: I suggest if-else

hugomg (commented):
I think we'll need some more time to think about where we want to go with this script.

Some parts of it, such as installing Luarocks, service the Github CI. Other parts, such as the usage, service our users and the README. Unfortunately, it seems that at times they pull in opposite directions.

install.sh
singul4ri7y: May I ask why?

singul4ri7y:
> Some parts of it, such as installing Luarocks, service the Github CI. Other parts, such as the usage, service our users and the README. Unfortunately, it seems that at times they pull in opposite directions.

A tool can be useful to multiple usecases, just because it is working for both CI and users doesn't mean it's bad in general. We don't need something so powerful like localua.sh, which does everything regarding Lua under the sun. We need a simple script which is really useful in CI and increases UX by not allowing users to read through the entire readme and copy pasting commands. Users can install the entire Pallene ecosystem just with a single command (Here ecosystem means, Pallene, Lua internals and Pallene Tracer ;)). If users want, say they want to update just the Lua internals, they can do it in a single command too. So, I cannot see how it is not for the users.

Is the script super feature-rich? No, we don't need that.
Is the script getting most of the work done? Yes.
Will be more features added to the script if need be, which conforms to Pallene ecosystem? Yes.

install.sh
hugomg: Readability; make the function only do one thing.

install.sh
singul4ri7y: But Lua internals will always be installed in /usr/local prefix. If there is a valid, `MYCFLAGS` way to install Lua to a different prefix, we should make it variable. Otherwise, it is fine.

singul4ri7y: Nevermind. Now `LUA_PREFIX` env variable controls related to install and lua prefix.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+# SPDX-License-Identifier: MIT
+
+####
+# One script to rule them all, one script to find them,
+# one script to bring them all and in the ease-of-use bind them.
+####
+
+# Versions
+LUA_VERSION=5.4.7
+LUAROCKS_VERSION=3.9.1
```
hugomg, line 13: Unlike the LUA_VERSION and PT_VERSION, the LUAROCKS_VERSION is only there because our CI needs a Lua version. In theory we could always go with a more recent version. (And in fact, it's probably about time we bump this to a newer Luarocks...)

install.sh
```diff
@@ -0,0 +1,152 @@
...
+
+# Where we will clone
+CLONE_DIR="/tmp/pallene"
+mkdir -p $CLONE_DIR
+
+# Current working directory
+CURR_DIR=$(pwd)
+
+# Where to install/locate Lua
+LUA_PREFIX=${LUA_PREFIX:-/usr/local}
```
hugomg, line 24: Please initialize everything that has a default value here on top. That way it's easier to see what all the configuration options are. (For example, PALLENE_LOCAL, etc)

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    echo "  Use LUA_PLAT to define Lua platform. Defualt is 'linux'."
+    echo "  Available platforms: guess aix bsd c89 freebsd generic ios linux linux-readline macosx mingw posix solaris"
+    echo "    e.g. LUA_PLAT=linux-readline ./install.sh lua"
+    echo
+    echo "  Use LUA_PREFIX to provide Lua install prefix. Default is '/usr/local'."
+    echo "    e.g. LUA_PREFIX=/usr/local ./install.sh lua"
+    echo
+    echo "Pallene Tracer:"
+    echo "  Use PT_TESTS=1 to run Pallene Tracer tests after installation, e.g."
+    echo "    PT_TESTS=1 ./install.sh ptracer"
```
hugomg, line 49: Can we remove the PT_TESTS functionality? I would rather have the CI run those in a separate step than integrate them into the installation script. It complicates the script, and also introduces a dependency on busted.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+    echo "  ptracer  - Install Pallene Tracer"
+    echo
+    echo "Lua Internals:"
+    echo "  Use LUA_CFLAGS, LUA_LDFLAGS, LUA_LIBS and LUA_OBJS to pass flags:"
+    echo "    LUA_MYCFLAGS=-DLUAI_ASSERT ./install.sh lua; or"
+    echo "    LUA_CFLAGS=-fsanitize=address LUA_LDFLAGS=-lasan ./install.sh lua; to enable Address Sanitization"
+    echo
+    echo "  Use LUA_PLAT to define Lua platform. Defualt is 'linux'."
+    echo "  Available platforms: guess aix bsd c89 freebsd generic ios linux linux-readline macosx mingw posix solaris"
+    echo "    e.g. LUA_PLAT=linux-readline ./install.sh lua"
```
hugomg, line 42: Default plat should be "guess"
In the documentation, we should encourage users to use linux-readline option.

install.sh
```diff
@@ -0,0 +1,152 @@
...
+        usage
+        ;;
+    -h)
+        usage
+        ;;
+    *)
+        usage
+        exit 1
+        ;;
+esac
```
hugomg, line 151: I wonder if there is a way to make the script smarter, and only install the things that have a new version.

hugomg (changes requested):

hugomg:
Closing because this has been obsoleted by the new installation process :)

PR #628, opened 2024-08-23
Making an API for doing flow analysis

gabrielsferre:
Abstracting away common patterns of the flow analysis code for reusability.

### DIFF
--- a/src/pallene/flow.lua
+++ b/src/pallene/flow.lua
@@ -20,7 +20,7 @@
 -- processing the block's commands. Each block also has a "kill" and a "gen" set that help transform
 -- the "start" set into the "finish" set. The "kill" set contains the values that will be removed
 -- from the running set while "gen" (as in "generate") contains the values that will be added to it.
--- The flow analysis algorithm's input is a collection of "kill" and "gen" sets for each block and
+-- The flow analysis algorithm's inputs are "kill" and "gen" sets for each block and
 -- the initial values for the "start" sets of each block. During it's runtime, the algorithm updates
 -- the "start" and "finish" sets in a loop until they all converge to some value. The algorithm
 -- requires a loop because a block's "start" set depends on the "finish" set of it's predecessors or
@@ -50,7 +50,6 @@
 --
 --     3.1) Inside the loop that iterates over a block's commands, call "flow.update_set" to update
 --     the set.
-
 
 local flow = {}
 
@@ -197,6 +196,10 @@ local function make_state_list(block_list, flow_info)
     return state_list
 end
 
+-- Does flow analysis on a list of ir.BasicBlock objects. "block_list" is the list of blocks and
+-- "flow_info" is an object of type flow.FlowInfo which contains information about how the flow
+-- analysis will be done. The function returns the list of starting sets, each set in the list
+-- corresponds to the basic block indexed by the same value.
 function flow.flow_analysis(block_list, flow_info)
                                -- ({ir.BasicBlock}, flow.FlowInfo) -> { block_id -> set }
     local state_list = make_state_list(block_list, flow_info)
@@ -267,6 +270,10 @@ function flow.flow_analysis(block_list, flow_info)
     return block_start_list
 end
 
+-- Updates a set "set" according to the "compute_gen_kill" function inside a ir.FlowInfo object
+-- "flow_info". "block_i" and "cmd_i" are the coordinates of the ir.Cmd object that will be used to
+-- update the set, "block_i" is the index of the block and "cmd_i" is the index of the command
+-- inside the block's list of commands.
 function flow.update_set(set, flow_info, block_i, cmd_i) -- (set, flow.FlowInfo, block_id) -> void
     local gk = flow_info.compute_gen_kill(block_i, cmd_i)
     for v,_ in pairs(gk.gen) do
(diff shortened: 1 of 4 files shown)


src/pallene/flow.lua
```diff
@@ -0,0 +1,309 @@
...
+    end
+    return state_list
+end
+
+function flow.flow_analysis(block_list, flow_info)
+
+    local state_list = make_state_list(block_list, flow_info)
+
+    local succ_list = ir.get_successor_list(block_list)
+    local pred_list = ir.get_predecessor_list(block_list)
```
hugomg, line 214: Can we compute these inside the if statements?

src/pallene/flow.lua
```diff
@@ -0,0 +1,309 @@
...
+
+    local block_order
+    local merge_src_list
+    local dirty_propagation_list
+    local order = flow_info.order._tag
```
hugomg, line 219: My heart feels that it would be nice to add comments saying what these variables contain.

src/pallene/flow.lua
```diff
@@ -0,0 +1,309 @@
...
+        end
+    end
+
+    repeat
+        local found_dirty_block = false
+        for _,block_i in ipairs(block_order) do
+            if dirty_flag[block_i] then
+                found_dirty_block = true
+                -- CAREFUL: we have to clean the dirty flag BEFORE updating the block or else we
+                -- will do the wrong thing for auto-referencing blocks
```
hugomg, line 263: What does "the wrong thing" mean?

src/pallene/flow.lua
```diff
@@ -0,0 +1,309 @@
...
+        else
+            tagged_union.error(order)
+        end
+    end
+    return state_list
+end
+
+function flow.flow_analysis(block_list, flow_info)
+
+    local state_list = make_state_list(block_list, flow_info)
```
hugomg, line 211: Is there one state per block? If so, we could make that more obvious by calling it "state_of_block" instead of "state_list" (which in theory could have any number of states, in any order)

hugomg (commented):
In my first look at the code, I don't understand whats the deal with FlowState.Build and FlowState.Apply.  The FlowState.Build looks familiar but I don't get what Apply is all about. Can you help me understand that? I think I need that to be able to complete the review.

src/pallene/flow.lua
gabrielsferre: There's no need, succ_list and pred_list are both used in each of the main if branches.

src/pallene/flow.lua
gabrielsferre: Means we'll always set the dirty flag to false, even if it is set to true during update_block, which might happen if the block  jumps to itself. I'll write a better comment.

gabrielsferre:
The "flow.flow_analysis" function returns only the "start" and "finish" sets for each block ("start" and "finish" might mean "input" and "output" or vice versa, depending on the order the flow analysis is being done), so we still have to determine what is the set of values for each command. The way to do it is using this FlowState.Apply thing, where you pass an object of type FlowState.Apply to the flow.update_set function and it updates the set for you. The reason I needed this two kinds of flow states (FlowState.Apply and FlowState.Build) is that I had to know whether I was updating the kill/gen sets or the actual set of values that is returned to the user.

Currently, the user of the API has to update the FlowState.Apply states manually when looping through commands, which to me feels kinda confusing but is the only way I could think of making it work if we didn't want to save a set for each command.

Is it ok to save a set for each command? If it is then this part of the API could be much simpler. We'd be able to do all the necessary processing inside flow.flow_analysis. We wouldn't be using all those sets inside the "loop until things converge" part of the flow analysis algorithm, only at the end, after we already have the correct input/output sets.

hugomg:
After thinking about it, I think the main problems with the FlowState is that the process_cmd callback is not a pure function. It calls kill_value() and does different things depending on whether we are in BUILD or APPLY mode.

One alternative that would be better, but still not my favorite, would be to turn gen_value and kill_value into callbacks passed to process_cmd. This way, we can use different callbacks when we're in BUILD and when we're in APPLY mode.

But I think the best way would be to turn process_cmd into a pure function. It could receive a single argument (the IR command) and return two tables (GEN and KILL). And while we're at it, rename it into "compute_gen_kill".
During the build phase we can combine the individual GEN/KILL into a GEN/KILL for the entire basic block. And during apply phase we can use them to compute the resulting sets for each command. 

I think it's ok if we call compute_gen_kill twice, once during build phase and once during apply phase. But caching the results is also OK. Focus on keeping the code as simple and easy to understand. We can worry about optimizations later.

gabrielsferre:
Ok, I'll try doing the pure version of process_cmd.

hugomg:
OK, please re-request the review when you're ready.

hugomg (approved):
Tá com uma cara boa, mas acho que a disposição dos comentários está desequilibrada. Tem um comentário grande demais no topo e zero comentários no resto. O melhor lugar pra dizer o que cada função recebe e retorna é do lado da própria função. (Isso é especialmente importante pra saber o que a função retorna, que nem sempre dá pra adivinhar pelo nome)

hugomg (approved):

hugomg:
Oi, você lembra em que pé está este código? Eu já esqueci.

Aparentemente lá atrás eu aprovei as mudanças mas nós nunca mergeamos?

gabrielsferre:
Acho que o código tá todo certo mas o problema tava na falta de comentários. Vou resolver isso aqui e pedir pra dar merge de novo

hugomg:
Obrigado

PR #630, opened 2024-08-31
--emit-lua: Don't "math.ln=math.log" at the start of every file

hugomg:
It cluttered the start of *every* file, for something that was a corner case. I've replaced it with a simple gsub for math.ln-->math.log. It's not perfect, but should be good enough until we can get rid of math.ln.


hugomg:
Wait a second... I need to edit the input *before* I compile it, otherwise the math will be off.

hugomg:
Properly implemented in #631

PR #631, opened 2024-09-05
Optional function arguments

hugomg:
This PR allows Pallene functions to receive optional arguments. Currently any rightmost "any" argument can be optional. In the future we may want to consider a special optional type, for added type safety.

This change allows us to get rid of the `math.ln` hack. Good riddance!


src/pallene/coder.lua
```diff
@@ -593,59 +593,99 @@ function Coder:lua_entry_point_definition(f_id)
...
+        else
+            tmpl = [[
+                int nargs = lua_gettop(L);
+                if (nargs < $min_nargs || nargs > $max_nargs) {
+                    pallene_runtime_arity_error(L, $fname, $min_nargs, $max_nargs, nargs);
+                }
+                for (; nargs < $max_nargs; nargs++) {
+                    ${push_nil}
+                }
+            ]]
```
singul4ri7y, line 623: If I am not wrong, Pallene does not support Variable Arguments. Thing is, variable arguments goes really well with the "Optional Argument" idea. Maybe we can have something like this instead: 

```
if nargs < min_args: 
    error

total_args = nargs - passed_arg
if total_args > 0:
    push_nils
else 
    pop_values
```

Or in case of more arguments than expected, instead of popping, we can pull them to vararg which can be accessed through `(...)` like Lua?

src/pallene/pallenelib.lua
```diff
@@ -493,16 +510,31 @@ static lua_Integer pallene_math_floor(lua_State *L, const char* file, int line,
...
+            PALLENE_UNREACHABLE;
+        }
+
+        if (base == l_mathop(10.0)) {
+            return l_mathop(log10)(x);
 #if !defined(LUA_USE_C89)
-    } else if (base == l_mathop(2.0)) {
-        return l_mathop(log2)(x);
+        } else if (base == l_mathop(2.0)) {
+            return l_mathop(log2)(x);
```
singul4ri7y, line 533: Fix indentation

singul4ri7y (commented):
It looks good overall, just some minor nits.

src/pallene/coder.lua
hugomg: Sorry, I didn't see the notification for your review so I already merged it...

Indeed, optional arguments does help us move towards varargs in the future! However, properly handling varargs will require many more changes than the one you suggested here. I decided it was probably best to deal with those only when we actually implement varargs.

src/pallene/pallenelib.lua
hugomg: Can you please clarify what you mean here? This indentation was intentional. It's similar to what Lua does in lmathlib.c

src/pallene/coder.lua
hugomg: For example, varargs require us to reorder the stuff in the stack. When a Lua function receives a stack like

     [args] [varargs]
     ^stack base
     
 It has to change it into

    [nils] [varargs] [args]
                     ^stack base

so that the varargs are in the "negative" slots and do not interfere with the math for the local variable stack slots.

src/pallene/pallenelib.lua
hugomg: Ifdefs are weird :)

PR #634, opened 2024-09-23
Make traceback tests independent of numbers

### LINKED ISSUE #617: "Skipped  379 frames" is too specific
For a different experiment I was working on, I changed LUAI_MAXCCALLS and that broke the tests that depended on the exact number of stack frames for a C stack overflow.

We should probably change these tests so that they are not sensitive to these small details. Number of frames, exact line numbers, etc.

singul4ri7y:
Changelog: 
 - Free traceback tests from relying on numbers which most-likely to change in the future, hence breaking.

Resolves #617


singul4ri7y:
@hugomg Yes. Because we have numbers in some test functions which also gets replaced unless we be specific.

PR #637, opened 2025-01-21
Clarify field identifier max length

bjornbm:


hugomg:
Thanks! I incorporated this suggestion in #638

PR #641, opened 2025-01-22
Remove gratuitous parentheses in examples

bjornbm:
Of course, ignore this suggestion if you want to keep the parenthesis for stylistic reasons. But to me they add noise and beg the question “are they needed?” (I had to test to learn that they are not).

Perhaps the “operator precedence” of `as` should be documented in the manual?


hugomg (approved):
100% agree on the parenthesis after assignment.

I kind of liked the other parenthesis for stylistic reasons, but you make a convincing argument to remove them.

PR #647, opened 2025-03-10
Test script portability

bjornbm:
Changes avoiding bash-specifics and GNU extensions for `grep`.

These now run on, for example, Alpine Linux (with BusyBox `grep`) and with `dash` shell.

### DIFF
--- a/run-lint
+++ b/run-lint
@@ -25,24 +25,24 @@ for file in \
     examples/*/*.lua  \
     examples/*/*.pln
 do
-    if grep --line-number "[$space$tab]$" "$file"; then
+    if grep -n "[$space$tab]$" "$file"; then
         # Forbid trailing whitespace because some editors like to automatically delete it.
         # Such whitespace can cause spurious diffs later down the road, when someone is working on
         # an unrelated pull request and their editor "helpfully" deletes the trailing whitespace.
         error "File $file has a line that ends in whitespace"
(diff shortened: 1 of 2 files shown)


hugomg:
I was more used to debian, which uses dash by default but also has bash installed and usable via shebang. If I understand you correctly, alpine doesn't have bash at all?

The testing scripts most often used by developers, which may be more willing to install dependencies. What's the context for alpine in particular? Is it for CI?

bjornbm:
Alpine is popular in CI and anywhere you want a small linux. I'm using it in a VM (with [macpine](https://beringresearch.github.io/macpine/)) to test some of the commits I'm sending you.

Alpine doesn't come with bash, but it can be installed (`apk add bash`).

In some environments bash is not available due to the license ([obscure example](https://bianshen00009.gitbook.io/a-guide-to-a-shell/basic-tutorials/configure-the-shell#why-dash)).

I'm not sure the long grep flags are supported on all the BSDs… it seems they are [not on OpenBSD](https://man.openbsd.org/grep.1).

Agreed that most developers will have/be willing to install bash on their dev machines. And GNU (or other) grep. And I could too on my VM. But I feel the cost of portability is low here as the scripts didn't really rely on “bashisms” or GNU extensions.

hugomg (approved):

Issue #640, opened 2025-01-22
README: Document the need to install pallene-tracer


bjornbm:
I understand that there is some ongoing effort #625 to improve the installation process. But in the meantime, the installation section of the README should be updated to at least mention that [pallene-tracer](https://github.com/pallene-lang/pallene-tracer) also must be installed before pallene.

hugomg:
Indeed. Could you please open a PR to improve that part of the manual? In this instance I think you're more qualified than I am to fix it.

In the long term I've been thinking that maybe we should move Pallene to some sort of configure-make-install installation method. That way the configure script could check for dependencies interactively and ask the user to install them. The problem is that I don't yet know how to best tie that together with Luarocks.

bjornbm:
I don't think I have any grand qualifications but sure! But I wonder if we should clear up https://github.com/pallene-lang/pallene-tracer/pull/27 first?

I could of course simply refer to the pallene-tracer README rather than trying to keep the pallene README in synch. But it is not the approach taken for the luarocks installation instructions in the README. 

Let me know what is your preference.

hugomg:
Do both. Copy the simple version of the installation instruction but refer to the readme if the person needs to configure pallene tracer differently

bjornbm:
Understood. But we need to agree first what the installation instructions are. See https://github.com/pallene-lang/pallene-tracer/pull/27. I'm specifically waiting for you to say whether you are happy with the proposed diffs there before copying anything to the pallene README. Thansk

singul4ri7y:
@hugomg Yikes, I completely forgot to change the README!!

Should I PR or @bjornbm would like to do it? And I agree with @hugomg here regarding the installation instructions, we should include only the simple version of pallene-tracer build command relevant to Pallene only.

bjornbm:
Please feel free to proceed of course! I was a bit stuck on 1. time and 2.
the pallene-tracer linker/export flags.

PR #648, opened 2025-03-10
README and CONTRIBUTING updates

bjornbm:
Various updates that seem pertinent.

26a33838b78e44138cc378129fbf62244fadf8b4 (`cd -`) is probably obvious to most developers but could help someone copy-pasting add hoping for the best. But please feel free to reject that one.

These do not address #640 (other than a minimal note/placeholder).


hugomg (approved):
Alright!
Let's leave the `cd -`. If someone says it doesn't work for them, we can remove it later.

PR #657, opened 2025-12-02
Parse pallene types file

igrvlhlb:
This pull request introduces support for Pallene type declaration files (`.d.pln`), allowing the compiler to parse and verify/typecheck type files in addition to regular Pallene source files. The changes include updates to the parser, typechecker, driver, CLI, and utility functions to handle the new file format, as well as new tests to verify correct parsing and type checking of type declaration files.

**Type Declaration File Support**

* Added a new AST union `TypeFile` in `src/pallene/ast.lua` to represent type declaration files and their components (`Typealias`, `Record`, `Decl`).
* Implemented parsing logic for `.d.pln` files in `src/pallene/parser.lua`, including the `TypeDeclarationFile` parser method and extension-based dispatch in the main parse function. [[1]](diffhunk://#diff-0a14cd680e1dedb2b17f72aaabcd2941e9d8673b7134039272e9b94ad8e8c424R311-R344) [[2]](diffhunk://#diff-0a14cd680e1dedb2b17f72aaabcd2941e9d8673b7134039272e9b94ad8e8c424R1213-R1227)
* Extended the typechecker to handle type declaration file ASTs, adding the `check_type_file` method in `src/pallene/typechecker.lua`. [[1]](diffhunk://#diff-6a62295852ef07bf09924550cd268b60a1dfe7c221533febda49e3fbc10a39b6R327-R375) [[2]](diffhunk://#diff-6a62295852ef07bf09924550cd268b60a1dfe7c221533febda49e3fbc10a39b6L57-R64)

**Compiler and CLI Integration (Extra)**

* Added a new CLI flag `--emit-types` to `src/pallene/pallenec.lua`, enabling users to generate `.d.pln` files, and integrated it into the main command dispatch. [[1]](diffhunk://#diff-fd9484c2f19dbaef282ff63b0888d1453430565f6f81b681e4e515cc421d4a49R32) [[2]](diffhunk://#diff-fd9484c2f19dbaef282ff63b0888d1453430565f6f81b681e4e515cc421d4a49R95)

**Utilities and Testing**

* Introduced utility functions in `src/pallene/util.lua` to correctly split and recognize multi-part extensions like `.d.pln`, supporting robust file name handling.
* Added tests for parsing and type checking type declaration files in `spec/parser_spec.lua`.


src/pallene/parser.lua
```diff
@@ -303,6 +308,40 @@ function Parser:Program()
...
+            if not decl.type then
+                self:recoverable_syntax_error(decl.loc,
+                    "type annotation expected in type declaration file")
+            end
+            decl = ast.TypeFile.Decl(decl.loc, decl.name, decl.type)
+            table.insert(decls, decl)
+        else
+            self:unexpected_token_error("a type declaration")
+        end
+    end
```
hugomg, line 333: If we encounter an invalid token, the while loop will exit and this function will return as if there was not a problem.

1. I think the while should be "while not end-of-file"
2. Please create test cases for the "expected a type declaration" errors

src/pallene/typechecker.lua
```diff
@@ -54,7 +54,14 @@ local Typechecker = util.Class()
 -- On failure, returns false and a list of compilation errors
 function typechecker.check(prog_ast)
     local ok, ret = trycatch.pcall(function()
-        return Typechecker.new():check_program(prog_ast)
+        local type_checker = Typechecker.new()
+        if prog_ast._tag == "ast.Program.Program" then
+            return type_checker:check_program(prog_ast)
+        elseif prog_ast._tag == "ast.TypeFile.Decls" then
```
hugomg: This kind of if-else-if doesn't feel right. The type (ast.Program) should be the same in all branches.

I suspect that you actually want to have separate type checking functions instead of a catch-all typechecker.check

src/pallene/util.lua
```diff
@@ -49,6 +49,56 @@ function util.split_ext(file_name)
...
+    local parts = util.split_all_ext(file_name)
+    if #parts >= 3 and parts[#parts] == "pln" and parts[#parts - 1] == "d" then
+        local base_name = table.concat(parts, ".", 1, #parts - 2)
+        return base_name, "d.pln"
+    elseif #parts >= 2 and recognized_extensions[parts[#parts]] then
+        local base_name = table.concat(parts, ".", 1, #parts - 1)
+        return base_name, parts[#parts]
+    end
+end
+
```
hugomg: I wonder if it would be simpler to not split the string. We can could the string itself to see if it ends in ".pln" or ".d.pln".

hugomg (changes requested):

src/pallene/parser.lua
igrvlhlb: > I think the while should be "while not end-of-file"

Should we still throw an error if we don't find at least one declaration? If so, we can go with a `repeat-until`

src/pallene/typechecker.lua
igrvlhlb: The same goes to `Parser.parse()`? Or is it ok for that case?

If I understand it correctly, you are suggesting that we have a new function in the typechecker and the responsibility to choose which one to call must be in `driver.compile_internal()`. Right?

src/pallene/parser.lua
hugomg: We should allow empty files, in case the module exports nothing. (Also test that...)

src/pallene/typechecker.lua
hugomg: Yes.

src/pallene/util.lua
igrvlhlb: Right. Also now I see that logic is not necessary, as "." is not an allowed character in file names.

So what I plan to do is this:
- get rid of that new function `util.split_pallene_ext()` (and `split_all_ext`)
- keep using `util.split_ext` with the following modification
```diff
 function util.split_ext(file_name)
-    local name, ext = string.match(file_name, "(.*)%.(.*)")
+    local name, ext = string.match(file_name, "(.-)%.(.*)")
     return name, ext
 end
```

That way `string.match()` will match `name` only until the first '.', and `ext` will be the rest of the string. Is that ok?

src/pallene/parser.lua
igrvlhlb: I tried to write the error test in a `assert_program_error()` fashion, but I couldn't get it to work. So I wrote it another way ([src/spec/parser_spec.lua:979-990](https://github.com/pallene-lang/pallene/pull/657/files#diff-de29d2e6ff7cc0c7da232de8161563d8c5b1d59bc35f38f1e83c06683e6214d0R979-R990))

hugomg (changes requested):
Acho que precisamos de funções separadas para o compile_internal e o parser.parse.

Pense em qual seria o tipo de parser.parse. Hoje é uma função que as vezes retorna um ast.Program e as vezes retorna um ast.TypeFile. Isso até dá pra escrever em Lua, que é uma linguagem dinâmica, mas não funcionaria se quisessemos converter esse código para Pallene no futuro.

Um problema semelhante acontece na compile_internal, que hoje está com um if que verifica a extensão e retorna tipos de dados diferentes dependendo do valor da extensão.

hugomg (approved):

PR #658, opened 2025-12-02
Add require statement

igrvlhlb:
This PR adds partial support for the `require` command in the compiler's frontend

Here is a non-exaustive list of changes:

AST (src/pallene/ast.lua)

### Parser and AST

 - Added `ast.Type.QualifiedName` for parsing imported types; [[1]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=QualifiedName%20%3D%20%7B%22loc%22%2C%20%22module%22%2C%20%22name%22%7D%2C)
 - Added `ast.Toplevel.Require` for parsing `require()` statements; [[2]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=Require%20%3D%20%7B%22loc%22%2C%20%22local%5Fname%5Fdecl%22%2C%20%22module%5Fname%5Fexp%22%7D%2C)
 - Added code identify `require()`s in regular `ast.Decl.Decl`s at the toplevel and convert them to `ast.Toplevel.Require`s; [[3]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=local%20function%20is%5Frequire%5Fcall,end,-301)
 - Modified `Parser:SimpleType()` to parse qualified type names; [[4]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=local%20name,name,-445)
 - Added constraints to `require()` usage -- they should only happen as a single variable declaration on the toplevel, before any other toplevel node (aside from the module declaration); [[5]](https://github.com/pallene-lang/pallene/pull/658/changes#diff-0a14cd680e1dedb2b17f72aaabcd2941e9d8673b7134039272e9b94ad8e8c424R224-R236) [[6]](https://github.com/pallene-lang/pallene/pull/658/changes#diff-0a14cd680e1dedb2b17f72aaabcd2941e9d8673b7134039272e9b94ad8e8c424R284-R287)

### Typechecker

 - Enabled `typechecker.Def.Import` with module and field fields; [[7]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=Import%20%3D%20%7B%20%22module%22%2C%20%22field%22%20%7D%2C%20%2D%2D%20string%2C%20string)
 - Added `Typechecker:import_symbols()` and `Typechecker:load_require_ast` to load and compile  a `.d.pln` file up to typecheck phase and register imported types and values imported from it to the symbol table; [[8]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=%2B-,function%20Typechecker,end,-310)
 - Added code to handle the new AST nodes aforementioned;
 - Added code to forbid `require()` usages as ordinary expressions; [[9]](https://github.com/pallene-lang/pallene/pull/658/changes#:~:text=if%20var%2Ename,var%2Ename%29);

### Driver

 - Removed the "compilation aborted" sentinel from `compile_internal_d_pln`'s `abort()` so errors in the type declaration file can be reported by te importing module; [[10]](https://github.com/pallene-lang/pallene/pull/658/changes#diff-dd1c837956bc7f1945fc9d7ffb37900f24edf914ea4c577e196d92d800265d09L210)

### IR

 - Added handlers for `ast.Toplevel.Require` and `typechecker.Def.Import`, although they issue errors as they're not implemented yet. [[11]](https://github.com/pallene-lang/pallene/pull/658/changes#diff-b7e15ec209b4a975b271666c15e5c787433e639a4fd5e953b41172b53baab263R367-R369) [[12]](https://github.com/pallene-lang/pallene/pull/658/changes#diff-b7e15ec209b4a975b271666c15e5c787433e639a4fd5e953b41172b53baab263R1059-R1061)

### DIFF
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -951,6 +951,14 @@ describe("Parser /", function()
             }, tl)
         end)
 
+        it("forbids require as toplevel assignment", function()
+            assert_program_error([[
+                local m = {}
+                m.foo = require("bar")
+                return m
+            ]], "calls to require are only allowed in toplevel declarations")
+        end)
+
         it("forbids require after a function definition", function()
             assert_program_error([[
                 local m = {}
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -906,6 +906,120 @@ describe("Parser /", function()
 
     end)
 
+    --
+    -- Require
+    --
+
+    describe("Require statements", function()
+
+        it("produces an ast.Toplevel.Require node", function()
+            local prog_ast = assert_parses_successfuly([[
+                local m = {}
+                local foo = require("bar")
+                return m
+            ]])
+            local tl = prog_ast.tls[1]
+            assert_is_subset({
+                _tag = "ast.Toplevel.Require",
+                local_name_decl = { _tag = "ast.Decl.Decl", name = "foo" },
+                module_name_exp = { _tag = "ast.Exp.String", value = "bar" },
+            }, tl)
+        end)
+
+        it("allows multiple requires at the top", function()
+            local prog_ast = assert_parses_successfuly([[
+                local m = {}
+                local a = require("mod_a")
+                local b = require("mod_b")
+                return m
+            ]])
+            assert_is_subset({ _tag = "ast.Toplevel.Require" }, prog_ast.tls[1])
+            assert_is_subset({ _tag = "ast.Toplevel.Require" }, prog_ast.tls[2])
+        end)
+
+        it("forbids require after a function definition", function()
+            assert_program_error([[
+                local m = {}
+                function m.f() end
+                local foo = require("bar")
+                return m
+            ]], "require statements must appear before any other toplevel statement")
+        end)
+
+        it("forbids require after a record definition", function()
+            assert_program_error([[
+                local m = {}
+                record R
+                    x: integer
+                end
+                local foo = require("bar")
+                return m
+            ]], "require statements must appear before any other toplevel statement")
+        end)
+
+        it("forbids require after a typealias definition", function()
+            assert_program_error([[
+                local m = {}
+                typealias T = integer
+                local foo = require("bar")
+                return m
+            ]], "require statements must appear before any other toplevel statement")
+        end)
+
+        it("forbids type annotation in require", function()
+            assert_program_error([[
+                local m = {}
+                local foo: integer = require("bar")
+                return m
+            ]], "type annotation is not allowed at require statements")
+        end)
+
+        it("forbids multiple assignment in require", function()
+            assert_program_error([[
+                local m = {}
+                local a, b = require("x"), require("y")
+                return m
+            ]], "cannot use a multiple-assignment in require statements")
+        end)
+
+        it("forbids require with no arguments", function()
+            assert_program_error([[
+                local m = {}
+                local foo = require()
+                return m
+            ]], "require() must be called with exactly one argument")
+        end)
+
+        it("forbids require with multiple arguments", function()
+            assert_program_error([[
+                local m = {}
+                local foo = require("a", "b")
+                return m
+            ]], "require() must be called with exactly one argument")
+        end)
+
+    end)
+
+    --
+    -- Qualified type names
+    --
+
+    describe("Qualified type names", function()
+
+        it("parses mod.Type as QualifiedName", function()
+            assert_type_ast("mod.MyType", {
+                _tag = "ast.Type.QualifiedName",
+                module = "mod",
+                name = "MyType",
+            })
+        end)
+
+    end)
+
+    --
+    -- Type declaration file
+    --
+
     describe("Type declaration file / ", function()
 
         local tdf = [[
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -937,6 +937,20 @@ describe("Parser /", function()
             assert_is_subset({ _tag = "ast.Toplevel.Require" }, prog_ast.tls[2])
         end)
 
+        it("allows require argument without parenthesis", function()
+            local prog_ast = assert_parses_successfuly([[
+                local m = {}
+                local foo = require "bar"
+                return m
+            ]])
+            local tl = prog_ast.tls[1]
+            assert_is_subset({
+                _tag = "ast.Toplevel.Require",
+                local_name_decl = { _tag = "ast.Decl.Decl", name = "foo" },
+                module_name_exp = { _tag = "ast.Exp.String", value = "bar" },
+            }, tl)
+        end)
+
         it("forbids require after a function definition", function()
             assert_program_error([[
                 local m = {}
--- a/spec/typechecker_spec.lua
+++ b/spec/typechecker_spec.lua
@@ -142,27 +142,38 @@ end)
 --
 
 describe("Require", function()
-
-    it("forbids require calls as expressions", function()
-        assert_error([[
-            function m.f()
-                local x = require("foo")
-            end
-        ]], "calls to require are only allowed in toplevel declarations")
-    end)
-
-    it("forbids require calls as expressions", function()
+    it("forbids require calls in toplevel assignments", function()
         assert_error([[
             m.x = require("foo")
         ]], "calls to require are only allowed in toplevel declarations")
     end)
 
     it("forbids non-string argument", function()
-        -- Compile only to typechecker; the parser will produce a Toplevel.Require
-        -- with a non-string exp which the typechecker should reject.
         assert_error([[
-            local x = require(42)
+            local x = require(1)
         ]], "require argument must be a string literal")
+    end)
+
+    it("forbids require calls as expressions", function()
+        assert_error([[
+            function m.f()
+                local x = require("foo")
+            end
+        ]], "the use of 'require' is forbidden here")
+    end)
+
+    it("forbids using require as value", function()
+        assert_error([[
+            local x = require
+        ]], "the use of 'require' is forbidden here")
+    end)
+
+    it("forbids assigning to require", function()
+        assert_error([[
+            local function f()
+                require = 1
+            end
+        ]], "the use of 'require' is forbidden here")
     end)
 
     it("forbids shadowing require as local variable declaration", function()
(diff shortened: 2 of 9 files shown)


src/pallene/assignment_conversion.lua
```diff
@@ -389,6 +389,8 @@ function Converter:visit_stat(stat)
 
     elseif  tag == "ast.Stat.Break" then
         -- empty
+    elseif  tag == "ast.Stat.Require" then
```
hugomg, line 392: Consider adding a comment to remind that Require only accepts string literals, and thus there are no "exp" children to visit.

src/pallene/ast.lua
```diff
@@ -20,17 +20,19 @@ define_union("TypeFile", {
...
-    Name     = {"loc", "name"},
-    Array    = {"loc", "subtype"},
-    Table    = {"loc", "fields"},
-    Function = {"loc", "arg_types", "ret_types"},
+    Nil           = {"loc"},
+    Name          = {"loc", "name"},
+    Array         = {"loc", "subtype"},
+    Table         = {"loc", "fields"},
+    Function      = {"loc", "arg_types", "ret_types"},
+    QualifiedName = {"loc", "module", "name"},
```
hugomg, line 28: Question: would it be better or would it be worse if we had a single constructor for both names and qualified names?

E.g. setting the module as nil for unqualified names, or representing qualified names as a dotted string.

src/pallene/to_ir.lua
```diff
@@ -1054,6 +1056,9 @@ function ToIR:exp_to_value(bb, exp, is_recursive)
                     -- See https://github.com/pallene-lang/pallene/issues/337
                     error("not implemented")
                 end
+            elseif def._tag == "typechecker.Def.Import" then
+                -- TODO: Handle module imported fields
+                return ir.Value.Nil -- added just so the compiler don't break while this case is not implemented
```
hugomg: maybe it's OK to break it? There is already a flag to generate the module file without running the to_ir

src/pallene/ast.lua
igrvlhlb: I can't clearly see any disadvantages neither on having a `QualifiedName` constructor, nor adding a new field `"module"` to the `Name` constructor. I'd say I prefere the new constructor as it provides more explicit handling of that particular case and clearly distinguishes it from simple names. Do you have an opinion/preference on that?

> or representing qualified names as a dotted string.

While that could also work, I'd avoid due to this drawback: although we could recognize and store the dotted string as the `Name` identifier with no problem, there would be no straightforward way to verify if the name is qualified or not other than looking for a dot in the string.

src/pallene/assignment_conversion.lua
igrvlhlb: Done

src/pallene/to_ir.lua
igrvlhlb: Done

src/pallene/ast.lua
hugomg: I don't have a strong preference. Let's go ahead with QualifiedName then.

One possible advantage is that this way it is is probably more difficult to accidentally treat a qualified name as if it were not qualified, because we don't have to do another if statement inside the branch that handles the Name constructor.

spec/parser_spec.lua
```diff
@@ -906,6 +906,120 @@ describe("Parser /", function()
 
     end)
 
+    --
+    -- Require
+    --
+
```
hugomg, line 912: Can we add tests that check for require without parenthesis?

```
local foo = require "bar"
local foo = require 'bar'
```

spec/typechecker_spec.lua
```diff
@@ -123,6 +137,184 @@ describe("Module", function()
...
+    end)
+
+    it("forbids require calls as expressions", function()
+        assert_error([[
+            local function foo(m: any) end
+            function m.f()
+                foo(require("bar"))
+            end
+        ]], "calls to require are only allowed in toplevel declarations")
+    end)
```
hugomg, line 169: Is this test redundant with the test above it?

src/pallene/parser.lua
```diff
@@ -200,10 +200,105 @@ local is_toplevel_first = Union({
...
 })
 
+local function is_require_call(exp)
+    return exp._tag         == "ast.Exp.CallFunc" and
+           exp.exp._tag     == "ast.Exp.Var"      and
+           exp.exp.var._tag == "ast.Var.Name"     and
+           exp.exp.var.name == "require"
+end
+
+local function has_require_exp(stat)
```
hugomg, line 210: This function is too dynamic right now. It receives an arbitrary "stat", checks if it has an "exps" field, and then does something with that.

Since this function is only used once, would it be OK to inline it? And would it make sense to only look for require in ast.Decl nodes?

spec/typechecker_spec.lua
```diff
@@ -123,6 +137,184 @@ describe("Module", function()
...
+
```
hugomg, line 170: It might be a good idea to also test what happens if we use require in other toplevel statements.

```
local m = {}
m.field = require("a") -- Stat.Assign
```

src/pallene/typechecker.lua
```diff
@@ -759,7 +840,11 @@ function Typechecker:check_var(var)
     if     tag == "ast.Var.Name" then
         local sym = self.symbol_table:find_symbol(var.name)
         if not sym then
-            type_error(var.loc, "variable '%s' is not declared", var.name)
+            if var.name == "require" then
```
hugomg, line 851: We have some special logic here to catch uses of the name "require" but it is not replicated in other places that call symbol_table:find_symbol. 

Perhaps it would be better to add an entry for "require" in the symbol table,  at the same time that we introduce entries for other builtin functions?

src/pallene/parser.lua
```diff
@@ -200,10 +200,105 @@ local is_toplevel_first = Union({
...
+end
+
 --
 -- Toplevel
 --
 
+function Parser:separate_requires_from_tl_stats(tl_stats)
+    assert(tl_stats._tag == "ast.Toplevel.Stats")
+    local tls = {}
+    local stats_group = {}
```
hugomg, line 249: It seems that we have some clever logic with commit_stats to allow require statements to be interspersed with other toplevel statements

    local x = require("x")
    function m.f() .... end
    local y = require("y")

Could we make it more restricted so that require statements are only allowed at the very top of the file, before everything else?

spec/typechecker_spec.lua
hugomg, line 170: Please add tests for what happens when the programmer creates a function or local variable called require.

It seems that the type checker is careful to examine the symbol table to see if the name "require" refers to the builtin function or if it was redefined by the programmer, but I could find no test cases that cover the latter case.

hugomg (changes requested):
Looks promising! There are still some parts of  driver.lua and typechecker.lua that I haven't reviewed closely, but I think I will be able to review those better if I look at them after we address the comments I left in this review.

spec/parser_spec.lua
igrvlhlb: Sure! I'll add that

spec/typechecker_spec.lua
igrvlhlb: You're right. So I'll keep only the one above it, but with this description (`forbids require calls as expressions`). Is that ok?

spec/typechecker_spec.lua
igrvlhlb: I realised that case wasn't covered. I'll add code to check that

src/pallene/parser.lua
igrvlhlb: That logic is in `Parser:tls_emerge_requires()`. Here [parser.lua:285-288](https://github.com/pallene-lang/pallene/pull/658/changes/BASE..1b4f946f6c95d748c313958e3f6b2b42becd07f4#diff-0a14cd680e1dedb2b17f72aaabcd2941e9d8673b7134039272e9b94ad8e8c424R285-R288) we look at the already transformed list of stats for ensuring that `require`s appear before everything else

spec/typechecker_spec.lua
igrvlhlb: Done

spec/typechecker_spec.lua
igrvlhlb: Done

spec/parser_spec.lua
igrvlhlb: Done

src/pallene/parser.lua
igrvlhlb: The code was flawed because we should not perform this check for every stat. I refactored it so that we only perform the check for `ast.Stat.Decl` and `ast.Stat.Assign`. The former case is where we expect to find a valid `require`, while the latter case exists solely so that we can throw a specific error indicating that such usage is disallowed.

Given that new check, do you think it makes sense keeping the `has_require_exp(stat)` function?

spec/typechecker_spec.lua
hugomg: I think this is the only comment that is still not addressed

hugomg (changes requested):

spec/typechecker_spec.lua
igrvlhlb: Added code for forbidding that local variables/functions shadow the name `require` at [f122ddd](https://github.com/pallene-lang/pallene/pull/658/commits/f122ddd12f5acd723329d789f27407aea625f7a2)

src/pallene/typechecker.lua
igrvlhlb: Added code for forbidding that local variables/functions shadow the name require at [f122ddd](https://github.com/pallene-lang/pallene/pull/658/commits/f122ddd12f5acd723329d789f27407aea625f7a2).

I did not add it to the symbol table as it could still be shadowed, but I added a check when adding symbols to the symble table so that it doesn't happen

hugomg:
Is this ready to merge? If yes, could you please rebase this on top of the master branch?

igrvlhlb:
> Is this ready to merge? If yes, could you please rebase this on top of the master branch?

@hugomg Yes, it's ready. I've addressed your comments and just rebased the branch on top of master.

hugomg (approved):

hugomg:
Ok! Please merge

PR #661, opened 2026-04-19
Lua 5.5.0 initial fixes

michaeam:
Trivial test, i.e. by far **not** complete: 

```
$ pallenec --emit-lua examples/arithmetic/arithmetic.pln
$ $ cat examples/arithmetic/arithmetic.lua
-- Copyright (c) 2020, The Pallene Developers
-- Pallene is licensed under the MIT license.
-- Please refer to the LICENSE and AUTHORS files for details
-- SPDX-License-Identifier: MIT

local m = {}

function m.add(a, b)
    return (a) + (b)
end

function m.subtract(a, b)
    return (a) - (b)
end

return m
```

### DIFF
--- a/pallene-dev-1.rockspec
+++ b/pallene-dev-1.rockspec
@@ -12,7 +12,7 @@ description = {
    license = "MIT"
 }
 dependencies = {
-   "lua == 5.4",
+   "lua >= 5.4, < 5.6",
    "lpeg >= 1.0",
    "argparse >= 0.7.0",
 }
--- a/src/pallene/C.lua
+++ b/src/pallene/C.lua
@@ -144,8 +144,9 @@ end
 function C.reformat(input)
     local out = {}
     local depth = 0
-    for line in input:gmatch("([^\n]*)") do
-        line = line:match("^%s*(.-)%s*$")
+    local line
+    for _line in input:gmatch("([^\n]*)") do
+        line = _line:match("^%s*(.-)%s*$")
 
         local nspaces
         if line:match("^#") then
--- a/src/pallene/pallenelib.lua
+++ b/src/pallene/pallenelib.lua
@@ -626,7 +626,13 @@ static TString *pallene_tostring(lua_State *L, const char* file, int line, TValu
             if (ttisinteger(&v)) {
                 len = lua_integer2str(buff, MAXNUMBER2STR, ivalue(&v));
             } else {
-                len = lua_number2str(buff, MAXNUMBER2STR, fltvalue(&v));
+                len = lua_number2strx(
+                        L,
+                        buff,
+                        MAXNUMBER2STR,
+                        "%" LUA_NUMBER_FRMLEN "a",
+                        fltvalue(&v))
+                    ;
                 if (buff[strspn(buff, "-0123456789")] == '\0') {  /* looks like an int? */
                   buff[len++] = lua_getlocaledecpoint();
                   buff[len++] = '0';  /* adds '.0' to result */
--- a/src/pallene/pallenelib.lua
+++ b/src/pallene/pallenelib.lua
@@ -464,7 +464,8 @@ static TValue *pallene_getstr(size_t len, Table *t, TString *key, int *cache)
     if (len <= LUAI_MAXSHORTLEN) {
         return pallene_getshortstr(t, key, cache);
     } else {
-        return cast(TValue *, luaH_getstr(t, key));
+        TValue idx;
+        return cast(TValue *, luaH_getstr(t, key, &idx));
     }
 }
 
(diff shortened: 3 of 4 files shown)


src/pallene/C.lua
```diff
@@ -144,8 +144,9 @@ end
 function C.reformat(input)
     local out = {}
     local depth = 0
-    for line in input:gmatch("([^\n]*)") do
-        line = line:match("^%s*(.-)%s*$")
+    local line
+    for _line in input:gmatch("([^\n]*)") do
+        line = _line:match("^%s*(.-)%s*$")
```
hugomg, line 149: Can we declare the variable inside the for loop?

    for _line in input:gmatch("([^\n]*)") do
        local line = _line:match("^%s*(.-)%s*$")

pallene-dev-1.rockspec
```diff
@@ -12,7 +12,7 @@ description = {
    license = "MIT"
 }
 dependencies = {
-   "lua == 5.4",
+   "lua >= 5.4, < 5.6",
```
hugomg, line 15: This should be exactly 5.5.

In fact, if Luarocks allowed it, we would specify the minor patch 5.5.0. Pallene is only compatible with one specific minor patch at a time, because we rely on internal APIs that can break even in a minor patch.

src/pallene/pallenelib.lua
```diff
@@ -464,7 +464,8 @@ static TValue *pallene_getstr(size_t len, Table *t, TString *key, int *cache)
     if (len <= LUAI_MAXSHORTLEN) {
         return pallene_getshortstr(t, key, cache);
     } else {
-        return cast(TValue *, luaH_getstr(t, key));
+        TValue idx;
+        return cast(TValue *, luaH_getstr(t, key, &idx));
```
hugomg, line 468: This looks sketchy. luaH_getstr now returns a lu_byte instead of a pointer. It probably makes no sense to cast the return value to TValue* ?

We need to look closer into this, but its possible that this API change is related to the 5.5 changes to the internal representation of arrays, which now stores the TValue tag separately from the payload. https://sol.sbc.org.br/index.php/sblp/article/view/30252

hugomg (changes requested):
Looks promising!

I'll wait until we sort out the lua-internals PR before we run this through the CI. The CI won't work unless that other PR is merged first.

src/pallene/C.lua
michaeam: Good point. It's the loop variable that's now RO (I god a bit confused). If using an extra variable, better keep the the scope tight.

src/pallene/pallenelib.lua
michaeam: Agreed. I just made the minimal effort tweaks (this snippet i stole from somewhere else, can't remember where) to just make it build, and was surprised it actually run as well.

src/pallene/pallenelib.lua
```diff
@@ -625,7 +626,13 @@ static TString *pallene_tostring(lua_State *L, const char* file, int line, TValu
...
                 len = lua_integer2str(buff, MAXNUMBER2STR, ivalue(&v));
             } else {
-                len = lua_number2str(buff, MAXNUMBER2STR, fltvalue(&v));
+                len = lua_number2strx(
+                        L,
+                        buff,
+                        MAXNUMBER2STR,
+                        "%" LUA_NUMBER_FRMLEN "a",
+                        fltvalue(&v))
+                    ;
```
michaeam, line 635: Badly formatted, not following projects code-standard. Will fix, if there's more around this line let me know.

pallene-dev-1.rockspec
michaeam: Hmm, yes. I was wondering about that.  Not sure how to handle even minor (you meant tweak right? i.e. the zero). I suppose even a finer granularity would be needed, worst case different `rockspec` name(s) would be needed to assure version compatibility everywhere it can break.. Will change to exactly `5.5.0` and check if `luarocks` is happy with it, otherwise pinning to `5.5.`

src/pallene/pallenelib.lua
michaeam: The following must be related to the internal Table representation changes in Lua 5.5.:

```
/tmp/lua_wl2viY:787:24: warning: initialization of ‘TValue *’ from incompatible pointer type ‘
Value *’ [-Wincompatible-pointer-types]                                                       
  787 |         TValue *slot = &x1->array[x5 - 1];
```

Specifically, but quite easy to reproduce, from:
``` 
pallenec examples/sum_of_array/sum_of_array.pln
lua examples/rectangle/main.lua
```

michaeam (commented):
Promising but more work is needed.

PR #667, opened 2026-05-09
Improve the build process

ninadsachania:


hugomg:
Hi! Is this PR obsolete? Should we close it?

ninadsachania:
Yes, we should close it. This is an old PR with different design decisions about improving the build process. I kept it around to crib from it.

PR #669, opened 2026-05-28
Vendor argparse version 0.7.1

ninadsachania:


hugomg (commented):
It there a reason why this file is being put in `src`? I expected that we would put the vendored stuff in `deps`.

ninadsachania:
> It there a reason why this file is being put in `src`? I expected that we would put the vendored stuff in `deps`.

We had decided that pure Lua deps will go inside the `src` directory and the rest will go inside `deps`.

ninadsachania:
I made the requisite changes:

- Moved `argparse.lua` from `src/` to `deps/argparse/`.
- Created a new branch called `vendor` that this PR will merge into.

VENDORING.md
```diff
@@ -6,3 +6,10 @@
 - **Tag:** `0.5.0a`
 - **Commit ID:** `9c3758bc48733d5b1bc4f950c426d9142d548a2b`
 
+## `src/argparse.lua`
```
luizromario-srios:
```suggestion
## `deps/argparse/argparse.lua`
```

luizromario-srios (changes requested):

VENDORING.md
ninadsachania: Fixed. Force pushed the fix so as to keep the commits clean.

luizromario-srios: You could also do fixup! commits as to preserve the PR history and then squash them once it's time to merge

ninadsachania:
Vendored the entire argparse repository instead of just `argparse.lua`.

luizromario-srios (approved):
LGTM

PR #671, opened 2026-06-12
Build and Install Pallene

ninadsachania:


Makefile
```diff
@@ -0,0 +1,69 @@
+INSTALL         = install
+INSTALL_PROGRAM = $(INSTALL)
+
+LUA_VERSION= 5.4
+
+# Generated by ./configure
+include config.mk
+
+RM     = rm -f
+RM_DIR = rm -rf
```
hugomg: Please put this together with INSTALL and INSTALL_PROGRAM

Makefile
```diff
@@ -0,0 +1,69 @@
...
+	$(RM) config.mk
+
+install: all
+	# Make all the requisite directories.
+	mkdir -p $(INSTALL_PREFIX)/include
+	mkdir -p $(INSTALL_PREFIX)/bin
+	mkdir -p $(INSTALL_PREFIX)/share/lua/$(LUA_VERSION)
+	mkdir -p $(INSTALL_PREFIX)/lib/lua/$(LUA_VERSION)
+
+	# Install all the dependencies first.
```
hugomg, line 33: Might be worth pointing out that we have to install the files one by one because the upstream projects do not provide an "install" target in their makefiles.

Please also add a reminder comment saying that any changes to install should be mirrored in uninstall. (bonus points if the test suite can check that uninstall works properly)

Makefile
```diff
@@ -0,0 +1,69 @@
...
+	cd deps/pallene-tracer && $(MAKE) install PREFIX=$(INSTALL_PREFIX)
+	$(INSTALL) -D deps/argparse/src/argparse.lua $(INSTALL_PREFIX)/share/lua/$(LUA_VERSION)
+	$(INSTALL) -D deps/lpeg/re.lua               $(INSTALL_PREFIX)/share/lua/$(LUA_VERSION)
+	$(INSTALL) -D deps/lpeg/lpeg.so              $(INSTALL_PREFIX)/lib/lua/$(LUA_VERSION)
+
+	# Install Pallene.
+	cp -r src/pallene $(INSTALL_PREFIX)/share/lua/$(LUA_VERSION)
+	cp -r src/bin     $(INSTALL_PREFIX)/share/lua/$(LUA_VERSION)
```
hugomg, line 45: I think we should use INSTALL instead of cp?

pallenec
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+LUA_VERSION="5.4"
+PREFIX="$(dirname "$(realpath "$0")")"
```
hugomg, line 4: It would be better to generate this file at build time. The path strings should be fixed strings and should not be recalculated at run-time based on argv $0.

I believe the Luarocks binary scripts are generated and don't peek at argv, right?

hugomg (changes requested):

luizromario-srios (changes requested):
```
make && make install
cd ./deps/pallene-tracer  && make
make[1]: Entering directory '/home/luizromario/pallene/deps/pallene-tracer'
cc -DPT_DEBUG -g -std=c99 -pedantic -Wall -Wextra -Wformat-security -I/usr/include -I.  -L/usr/lib -Wl,-E pt-lua.c -o pallene-lua -llua -lm
pt-lua.c:58:10: fatal error: lua.h: No such file or directory
   58 | #include "lua.h"
      |          ^~~~~~~
compilation terminated.
```

I just noticed that pallene-tracer and lpeg makefiles don't follow the prefix defined in our `configure` script. This caused a problem in my build specifically, since I installed lua-internals in a custom user prefix, not in the system. Maybe we don't need to deal with it right now, but we need to think what we should do.

ninadsachania:
> ```
> make && make install
> cd ./deps/pallene-tracer  && make
> make[1]: Entering directory '/home/luizromario/pallene/deps/pallene-tracer'
> cc -DPT_DEBUG -g -std=c99 -pedantic -Wall -Wextra -Wformat-security -I/usr/include -I.  -L/usr/lib -Wl,-E pt-lua.c -o pallene-lua -llua -lm
> pt-lua.c:58:10: fatal error: lua.h: No such file or directory
>    58 | #include "lua.h"
>       |          ^~~~~~~
> compilation terminated.
> ```
> 
> I just noticed that pallene-tracer and lpeg makefiles don't follow the prefix defined in our `configure` script. This caused a problem in my build specifically, since I installed lua-internals in a custom user prefix, not in the system. Maybe we don't need to deal with it right now, but we need to think what we should do.

Yeah, right now we're assuming that the user has installed our special Lua in a known location and the header files and the libraries are available on the `PATH`.

This will not be a problem once we vendor our special Lua.

Makefile
ninadsachania: OK. Will change.

Makefile
ninadsachania: I tried that. `install` doesn't really work with directories.

pallenec
ninadsachania: OK. Will change.

Makefile
ninadsachania: OK. Will change.

luizromario-srios (approved):
LGTM

PR #676, opened 2026-08-01
Upgrade Pallene Tracer to version 0.6.0

ninadsachania:


hugomg (commented):
The PR as a whole looks confusing because github is computing a gigantic diff that includes everything we vendored so far. Can you figure out why that happened?

As for the pallene tracer stuff, I checked the last 3 commits only and those seem to be OK.

PR #679, opened 2026-09-06
MacOS - arm64

humbertodias:
## Changes
- Fix macOS build
- Added CI for macOS
Screenshot 2026-09-06 at 1 56 15 AM

## How to test
Compile + Run
```sh
./configure --prefix=/tmp/pallene
make
PATH=$PATH:/tmp/pallene
cd examples/factorial
pallenec factorial.pln
pallene-lua -l factorial
```
output
```
Lua 5.4.7 with core API  Copyright (C) 1994-2024 Lua.org, PUC-Rio
> factorial.factorial(5)
120
```


hugomg:
Hi Humberto,

Looks like this is your first time here. Could you tell us a bit about how you came across Pallene and how you're using it?

humbertodias:
> Hi Humberto,
> 
> Looks like this is your first time here. Could you tell us a bit about how you came across Pallene and how you're using it?

Hello Hugo,
I came across Pallene while researching ways to use a Lua-like language for game development on constrained/retro platforms.

I'm currently working on a game written in another language and exploring a possible migration to Lua, I'd like to experiment with compiling a restricted Lua/Pallene-like language to C so that the same game logic could potentially target platforms such as the Sega Genesis using SGDK - MC68000.

hugomg:
Neat! @igrvlhlb, could you please take a look at this PR?

igrvlhlb:
> Neat! @igrvlhlb, could you please take a look at this PR?

@hugomg it seems to work! I was able compile the project without any manual intervention.
Thank you @humbertodias for this PR. Those are issues we've recently found but haven't addressed yet.

hugomg:
I can't test these changes myself so I'm glad to hear that it works for you. 

Looking at the choose I have two main questions: 

1. I remember being told by Igor someone that the replacement for "install -D" would need to be a combination of "mkdir -p" and plain "install". But it seems this PR just does a plain install the mkdirs. What gives? 

2. In this PR we do feature detection inside the makefiles, using gnu makefile if statements. Would it be possible to instead do the feature detection in the "configure" script?

igrvlhlb:
>     1. I remember being told by Igor someone that the replacement for "install -D" would need to be a combination of "mkdir -p" and plain "install". But it seems this PR just does a plain install the mkdirs. What gives?

About (1), I didn't notice that the directories were actually being created [just above](https://github.com/humbertodias/pallene/blob/8ab34bc601702315834c33d7dd7d7a07906b5b0e/Makefile#L36-L39) the install commands. So while the `-D` flag was wrong on macOS, I guess it was redundant on Linux

